PR #52, opened 2013-08-03
Master

Wet-Host:
Updated the code for a few changes to support EASY sorting of the levels.

1) TODO not sure how you wanted to handle it.. was the 

`sort_order` int(11) NOT NULL DEFAULT '0',    added to the Table structure for table `wp_pmpro_membership_levels`

2) Fixed what looked to be a bug for the orginal order by clause since if that condition was not true the order by would be skipped.

### DIFF
--- a/adminpages/advancedsettings.php
+++ b/adminpages/advancedsettings.php
@@ -27,6 +27,10 @@
 		//tos
 		pmpro_setOption("tospage");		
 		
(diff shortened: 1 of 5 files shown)


strangerstudios:
Closing this since the drag and drop sorting included in a recent release addresses this.

PR #145, opened 2014-08-29
Resolve issues accessing non-existent membership object when canceling a membership

Mrnt:
When changing a user's membership to 'None' in the Wp Admin user profile, several debug notices are generated regarding accessing non existent membership attributes on the user object. This pull request is intended to address that.


JoelStickney:
Line 759: I believe the ending period causes a double period in the user email (admin email doesn't suffer from this). Example:

An administrator at Site has changed your membership level.

The new level is Gold..

strangerstudios:
Thanks for this. Closing as another patch addresses this in a similar way.

PR #146, opened 2014-09-19
Corrected acronym "Card Verifcation Value"

roufamatic:
The correct term is CVV (Card Verification Value) but the page uses CCV instead. This fixes that error. (Didn't touch the image names though.)


strangerstudios:
Thanks!

PR #157, opened 2014-12-31
fix pmpro_has_membership_access() function

ouahabi:
pmpro_has_membership_access function conflict with some other plugins like "TimberLibrary" especially when the global $post is null, and it generates a runtime warning.
So we should verify the variable before the assignment instruction (Line 10).


strangerstudios:
Not sure when, but I also made this change. It's in the 1.8 version coming out soon. Thanks for this!

PR #165, opened 2015-01-19
I18N issue

eighty20results:
"Print" was hardcoded as reported by @ecoeficiente. Now translatable.


strangerstudios:
Merged in. Thanks.

Issue #169, opened 2015-02-16
Warning in includes/profile.php


eighty20results:
With WP_DEBUG enabled I see the following warning:

`Trying to get property of non-object in paid-memberships-pro/includes/profile.php on line 43`

Should probably ensure that $users->membership_level->ID is set before comparing it.

PR #170, opened 2015-02-16
Fix issue reported against profile.php

eighty20results:
Switch to using selected() and test whether $user->membership_level->ID is set before attempting to use it.


eighty20results:
Fixes issue #169

PR #178, opened 2015-03-08
Fix permissions

meths:
There is no reason for files to be executable and a plugin dealing with personal details and linking to payment gateways should have security as a top priority.

Follow best practice as outlined here: http://codex.wordpress.org/Hardening_WordPress


strangerstudios:
You're right. I accidentally changed them during another commit. I'll get them fixed up.

PR #181, opened 2015-03-10
Adding a new filter called pmpro_stripe_create_subscription_array

nickd32:
**NOTE**: The only real change here is lines 1409-1410.  The rest was removing excess whitespace and converting all line endings to CRLF.  (general cleanup)

--
This change is to allow other plugins to modify the subscription object before sending to Stripe. This is most useful for setting a custom ‘trial_end’ timestamp inside Stripe (which cannot be done inside the Stripe plan object).

https://stripe.com/docs/api/php#create_subscription

The Stripe plan object only lets you set a number of ‘trial_period_days’ — which is not granular enough if you want subscriptions to renew at a specific time of day — like noon or 3pm or whatever.

--
This pull request suggested by Jason in the v1.8 BETA blog post:
http://www.paidmembershipspro.com/2015/03/v1-8-beta-download-available-now-full-release-in-1-week/#comment-21751

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1,10 +1,10 @@
-<?php	
+<?php
 	//include pmprogateway
-	require_once(dirname(__FILE__) . "/class.pmprogateway.php");		
-	
+	require_once(dirname(__FILE__) . "/class.pmprogateway.php");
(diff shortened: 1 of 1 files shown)


strangerstudios:
I like it. Merged into 1.8.1 and will go out now. Thanks.

PR #183, opened 2015-03-12
Adding Bitcoin / Litecoin / Dogecoin / Darkcoin / Altcoins Gateway via GoUrl.io

cryptoapi:
Support Bitcoin, Litecoin, Paycoin, Dogecoin, Darkcoin, Speedcoin,
Reddcoin, Potcoin, Feathercoin, Vertcoin, Vericoin virtual coins in Paid Memberships Pro.

Screenshots - https://gourl.io/bitcoin-payments-paid-memberships-pro.html#screenshot

--------------------
Also useful if you can create new function $order->payment_complete() similar as in woocommerce (when an order is paid for). See here - http://docs.woothemes.com/document/payment-gateway-api/


spdcoin:
like it :+1:

RugWarrior:
:+1:

strangerstudios:
Thanks so much for this! I've been looking through the code. Part of the recent modularization of the gateway code was so we can bundle things like this into addons/plugins instead of the core plugin. The plugin here would basically be the gateway class file and some code to make sure it's included.

I realize that this will give the old gateways a kind of special status, but especially when a gateway can't support all features (trials and recurring billing) it's important to avoid cluttering up our gateway list.

Would you be okay if I moved this code into an addon and offered it through our site?

cryptoapi:
RE > Would you be okay if I moved this code into an addon and offered it through our site?
Yes, it is ok. Please make it available for lite and pro users. Our plugin have GPL licence :).
We can also publish it as addon on wordress.org

----
Please update your two main core files -
1) adminpages/paymentsettings.php
2) preheaders/confirmation.php

as per 
https://github.com/cryptoapi/paid-memberships-pro/commit/c2290fa0e1414bfc80c2c5f0caff76d954f71d58
----

Without that updates, our gateway cannot work properly as separate paidmembershipspro addon.

Thank you

cryptoapi:
Hi,

We have published bitcoin gateway on wordpress.org  -  https://wordpress.org/plugins/gourl-bitcoin-paid-memberships-pro/

Please can you add in your plugin file  "/paid-memberships-pro/preheaders/confirmation.php" in line 9 code -  && pmpro_getOption("gateway") != "gourl" 

Here description - 
https://gourl.io/bitcoin-payments-paid-memberships-pro.html#notes

cryptoapi:
Published as separate bitcoin gateway plugin for pmpro - https://wordpress.org/plugins/gourl-bitcoin-paid-memberships-pro/

PR #190, opened 2015-03-26
BUG: fixed pmpro_formatPrice function which casted PHP notices everywhere

toomuchdesign:
When calling number_format function we check if
$pmpro_currencies[$pmpro_currency]['decimal_separator'] and
$pmpro_currencies[$pmpro_currency]['thosuands_separator'] values are set
before passing them as arguments.


strangerstudios:
Going into 1.8.2.2. Thanks!

PR #205, opened 2015-04-23
Add support GoUrl Bitcoin Gateway plugin

cryptoapi:
Add support Bitcoin Gateway plugin - https://wordpress.org/plugins/gourl-bitcoin-paid-memberships-pro/

Two small changes only.

### DIFF
--- a/adminpages/paymentsettings.php
+++ b/adminpages/paymentsettings.php
@@ -148,7 +148,7 @@
 					Currency and Tax Settings
 				</td>
 			</tr>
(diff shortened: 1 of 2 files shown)


strangerstudios:
Merged in, but changed a bit. To prevent the need for future plugins to tweak this I put gourl in an array of gateways that allow access to the confirmation page early (i.e. have a pending status) and added a filter so others can use the filter.

See here: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/preheaders/confirmation.php#L8-L17

Thanks again for the gateway addon. I should have this version of PMPro pushed out today and will blog about the GoURL/Bitcoin plugin later this week.

PR #207, opened 2015-04-30
Update french translation

yannkozon:


strangerstudios:
Thanks for this!

I like the type="email" stuff, but it can break in older browsers and tends to mess up design stuff. I haven't done testing recently, so maybe it's time to make the jump. Perhaps we can add it in via JS using an HTML5 test function for support?

Otherwise translation stuff looks good and I would commit this without the html5 email.

yannkozon:
If you use type="email" with older browsers like IE6-7-8, the email type isn't recognized and it takes default value : type="text", but I can add support detection.

strangerstudios:
I realize the fields still work like that, but theme CSS often specifically mentions type=text/etc... or at least a year or so ago when I last checked into this, there were a number of themes that would misstyle the email fields upsetting folks.

So we're kind of holding back here waiting for the theme CSS to update with the times.

strangerstudios:
Hmmm an option would be to change the fields to email type but use a filter like pmpro_email_field_type to allow folks to override this to have their themes compatible/etc.

strangerstudios:
I merged in the i18n stuff and the new FR translation. I left out the checkout.php changes. I need to think about it.

I'm open to the type="email" change, we'll just want a filter for easy fixes for themes that break design with that. (The real answer is to update the theme CSS, but that's different for each theme easy for us to have a one liner for folks.)

Also, I think the change the * is good, but there are gists floating around to move the * to the left/right that rely on the span there. Perhaps we can put the span around the abbr.

Thanks!

PR #212, opened 2015-05-20
Add filter and accessibility

yannkozon:
Add filter 'pmpro_email_field_type' for email type field and add acronym for asterisk with span - required field


strangerstudios:
Hi, man. Thanks for looking into this. I'm going to merge this in. I like the email tag stuff. It has the filter for people to fall back if they need to. I thought it might make sense to have that filter return "email" or "text" instead of true or false, but really what other field type are they going to make the email field. So your code is good.

I'm going to look into the abbr to acronym change. Why did you choose that. It my brief look into it, it seems like abbr "won" the browser war here. Do IE browsers still choke on it though? And why add the span wrapper? Will need to make sure our styles work with it and that the register helper is updated to add * in the same way.

Thanks again. These changes should make it into the 1.8.4.5 release.

PR #220, opened 2015-05-24
Use isset, add exit

### LINKED ISSUE #216: PHP notice when clicking cancel
If a user happens to find and click the `Membership Cancel` page while logged out, this notice is displayed:

    PHP Notice:  Trying to get property of non-object in /wp-content/plugins/paid-memberships-pro/pages/cancel.php on line 44

Latest PMP + WP

sumobi:
Fixes #216

`empty($current_user->membership_level->ID)` returns null since it doesn't exist in `$current_user` when a user is logged out. 

This checks if it's set and if not, redirects the user to the membership levels page as the code expects.

### DIFF
--- a/preheaders/cancel.php
+++ b/preheaders/cancel.php
@@ -7,10 +7,10 @@
 	if($current_user->ID)
 		$current_user->membership_level = pmpro_getMembershipLevelForUser($current_user->ID);
 
-	//if they don't have a membership, send them back to the subscription page
-	if(empty($current_user->membership_level->ID)) {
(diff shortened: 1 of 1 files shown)


strangerstudios:
Thanks. Going to merge this. Might change it to use pmpro_hasMembershipLevel() function which will catch logged out users as well. Going to put the inline comment back in too. Any reason you took that out?

sumobi:
@strangerstudios Oops didn't mean to delete the comment :)

PR #225, opened 2015-06-23
Add extra check for shortcode overrides when the child theme is active 

seventhqueen:
Please also apply this extra check for the case when the child theme is active but it does not have a plugin template override. Thanks


strangerstudios:
Yeah, this totally makes sense and I actually thought were doing this already. Thanks for the pull request. It will be included in next version push.

PR #236, opened 2015-08-12
Removed confirmation message in email placeholder

messica:


strangerstudios:
Thanks! Had it done already though.

PR #242, opened 2015-09-14
Remove TML style conflicts

adamhurlebaus:
With the new Theme My Login (currently in beta) and Wordpress 4.3 updates, this edit fixes style issues with the front end generate password fields.  See bug here: https://github.com/jfarthing84/theme-my-login/issues/14

### DIFF
--- a/css/frontend.css
+++ b/css/frontend.css
@@ -177,9 +177,6 @@ li.pmpro_more {list-style-type: none; text-align: center; margin-left: -20px; pa
 
 .pmpro_small {font-size: .8em;}
 
-/* Fixes for Twentyfiften and other themes. */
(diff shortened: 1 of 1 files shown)


strangerstudios:
Thanks for the heads up. It's possible that CSS is needed in some other cases. We endorse TML though, so I'll try to make an update that accounts for the bug you found.

strangerstudios:
Thanks. We're going to remove that line. It actually only fixed issues caused by asterisks added by Register Helper, so we're going to fix Register Helper to not show those asterisks.

There may be some themes that look wonky with 100% width, but people seem to like them so we'll leave this and help people with custom CSS if they have issues from that. Thanks!

PR #252, opened 2015-10-20
Ensure the object used for body class is of type post

andrejcremoznik:
The function `pmpro_body_classes()` returnes a notice error on archives because in that case `$post` object does not have the `ID` property. On taxonomy archives there's `term_id` instead of `ID` for instance.


strangerstudios:
Thanks!

PR #255, opened 2015-10-26
Now expiring members before sending expiration warnings.

messica:


strangerstudios:
Changed the level meta stuff a bit and am going to disable it for 1.8.6.5 but plan to turn it on in 1.8.7

PR #259, opened 2015-11-16
cleared the bug undefined constant host

pranjithkumar:
cleared the bug while install
error throne in debug.log

PHP Notice:  Use of undefined constant host - assumed 'host'


strangerstudios:
Thanks for the pull request, but this has been cleared up already. Make sure to pull the latest PMPro updates upstream.

https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/upgradecheck.php#L462

PR #260, opened 2015-11-20
Do not simply change all email content type to text/html because this…

Mrnt:
… causes some text emails (eg password reset) to get their formatting munged.

What happens is that PMP does not always replace \r\n with `<p>` or `<br>` to format it as an HTML email and the email is passed to PHPMailer as content type text/html so it strips out the \r\n with the result that paragraphs run together. This is happening with password reset emails presumably because they include a url in the message that is wrapped in angle brackets, e.g.:

```
some message text
<http://wp.devmachine.internal/resetpass/?key=wJ52260jeJqyPTvMvjWN&login=admin> 
```

which fools the test on line 111

Instead we should always convert emails with content type text/plain to HTML in PMP.

### DIFF
--- a/includes/email.php
+++ b/includes/email.php
@@ -62,13 +62,20 @@ function pmpro_send_html( $phpmailer ) {
 	
 	//to check if we should wpautop later
 	$original_body = $phpmailer->Body;
-	
+
 	// Set the original plain text message
 	$phpmailer->AltBody = wp_specialchars_decode($phpmailer->Body, ENT_QUOTES);
-	// Clean < and > around text links in WP 3.1
-	$phpmailer->Body = preg_replace('#<(http://[^*]+)>#', '$1', $phpmailer->Body);
-	// Convert line breaks & make links clickable
-	$phpmailer->Body = make_clickable ($phpmailer->Body);
+
+	if ($phpmailer->ContentType == 'text/plain')
+	{
+		// Convert text to HTML
+		// Clean < and > around text links in WP 3.1
+		$phpmailer->Body = preg_replace('#<(http://[^*]+)>#', '$1', $phpmailer->Body);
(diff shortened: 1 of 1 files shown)


strangerstudios:
Hi guys. I will make some effort to work through this. I agree that issues do arrive. A lot of our users do however expect all emails to be turned into HTML emails when they use this feature of the plugin. Similar to how we have the email name/address override optional for PMPro stuff only or all emails, we can make the HTML formatting optional for PMPro stuff only or all emails.

We also can just make this work better with default WP emails. Although there are issues when other plugins that affect email (like WooCommerce) get involved.

Mrnt:
Hi there. Thanks for the feedback. Have you had any additional thoughts? 

While I understand some users may expect all emails to be handled as html, the option in the plugin only indicates that it will override the addressing of the email globally vs just for PMP. It does not indicate that it will automatically handle all emails as HTML (even if it is not correctly formatted as HTML).

I feel it is rather bad form to globally convert the email format and munge messages as a result, which is what the proposed fix was attempting to address - ie all text emails get converted to html, but more robustly than at present.

Maybe there should be an additional option in the email section of the plugin that allows the user to indicate whether all email should be handled as HTML. It probably should only be available if the user unchecks the 'Only Filter PMPro Emails?' option and if enabled then enables the code that converts text mails to HTML (which needs to be more robust).

clifgriffin:
+100

This is absolutely an essential change. A complex website has many sources of emails, and it's not possible to assume they can all be converted to HTML without any problems.

strangerstudios:
I made the html filter stuff smarter in a recent update: https://github.com/strangerstudios/paid-memberships-pro/commit/e6e7322870fed47649881640e2b7f42fee4b64d5

Let me know if this fixes things for y'all.

PR #261, opened 2015-11-24
Fix email content type change affecting all WP email

andrejcremoznik:
Sending all WP email as `text/html` breaks some things. For instance the password reset email that has no paragraph tags and doesn't get filtered by `wpautop` looks terrible as a oneliner.

Plugins should scope such functionality.


andrejcremoznik:
Ha, I didn't even notice there's already a [PR](https://github.com/strangerstudios/paid-memberships-pro/pull/260). That one is probably better.

PR #265, opened 2015-12-10
Avoid warning on "Add new order" in admin area

eighty20results:
User gets PHP warning message because MemberOrder::__construct() returns boolean, and not an actual stdClass() object when creating new MemberOrder. IMHO, MemberOrder::__construct() should always return a true MemberOrder object. However, considering the number of places where MemberOrder is instantiated without an argument (and as a result will return a boolean true, rather than an actual object), there may be compatibility reasons to _not_ do so. This will work around the warning in the page instead.

### DIFF
--- a/adminpages/orders.php
+++ b/adminpages/orders.php
@@ -303,7 +303,15 @@
 			else
 			{
 				$order = new MemberOrder();			//new order
-
+				
+				// empty order object?
+				if (true === $order) 
+				{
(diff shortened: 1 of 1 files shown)


strangerstudios:
I agree the fix should be in the member order class. I think we can initiate a blank object and still return true. So add this to the else state in new MemberOrder:

    $this = new stdClass();
    return true;

What do you think?

PR #289, opened 2016-01-31
cleared the bug undefined constant host

pranjithkumar:
cleared the bug while install
error throne in debug.log

PHP Notice:  Use of undefined constant host - assumed 'host'


strangerstudios:
Not sure when, but this was already fixed in the latest version. Going to close the pull request. Thanks for helping though. Let me know if I missed something.

https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/upgradecheck.php#L581

PR #290, opened 2016-02-03
ENH: Refactored & enhanced capability management

eighty20results:
Moved capability add/remove to own functionality file & added function to add/remove capabilities to a role. 
Flushes capabilities if admin user logs in to admin pages & doesn't have correct PMPro capabilities yet.


eighty20results:
Also snuck in updated Norwegian translation

eighty20results:
Proposed change to memberslist-csv.php - Goal: Reduce consumed memory and # of times we hit the DB for the export operation(s)

eighty20results:
Adding a couple of `unset()` calls to free memory as we go.

eighty20results:
And then some cleanup

eighty20results:
Fix for Stripe Webhook where it looks like the amount variable is not present in more recent version of their API

eighty20results:
Discount codes have to be alphanumeric or contain the '_' character to be "supported", so we'll enforce it.

eighty20results:
Rewrote the rewrite of memberslist-csv.php. Added memory & time benchmarks to allow testing of various `pmpro_set_max_user_per_export_loop` filter settings (default is 2000). See comments in source for tests on a MacBook Pro w/SSD & 16GB RAM (PHP memory_limit = 256MB during tests)

PR #309, opened 2016-03-29
FIX: Didn't include add-ons in check for access

eighty20results:
Was possible that an add-on would have denied access to the content for the current user, but since we never check we don't know.

### DIFF
--- a/shortcodes/membership.php
+++ b/shortcodes/membership.php
@@ -79,6 +79,9 @@ function pmpro_shortcode_membership($atts, $content=null, $code="")
 		}
 	}
 
+	// force check against all other access add-ons & filters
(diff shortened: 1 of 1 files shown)


strangerstudios:
I fixed the same issue (I think) by this update to pmpro:
https://github.com/strangerstudios/paid-memberships-pro/commit/9920486ee0c092fe1498a3006e570e97136c6ec2

And this one to pmpro-email-confirmation:
https://github.com/strangerstudios/pmpro-email-confirmation/commit/5d84b02f8303cb7785b4fddb50e5c84ed8779b84

Only updated on GitHub for now. I hope to push these both out live tomorrow (4/1/2016). No foolin.

PR #310, opened 2016-03-30
using startdate to calculate enddate

tracend:
I had a problem using the plugin when the start date was in the future, so the end date that was calculated, based on the current time, appeared to be wrong. 

This small fix addresses this case.

### DIFF
--- a/preheaders/checkout.php
+++ b/preheaders/checkout.php
@@ -616,9 +616,12 @@
 		{
 			do_action('pmpro_checkout_before_change_membership_level', $user_id, $morder);
 
+			//set the start date to NOW() but allow filters
+			$startdate = apply_filters("pmpro_checkout_start_date", "'" . current_time("mysql") . "'", $user_id, $pmpro_level);
+
 			//calculate the end date
 			if (!empty($pmpro_level->expiration_number)) {
-				$enddate = "'" . date("Y-m-d", strtotime("+ " . $pmpro_level->expiration_number . " " . $pmpro_level->expiration_period, current_time("timestamp"))) . "'";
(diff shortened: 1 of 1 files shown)


strangerstudios:
Hmmm. This makes some sense, but there are also sites relying on the old behavior. A common use case for filtering the start date is to add a 30 day trial to a 1 year long membership. Typically, it's intended that the end date remain 1 year from checkout vs 13 months from checkout.

How are you adjusting the start date? Maybe you can accomplish what you need via a filter already in the plugin or one we can add. We have this code in core already that "extends" a membership vs having it end 1 period from checkout if checking out for the same level with an end date:
https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/filters.php#L10

We have some similar gists that do this always or for recurring payments as well.

tracend:
My use case is a bit different. 

To be specific, this site is for joining training sessions that run in weekly intervals. So if you sign up online any day of the week you will be bumped to the next session on a specific day of the week. 

In this case the end date (after a set number of weeks) needs to be calculated based on the start date, not the purchase date. 

Any help how to achieve this with the existing plugin appreciated.

strangerstudios:
I added a new filter pmpro_checkout_end_date that works similar to the pmpro_checkout_start_date filter. This will go out in 1.8.9 today. There was a way to do this using the pmpro_checkout_level filter to basically change the expiration from 1 week to however many days it is until next next Monday, but this is cleaner.
https://github.com/strangerstudios/paid-memberships-pro/commit/0ef1925b2aae67d7e11eeefd82fb73f39171221e

And here is an example using similar to what you need I think:
https://gist.github.com/strangerstudios/d2e1144158f840fceaba5bbbad5f01fc

tracend:
Looks good. Thanks for addressing this. Feel free to close this pull request when ready.

PR #317, opened 2016-04-13
Update ipnhandler.php

pbaylies:
Updated URL and post behavior for IPN verification, see here: https://www.paypal.com/cgi-bin/webscr?cmd=p/acc/ipn-info-outside


strangerstudios:
Going to test further, but I'll merge this. Thanks!

PR #323, opened 2016-05-01
FIX: Would sometimes throw uncaught exception error 

eighty20results:
FIX: Would sometimes throw uncaught exception error during processing of webhook event.


eighty20results:
Order export includes WP Style formatting changes

PR #327, opened 2016-05-16
Enhanced: Added State search filter

contemplate:
- ability to filter by State
- helpful for calculating state sales tax

### DIFF
--- a/adminpages/reports/sales.php
+++ b/adminpages/reports/sales.php
@@ -108,6 +108,11 @@ function pmpro_report_sales_page()
 		$l = intval($_REQUEST['level']);
 	else
 		$l = "";
+		
+	if(isset($_REQUEST['b_state']))
+		$s = strtoupper($_REQUEST['b_state']);
+	else
+		$s = "";		
 	
 	//calculate start date and how to group dates returned from DB
 	if($period == "daily")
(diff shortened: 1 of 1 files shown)


strangerstudios:
I like the idea, but I think it would work better as an addon. Or maybe a code gist that allowed you to do a search like bstate:PA. Are you a PMPro member? If you post to the forums, we can help you to work through that.

contemplate:
Yep I'm a member!  Yeah I had tried to just make a copy of the Sales & Rev report for an extend plugin of sorts but then things got too complicated.  I'll post to the forum.

adminpages/reports/sales.php
```diff
@@ -108,6 +108,11 @@ function pmpro_report_sales_page()
 		$l = intval($_REQUEST['level']);
 	else
 		$l = "";
+		
+	if(isset($_REQUEST['b_state']))
+		$s = strtoupper($_REQUEST['b_state']);
```
sc0ttkclark, line 113:
```suggestion
		$s = sanitize_text_field( strtoupper( $_REQUEST['b_state'] ) );
```

sc0ttkclark:
Closing based on previous comment from Jason.

PR #337, opened 2016-05-30
Multiple commits

messica:
…dson checkout review page.


strangerstudios:
This makes some sense. I've been individually updating a lot of addons that use these filters to hide their content when on the review page. If we make this change, it does that in one place vs several. However, I need to review the addons to see if removing these hooks would adversely affect any addon. i.e. do any of them require code to be executed or content to be output here.

PR #349, opened 2016-07-19
FIX: Didn't respect "pmpro_include_billing_address_fields" filter

eighty20results:


strangerstudios:
I think all of these fixes have been committed via other pull requests/etc.

eighty20results:
PayPal IPN handler returns 0.00 as the paid amount needs to be merged

eighty20results:
Added action hook to let user(s) post-process a given IPN event. Can be used to replay IPN events and only save orders for IPNs that were not previously correctly saved (incorrect amounts, for instance).

PR #356, opened 2016-08-03
Checkout Preheader for MMPU (plus related function includes)

Square-Lines:
Mostly, this is a reworking of the preheader for the checkout page to
accommodate MMPU. But to support it, a new function needed to be added
- pmpro_areLevelsFree - that checks if multiple levels are all free;
and the pmpro_checkDiscountCode function was also reworked to
accommodate MMPU in addition to the current static level call.

There are some unresolved issues with the checkout preheader still:

- current_user->membership_level is set on line 6 (for no reason?) but
then on line 826. This is problematic both for setting, but also later
for getting.

- I left a note on line 61 for Jason to check on.

- We’ll still need to modify pmpro_changeMembershipLevel a little bit,
but can get that next.

- Emails are still oogly; they will send once per level purchased,
which isn’t what most purchasers are expecting, I think. Probably need
to move those calls from line 829 and put them in their own conditional
block around line 854.

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -39,7 +39,6 @@
 			$order->subtotal = "";
 			$order->tax = "";
 			$order->couponamount = "";
-			$order->checkout_id = "";
 			$order->total = "";
 			$order->payment_type = "";
 			$order->cardtype = "";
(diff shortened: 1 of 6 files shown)


strangerstudios:
Ironically (based on the pull request title) merged in everything minus the preheaders/checkout.php per our discussion.

PR #358, opened 2016-08-10
Update checkout.php

smeric:
applying filter "the_content" to the level description breaks my site on line  on line 56 :)
I think that's because this page is included in a shortcode used inside a page content on which the filter is already applyed ?
not sure, but replacing it with wpautop seems to do the trick !

### DIFF
--- a/pages/checkout.php
+++ b/pages/checkout.php
@@ -53,7 +53,7 @@
 
 				<?php
 					if(!empty($pmpro_level->description))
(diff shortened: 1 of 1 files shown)


strangerstudios:
Similar to the request on the confirmation page, you can now filter pmpro_level_description to apply wpautop to it and use this code to remove the the_content filter we setup for the descriptions:

https://gist.github.com/strangerstudios/958fc314d225a7e4b2727620a2449afd

PR #359, opened 2016-08-12
Update confirmation.php

smeric:
see pull request #358 as this is the same probleme here..

### DIFF
--- a/pages/confirmation.php
+++ b/pages/confirmation.php
@@ -33,7 +33,7 @@
 		
 		$confirmation_message = apply_filters("pmpro_confirmation_message", $confirmation_message, $pmpro_invoice);				
 		
-		echo apply_filters("the_content", $confirmation_message);		
+		echo wpautop( $confirmation_message );
 	?>
 	
 	


strangerstudios:
I believe that there are some users who are expecting the confirmation message and level description to process shortcodes/etc and turning it off entirely would upset them or cause issues on their site.

That said, while working through this I did experience issues when putting short codes inside of the confirmation message and/or level description. My checkout page was halting at random intervals. It was a strange issue, and actually while trying to track it down it mysteriously stopped happening. Really disconcerning.

In any case, here is my attempt to allow for filtering of these strings while also allowing you to disable that in a way that will not break when you upgrade pmpro:

https://github.com/strangerstudios/paid-memberships-pro/commit/ea460d40d6d929bcd26bd29f7a7a5716276f389c

And here is a gist to disable the new functions I added to add the_content filters to the strings and use wpautop instead:
https://gist.github.com/strangerstudios/958fc314d225a7e4b2727620a2449afd

I may push this out later today or this weekend. Let me know if you see any issues with this. Thanks!

strangerstudios:
The filters I added were pushed out. Closing this issue, but if you are still having problems (after using the gists linked to) let us know in a new issue.

PR #372, opened 2016-09-26
ENHANCEMENT: wpdb->prepare() for membershiplevels.php

andrewlimaza:
added SQL statements with $wpdb->prepare() 

Please feel free to comment if I have missed anything


eighty20results:
This page could be a bit simpler if we used `$wpdb->replace()` - see: https://codex.wordpress.org/Class_Reference/wpdb#REPLACE_row

strangerstudios:
Changed to use REPLACE (Thanks, Thomas) and merged in.

PR #380, opened 2016-10-05
Shift Hook : pmpro_before_change_membership_level

MrVibe:
The hook pmpro_before_change_membership_level should be placed before the membership levels are deactivated. So that the $level_id returns the old_level of the user.
The current implementation of the hook returns the new level id in the parameter passed which makes it impossible to detect the old membership level when a user changes/switches the membership.


strangerstudios:
I think this is a good idea. In some other cases similar to this, we would have code that specified a cancelled/etc status when getting the $order->getLastMemberOrder($user_id, 'cancelled') or something similar. This is a new hook, so I don't believe it's being used that way. I just want to make sure I don't break anything by merging this in, but I will probably include this in the next update. Thanks for the contribution.

PR #382, opened 2016-10-05
BUG/ENHANCEMENT: Only load PMPro JS file when needed

eighty20results:


eighty20results:
Updated pages/levels.php to reflect the fact that sorted levels & levels get loaded by preheaders/levels.php.

PR #383, opened 2016-10-06
ENHANCEMENT: Now only showing Update Billing on Account Page for recurring levels

greathmaster:


strangerstudios:
This is good. We should probably add a filter as well so new gateways can be included in the list of gateways we show this for.

PR #390, opened 2016-10-19
BUG: Returns invalid variable name in setGateway()

eighty20results:


strangerstudios:
I'm pretty sure that we set ->gateway to a string of the gateway name and ->Gateway (caps) to the gateway object. Is there another way to avoid the variable name warning?

eighty20results:
Maybe move the return to L176?

strangerstudios:
I think instead we need an else condition here to return false if there is no gateway var passed in/etc. https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/class.memberorder.php#L171-L172

strangerstudios:
Fixed by this commit I think: https://github.com/strangerstudios/paid-memberships-pro/commit/739ed8657a7a13a21dfc480d4e6bad14091a956d

PR #403, opened 2016-11-29
ENHANCEMENT: Filtered display of "Choose Payment Method" area on checkout page

eighty20results:


strangerstudios:
Merged in. I'm going to review the profile timezone/etc updates. I will at least remove the debug stuff, but might tweak the timezone set/reset as well.

PR #405, opened 2016-11-30
BUG: PayPal IPN Handler - Invalid attempt to cast inside empty()

eighty20results:


strangerstudios:
I couldn't recreate the error. It might have to do with certain PHP versions or other conditions. But since empty('0.00') evaluates to true anyway, I merged it in with confidence. Thanks.

Issue #326, opened 2016-05-14
Search at Discount page isn't working


capripio:
Hey,
Search textbox isn't working in discount coupon page, its seems you have added query for search, but variable "$s" isn't defined. for now I've added following code:

```
if(isset($_REQUEST['s']))
    $s = $_REQUEST['s'];
```

at "discountcodes.php"

PR #410, opened 2016-12-08
fix for #326: Search at Discount page isn't working

dkd903:
A very minor fix.


strangerstudios:
Thanks for your patches!

PR #413, opened 2016-12-20
Placeholder account number (X's) should be 12 not 13 in length

rafecolton:
Otherwise it makes it look like the credit card number is 17 digits instead of 16


strangerstudios:
lol nice catch. merging into the update going out right now.

PR #414, opened 2016-12-22
BUG: Fixed expiration date issue on Edit User page when timezone set …

greathmaster:
…to UTC+0. Expiration date would show as the current date. WordPress option "timezone_string" is empty when UTC+0.


eighty20results:
There's another issue... :( The `date_default_timezone_set()` function doesn't handle UTC offset timezones.

PR #417, opened 2017-01-03
Fixed Braintree payments

ttshivers:
Braintree payments do not work unless you use these new non-deprecated functions. Found this fix at: http://doginthehat.com.au/2015/03/deprecated-error-with-lemonstand-v1-and-braintree-payments/

### DIFF
--- a/includes/lib/Braintree/Braintree/Util.php
+++ b/includes/lib/Braintree/Braintree/Util.php
@@ -136,7 +136,9 @@ class Braintree_Util
      */
     public static function delimiterToCamelCase($string, $delimiter = '[\-\_]')
     {
-        return preg_replace('/' . $delimiter . '(\w)/e', 'strtoupper("$1")',$string);
+        return preg_replace_callback('/' . $delimiter . '(\w)/',function($matches) {
+            return strtoupper($matches[1]);
+        }, $string);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Manatee-Outlaw:
This solution actually solved a long-running problem my company had with Braintree. Neither one of us knew what was going on!

ttshivers:
I ended up figuring out this solution after I got these errors:
```
PHP Warning:  preg_replace(): The /e modifier is no longer supported, use preg_replace_callback instead in /var/www/html/wp-content/plugins/paid-memberships-pro/includes/lib/Braintree/Braintree/Util.php on line 164
PHP Warning:  XMLWriter::startElement(): Invalid Element Name in /var/www/html/wp-content/plugins/paid-memberships-pro/includes/lib/Braintree/Braintree/Xml/Generator.php on line 71
```

eighty20results:
It's ultimately up to Jason, but updating the actual library to a more recent version where Braintree directly supports PHP7 seems like a more clean and future proof solution than taking on the support burden of their library with our resources. But that's just my opinion... :)

ttshivers:
Sure, I agree that updating the library is a much better solution, but the way it is right now, Braintree doesn't work with pmpro at all with PHP7, so until you do update the library, this might be an acceptable temporary fix.

strangerstudios:
Thanks for this! I'm going to merge and test right now. Will have it pushed out ASAP.

I agree that we need to move to PHP7 soon, but in the meantime we'll have to maintain this old library.

jbruggeling:
Is there any timeline when you will move to a new library? Would be awesome to pass the custom string through the braintree api so users can set the company / website name that is used on the credit card statement.

jbruggeling:
Also is there anything else that will break if i turn on php 7?

ttshivers:
I am using php 7 and everything else is working fine.

PR #418, opened 2017-01-10
BUG: Cancellations showing 0 for All Time under Membership report

contemplate:
The last bug fix didn't actually fix the issue.  This fixes it for realz!

### DIFF
--- a/adminpages/reports/memberships.php
+++ b/adminpages/reports/memberships.php
@@ -224,8 +224,7 @@ function pmpro_report_memberships_page()
 		else
 			$sqlQuery .= "WHERE mu1.status IN('inactive','expired','cancelled','cancelled_admin') ";
 			
-		$sqlQuery .= "AND mu2.id IS NULL 
-		AND mu1.startdate >= '" . $startdate . "' 
+		$sqlQuery .= "AND mu1.startdate >= '" . $startdate . "' 
 		AND mu1.startdate < '" . $enddate . "' ";
 		 
 		//restrict by level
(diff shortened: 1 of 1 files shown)


strangerstudios:
Thanks for the pull request! I think we can merge it as is, but had a question at one point: https://github.com/strangerstudios/paid-memberships-pro/pull/418/files#r96014772

contemplate:
Great! Maybe I'm missing something.  What is the question? :)

strangerstudios:
There was a line in the query that you removed:

    $sqlQuery .= "AND mu2.id IS NULL 

This the query left joins onto mu2 and checks to see if the user with an entry with status "cancelled"/etc has a newer entry with status "active" that was created within 24 hours of the cancellation... basically showing that the user upgraded/downgraded instead of cancelled.

This isn't really necessary anymore since we have specific admin_changed and changed status for when this happens, but older sites (before 2015 or so) will have users who upgraded and would be counted as cancels without this. That might be okay.

If this is confusing to you, I'll noodle that over and add it back in or keep it out and merge the rest of your fixes.

contemplate:
Actually that exact line directly fixed the Zero cancellations issue on two site we built in 2015 & 2016.  I remember months ago when I was debugging this and manually running this query through PHPMYADMIN with user ids I know that should show up as cancelled that it was produce no results.  I removed that and it showed the correct results.  I hadn't thought about the situation you are describing.

strangerstudios:
I'm going to merge this in with the reverse compatibility stuff out. I'll also remove the related left join in the query. Thanks!

Issue #422, opened 2017-01-27
Apply PHP_CodeSniffer WordPress Coding Standards


salcode:
See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards

After PHP_CodeSniffer and these coding standards are installed, running the following from the command line

```
phpcbf --extensions=php ./paid-memberships-pro/ --standard=WordPress-Core -p
```

will apply the changes PHP_CodeSniffer can automatically apply to the php files.

`WordPress-Core` is a sub-set of all the available WordPress coding standards, see https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards#standards-subsets for the full list.

dparker1005:
Thanks for opening this. I’m going to close this older cleanup issue because we now have a `phpcs.xml.dist` ruleset in the repo using WordPress Coding Standards, and our contributor docs ask new changes to follow those standards.

There is still older code that does not fully match WPCS, but keeping one broad issue open for a full-codebase formatting pass is not very actionable. Future work here is better handled as smaller cleanup PRs or a specific issue for PHPCS tooling/CI.

PR #423, opened 2017-01-27
Apply PHP WordPress Coding Standards

salcode:
These changes were generated with the command

```
phpcbf --extensions=php ./paid-memberships-pro/ --standard=WordPress-Core -p
```

Since these changes were auto-generated, I'd recommend reviewing them
before applying them.

See #422


strangerstudios:
Neato. I'll try to find a weekend to review everything... and maybe delay this a bit to not interfere with other open pull requests (which will cause major merge issues). But thanks for sharing that command.

ideadude:
We're now applying WPCS on a case by case basis as we update and refactor old code.

PR #426, opened 2017-02-01
Add filter for modifying the order description in Stripe

rafecolton:
Will enable users to add unique identifiers or otherwise modify order descriptions in order to aid accounting departments in differentiating these payments from other lines of business paid through the same gateway


strangerstudios:
Thanks! I like this and will squeeze into the 1.8.13.6 update going out today.

PR #441, opened 2017-03-18
ikimmybee

ikimmybee:
Requesting pull.


strangerstudios:
The v2.0 branch is wildly out of date. We keep it around for reference. Everything that was being developed there has found its way into v1.8 and/or the MMPU plugin here: https://github.com/strangerstudios/pmpro-multiple-memberships-per-user

I might remove it from the repo here because it's confusing, but before I do. Was there a reason you were using this branch over the master or dev branch? Or maybe you just thought it was more current (understandable with the version numbers).

PR #450, opened 2017-04-26
ENH: Incorrectly attempting to translate variables

eighty20results:


strangerstudios:
These values shouldn't be translated at all... or they should be translated when added to the filter. I'll pull this in and tweak a bit more. Thanks.

strangerstudios:
https://github.com/strangerstudios/paid-memberships-pro/commit/396fd8d6fd34434aa130bb2ed2d74b48335516fe

PR #456, opened 2017-05-07
BUG/FIX: Wouldn't allow cancellation

eighty20results:
If a member attempted to cancel, the level check would incorrectly return false if the member already had the membership level.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -974,7 +974,7 @@ function pmpro_changeMembershipLevel($level, $user_id = NULL, $old_level_status 
 	if(!is_array($level))
 	{
 		//are they even changing?
-		if(pmpro_hasMembershipLevel($level, $user_id)) {
+		if(!pmpro_hasMembershipLevel($level, $user_id)) {
 			$pmpro_error = __("not changing?", 'paid-memberships-pro' );
 			return false; //not changing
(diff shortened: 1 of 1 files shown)


planetjon:
Are you going to fix this up? I'm still blocked by the same issue and have an identical commit in my fork.

ideadude:
No fix is required here.

    // if it's a custom level, they're changing
    if ( ! is_array( $level ) ) {
      // are they even changing?
      if ( pmpro_hasMembershipLevel( $level, $user_id ) ) {
	      $pmpro_error = __( 'not changing?', 'paid-memberships-pro' );
	      return false; // not changing
      }
    }
    
 If $level is not an arary (i.e. it's a level id number) we check if the user already has the level you are changing to, and if so the function bails, returning false.
 
 So if you call pmpro_changeMembershipLevel( 1, 1 ) and the user with ID = 1 already has level one, this function will bail and do nothing. It returns false. It didn't change the level.
 
 Similarly, if you call pmpro_changeMembershipLevel( 0, 1 ) and the user with ID = 1 has no levels, what is there to cancel? We bail and return false.
 
 If this is failing in a specific case, please submit an issue and let us know so we can come up with the appropriate fix.

PR #464, opened 2017-05-23
Ensure discount codes passed in URL is preserved through PayPal review step

rafecolton:
If a site chooses to offer discount codes but declines to use the discount code UI, in order to reduce checkout page friction, it is possible to include `?discount_code=<code>` in the URL. This works in conjunction with hiding the checkout form field by returning `false` from the `pmpro_show_discount_code` filter. 

In this case, the discount code gets lost at the PayPal order review step. This diff ensures that the discount code, as passed back by PayPal in the callback URL, gets submitted with the checkout form at order review.

### DIFF
--- a/pages/checkout.php
+++ b/pages/checkout.php
@@ -16,6 +16,9 @@
 
 	<input type="hidden" id="level" name="level" value="<?php echo esc_attr($pmpro_level->id) ?>" />
 	<input type="hidden" id="checkjavascript" name="checkjavascript" value="1" />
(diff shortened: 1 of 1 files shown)


rafecolton:
Corresponds to [this](https://www.paidmembershipspro.com/forums/topic/discount-code-doesnt-work-w-paypal-express/#post-106873) members support forum issue

strangerstudios:
Thanks a bunch! Merging in and pushing out ASAP

PR #485, opened 2017-08-16
BUG: Level order was not updating from Levels admin page.

greathmaster:


strangerstudios:
Needs sanitization (along with handling an array of integers), but otherwise looks good.

PR #493, opened 2017-09-01
BUG FIX: PayPal not accepting amounts over $1000

andrewlimaza:
BUG FIX: PayPal was not accepting amounts over $1000, changed `number_format` to include decimals.


strangerstudios:
We need to test this with non-decimal currencies (Yen? Rand?) to make sure it still works. I guess we were already doing the number format thing, just need to not include the commas. So that might be a different bug. But good to test anyway.

andrewlimaza:
@strangerstudios I'm testing with a lot of currencies at the moment and so far so good - tested on Rand, Euro and still testing more. So far so good.

andrewlimaza:
The fix forces `,` for thousands and `.` for cents - seems to work. 👍

PR #502, opened 2017-09-16
BUG: Code ID passed to hook pmpro_save_discount_code() was -1 for newly…

greathmaster:
… added codes.


strangerstudios:
Have you tested this? Looking at the code, I'm not sure that $saveid is set for new discount codes. Let's make sure this change doesn't break anything for updating existing codes/etc either.

PR #513, opened 2017-09-21
ENHANCEMENT/FIX: Use AJAX over direct execution

eighty20results:
When displaying Web Hook URL setting, we should do the same for Braintree as we do for all other gateways.


eighty20results:
Just tested against the Braintree sandbox and PMPro 1.9.4 and got a successful response from Braintree (no errors) when using `wp-admin/admin-ajax.php?action=braintree_webhook` rather than direct execution.

ideadude:
I mentioned in Slack, but I'd also like to test this on a live Braintree account. I feel like this could have been an "only on live servers" type error.

Issue #481, opened 2017-07-25
Members List - changes in pmpro_members_list_csv_default_columns are not reflected in CSV header row


mathieuhays:
Removing columns using `pmpro_members_list_csv_default_columns` is not reflected on the first row (Column names).

This results in the exported CSV having values associated with the wrong column heading.

Use-case:
```php
add_filter( 'pmpro_members_list_csv_default_columns', function( $default_columns ) {
	$to_remove = [
		'pmpro_bfirstname',
		'pmpro_blastname',
		'pmpro_baddress1',
		'pmpro_baddress2',
		'pmpro_bcity',
		'pmpro_bstate',
		'pmpro_bzipcode',
		'pmpro_bcountry',
		'pmpro_bphone',
	];

	// Remove billing info
	foreach ( $default_columns as $index => $column ) {
		if ( ! empty( $column[1] ) && in_array( $column[1], $to_remove ) ) {
			unset( $default_columns[ $index ] );
		}
	}

	return $default_columns;
});
```

I understand `pmpro_members_list_csv_heading` can be used as a workaround but it makes more sense to handle everything directly in `pmpro_members_list_csv_default_columns`.

I believe this could be sorted easily having the label as the third index in the column array:
`array( 'theuser', 'user_login', 'Username' )` and then generate the `$csv_file_header` string after the 'default columns' hook.

I'm happy to do a pull request if that can help.

strangerstudios:
Hi, thanks. I agree. It would be nice it there was just one filter and/or it was easier to remove default headings from the CSV export.

In refactoring this, I worry about affecting the existing code that expects an array of strings with numeric keys. Perhaps if we set the keys to the headings instead... this may not break existing code, but our code could recreate the header from the array keys (and then still run the existing filter there). Does that make sense.

I'm open to a patch for this. We'd want to do something similar for the orders export as well, but one step at a time.

PR #517, opened 2017-09-25
Enhance memberlist csv column filtering

mathieuhays:
Implements #481.

I've changed the way the columns array works. I've added the heading as the key (as you mentioned @strangerstudios) and I also added the ability to specify a callback which enabled me to include columns like "expires" or "joined" as filterable columns.

`$extra_columns` are merged with `$default_columns` since the behaviour is now similar.

`pmpro_members_list_csv_heading` still works but will break if someone tries to add/remove columns using this filter.

I've done some tests and we can now remove and/or re-order columns using `pmpro_members_list_csv_default_columns`. Actually it's now easier to remove/re-order columns because we are using the heading as the key.

I haven't managed to test the use-case for "oldmembers".

### DIFF
--- a/adminpages/memberslist-csv.php
+++ b/adminpages/memberslist-csv.php
@@ -85,56 +85,52 @@
 	else
 		$headers[] = 'Content-Disposition: attachment; filename="members_list.csv"';
 
-	//set default CSV file headers, using comma as delimiter
-	$csv_file_header = "id,username,firstname,lastname,email,billing firstname,billing lastname,address1,address2,city,state,zipcode,country,phone,membership,initial payment,fee,term,discount_code_id,discount_code,joined";
-
-	if($l == "oldmembers")
-		$csv_file_header .= ",ended";
-	else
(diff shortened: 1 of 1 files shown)


strangerstudios:
This is super cool, but won't make the 1.9.4 release. I do want to take a deeper look at this before the next release for potential inclusion. I hope it's not too much trouble for you to repatch your site if you are using this already.

sc0ttkclark:
We'll need to update the branch on mathieuhays:csv-header-enhancement with the latest from our current dev branch to move forward here. The changes in the PR look pretty good, I'd only suggest a few tweaks for new hooks (hook docs info which I can help with after PR is refreshed) and possibly a couple of code formatting things.

dparker1005:
Hi there @mathieuhays, thank you again for this work.

I am going to close this PR for now due to inactivity and complex merge conflicts that would need to be resolved. If portions of these changes are still needed, we can open new PRs off of the current dev branch.

PR #524, opened 2017-09-28
Added icon and css

pbrocks:
Adding dashicon to Membership levels to indicate drag and drop of levels. Aslo some CSS to position icons and center table cells. http://take.ms/3YUyH

### DIFF
--- a/adminpages/membershiplevels.php
+++ b/adminpages/membershiplevels.php
@@ -650,7 +650,7 @@
 
 		<?php if(empty($_REQUEST['s']) && count($reordered_levels) > 1) { ?>
 			<br class="clear" />
-		    <p><?php _e('Drag and drop membership levels to reorder them on the Levels page.', 'paid-memberships-pro' ); ?></p>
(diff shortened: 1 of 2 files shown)


strangerstudios:
I feel like the drag icon in every row is an eyesore.

I think the mouse cursor should become a draggable icon when hovering over that row (but links should still be hand pointers). Small issue is that the text cursor icon is nice for copying text out of the rows... but again that's already a little janky with the dragging.

I think we could have the whitespace where the drag icon is in the rows to make it easier to drag. On mouseover it could show a light gray hamburger icon (three or four horizontal lines... or I've see link 4 divots used as a "you can drag this icon").

Trello has a neat rotate the row X degrees when dragging and highlight where the block will be dragged to. Nice for future.

Let's confirm a design with Kim and update this.

pbrocks:
I can agree that the eyecon is and eyesore, but there is something to be said about iconicity. I think it's important to view it thru the lens of someone who doesn't know what's going on. The icon indicates that the row itself has an action associated with it, something that isn't obvious or could be assumed. We can look at different solutions, but I think having something overt, as opposed to on hover, is useful.

PR #533, opened 2017-10-07
BUG FIX: Not casting variable variables to strings

eighty20results:


eighty20results:
FWIW, this is a fix to something that breaks w/PHP 7

PR #534, opened 2017-10-09
Enhancement: Filter for membership shortcode access

andrewlimaza:
Added a filter for [membership] shortcode

### DIFF
--- a/shortcodes/membership.php
+++ b/shortcodes/membership.php
@@ -79,6 +79,9 @@ function pmpro_shortcode_membership($atts, $content=null, $code="")
 		}
 	}
 
+	// Filter the $hasaccess so we can overwrite this for other add ons.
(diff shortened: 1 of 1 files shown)


strangerstudios:
Good to go. Can we comment it slightly better. We should comment all new filters like this:
https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/functions.php#L661-L670

Fix ones we run across and also update the CONTRIBUTE.md to point to this as an example of commenting on a filter (put in a gist in case the line numbers change) and also point to an example of commenting up a function/method... instead of linking to the EDD example. :/

andrewlimaza:
Will update this.

This filter will work good for other add-ons such as PMPro Approvals etc, we can remove access to shortcode only content as well.

PR #537, opened 2017-10-13
Warning Message for enabling recurring and expiration

andrewlimaza:
Show a pop-up alert when a user selects both


eighty20results:
Wouldn't it make sense to stay consistent with how we message/notify on the admin pages?

Everywhere else, we set the pmpro_msg/pmpro_msgt, but now we're suddenly starting to throw modal pop-ups (which quickly become annoying).

andrewlimaza:
maybe.

andrewlimaza:
Can probably do away with the popup and just show the warning text below the 'expiration settings'. This way it is way more clear and shows as the user selects the checkbox.

andrewlimaza:
I removed the code for the `alert`

andrewlimaza:
Resubmitting a new PR for this.

PR #540, opened 2017-10-16
Updates recaptchalib to use POST instead of GET/file_get_contents

corin12355:
Occasionally, on some systems, file_get_contents will not work (returns false or whatever) when being used with Google reCAPTCHA.

This updates PMPro to instead use cURL and uses POST instead, as per the documentation.

### DIFF
--- a/includes/lib/recaptchalib.php
+++ b/includes/lib/recaptchalib.php
@@ -65,24 +65,20 @@ class pmpro_ReCaptcha
     }
 
     /**
-     * Submits an HTTP GET to a reCAPTCHA server.
+     * Submits an HTTP POST to a reCAPTCHA server.
      *
      * @param string $path url path to recaptcha server.
      * @param array  $data array of parameters to be sent.
      *
-     * @return array response
+     * @return string response
      */
-    private function _submitHTTPGet($path, $data)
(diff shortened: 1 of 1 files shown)


eighty20results:
+1

Hope this gets pulled in quickly since `file_get_contents()` is a theoretical attack vector as well.

strangerstudios:
Is there an updated recaptchalib.php (I feel like I was using something official) we could update to?

Could we use the WordPress HTTP API instead (which uses curl internally)?
https://codex.wordpress.org/HTTP_API

eighty20results:
Think we probably need to transition to reCAPTCHA v2 which is fully JS based: https://developers.google.com/recaptcha/docs/faq

corin12355:
Just updated it to use the WP HTTP API, also fixed a few other bugs/error handling.

Personally, reCAPTCHA v2 will likely break a lot of custom PMPro checkout templates.
That'd probably be something for another PR though.

ideadude:
Merged in and going out in the 1.9.4.1 update (very soon). Thanks again! Great code and this process was smooth.

corin12355:
Great stuff! Glad to help :)

PR #558, opened 2017-10-30
Page functions

andrewlimaza:
Added page functions -

* pmpro_post_has_shortcode
* pmpro_is_page
* is_pmpro
* is_pmpro_account
* is_pmpro_billing
* is_pmpro_cancel
* is_pmpro_checkout
* is_pmpro_confirmation
* is_pmpro_invoice
* is_pmpro_levels


eighty20results:
FWIW, you should probably also make sure the 'page_content' object variable is set in https://github.com/strangerstudios/paid-memberships-pro/pull/558/files#diff-f81f3dff36bf04b7a070d642c858e883R2676

eighty20results:
What's the cost/benefit of negating the standard wordpress functions for `is_page()` or `has_shortcode()` with our own implementations? Basically, what are the use cases for those 2 that aren't already covered by the core WP functions? Is there a real a cost/benefit or user value to having them (that makes them stand out as solving a problem not addressable by the pre-existing WordPress functions).

PR #567, opened 2017-11-08
BUG/ENHANCEMENT: Comparison of email and confirm email fields at che…

greathmaster:
…ckout was case sensitive.


ideadude:
@eighty20results I agree. We had another issue to remove the verify fields and should have that ready for 1.9.5. In the meantime, let's fix this check. We'll probably have a filter to re-enable it for folks who still want it.

PR #568, opened 2017-11-09
BUG FIX: wp_unslash() for instructions

eighty20results:
Unslash the check payment instructions before displaying them


ideadude:
Your other PR had this in it as well.

PR #576, opened 2017-11-20
PMPro level information shortcode

andrewlimaza:
Added in pmpro_level_info shortcode so user’s may retrieve information
of a particular level.

Supports all the level fields in the DB table.


andrewlimaza:
@ideadude should we do `[pmpro_level]` or `[pmpro_level_info]`

eighty20results:
From a user experience perspective, `[pmpro_level]` seems way too similar to `[pmpro_levels]` with vastly different expectations on functionality provided.

PR #581, opened 2017-11-22
BUG FIX: Using UNIX_TIMESTAMP() causes unexpected conversion between GMT and local timezone, resulting in loss of time for certain records

eighty20results:
BUG FIX: Using UNIX_TIMESTAMP() causes unexpected conversion between GMT and local timezone, resulting in loss of time for certain records
BUG FIX: JavaScript error for level profile fields (on profile page)
ENHANCEMENT: Prevent more SQL injections (making sure) / Better sanitation for more SQL queries
ENHANCEMENT/FIX: Using table prefix for column names (avoid namespace/command issues in DB)
ENHANCEMENT/FIX: String wasn't really translatable (included HTML)


eighty20results:
Fixed conflict in includes/functions.php and fixed issue with missing enddate value (null) when inserting record

ozsolarwind:
Gidday Thomas

We had some discussions on the topic about two years ago. At the time our (Wordpress) system was located 2 or 3 hours ahead of our local timezone, which caused the problem of decrementing the enddate whenever a member record was updated (even without any change to the date being made). Our servers were then relocated to the local timezone and the problem went away. They've since been re-relocated to 2 - 3 hours ahead and the problem has reappeared. When the problem initially occurred I added the following code to   pmpro_get_membership_level_for_user filter:
`
/*
*  temporary fix to reset enddate when server is in a different timezone
*/
    if (!empty($user_id)) {
        if ($user_id != $current_user->ID) {
            $enddate = $wpdb->get_row("SELECT enddate FROM {$wpdb->pmpro_memberships_users}
                                       WHERE user_id = $user_id AND status = 'active'");
            $enddate = strtotime($enddate->enddate);
            $level->enddate = (string) $enddate;
        }
    }
/*
*  end of temporary fix
*/
`

I removed that change when our servers were relocated to the local timezone. This time around I've added the following code to (my) plugin init which is probably neater (I could even add a test to only do the SET time_zone if there was a difference.

`
/*
* Fix problem with timezone differences between server and local
*/
function sync_timezones() {
// from: https://www.sitepoint.com/synchronize-php-mysql-timezone-configuration/
global $wpdb;
$now = new DateTime();
$mins = $now->getOffset() / 60;
$sgn = ($mins < 0 ? -1 : 1); $mins = abs($mins); $hrs = floor($mins / 60); $mins -= $hrs * 60; $offset = sprintf('%+d:%02d', $hrs*$sgn, $mins); $sqlQuery = "SET time_zone='$offset';"; $ok = $wpdb->get_var($sqlQuery);
return $options;
}
add_action('init', 'sync_timezones');
`

Finally, a summary of the problem (and why; but I guess you know this anyway):

The problem relates to updating membership records when the server the site is hosted on is in an earlier timezone such that whenever you update a member (record) their expiry date decrements by one day. The problem occurs because PMPro saves a membership enddate as a text string such as 'yyyy-mm-dd 00:00:00' without any time value, for example '2019-11-04 00:00:00'. It retrieves the enddate as a unix timestamp from the MySQL server. If the server is in an earlier timezone than the local timezone this returns a local datetime for the day before, for example if the local time was three hours behind in the example above it would return '2019-11-03 21:00:00'. When the member record is saved PMPro detects a changed date and updates the record using just the date components, so it becomes '2019-11-03 00:00:00'. Thus, every time you update the use record their expiry date becomes one less.

Kind regards

Angus

PR #603, opened 2018-01-25
Revert of javascript check

skiningham:
I'm basically reverting the changes made in #144. Does anyone (@ideadude?) remember why it was added? If we assume that javascript must be enabled in order for paid-memberships-pro to function correctly [1], is this really the best way to detect that? In my experience, and the experience of many googlers, this check leads to numerous false positives including some due to errors outside of the scope of paid-memberships-pro.

When it comes to accepting money from my users, the false positives are very painful. Does the plugin still need this check? If yes, should `pmpro_require_javascript_for_checkout` default to `false`?

----------
[1] In my limited testing with PayPal Express, the registration flow works fine without JS. I assume other gateways aren't as accessible.

### DIFF
--- a/pages/checkout.php
+++ b/pages/checkout.php
@@ -15,7 +15,6 @@
 <form id="pmpro_form" class="pmpro_form" action="<?php if(!empty($_REQUEST['review'])) echo pmpro_url("checkout", "?level=" . $pmpro_level->id); ?>" method="post">
 
 	<input type="hidden" id="level" name="level" value="<?php echo esc_attr($pmpro_level->id) ?>" />
-	<input type="hidden" id="checkjavascript" name="checkjavascript" value="1" />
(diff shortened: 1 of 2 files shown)


ideadude:
The problem is that we embed JS into the HTML of the checkout page. If some other JS breaks on the checkout page our JS is not going to run. For some use cases and gateways this is fine. For others, like Stripe, this breaks the checkout completely.

The real solution here is for us to move the JS into a .js file. This way it won't be effected by other JS breaking on the checkout page. Once we do that, we can remove this extra check.

I'll leave this open until we do that (before the end of the year I think).

In the meantime, you can use a custom checkout.php that doesn't do the JS and just has the hidden input always present.

dparker1005:
Closing as the JS checked has been moved into a `.js` file.

PR #604, opened 2018-01-26
BUG: Discount code uses were not being counted with PayPal Standard

greathmaster:


sc0ttkclark (commented):
This should be handled now, the conflict showing is related to new logic that addresses this.

Can we get a confirmation it's working as expected in newer PMPro versions?

dparker1005:
Closing as this has been resolved.

PR #619, opened 2018-02-01
Fix expiry not working with multiple memberships

sebkajeka:
Pmpro with the additional multiple memberships plugin fails to expire users, resulting in daily email spam reminders about an expired membership and a permanent membership. This is due to a missing optional argument in the change membership function in the core pmpro.


ideadude:
Look good! Thanks a lot. I plan to push the next release out tomorrow.

PR #620, opened 2018-02-03
BUG: Billing fields not being required on some gateways

greathmaster:
If the Stripe setting for "Show Billing Address Fields" was set to "No" and saved, but a different gateway such as Braintree was used, the address fields would show as not required. 

This line from classes/gateways/class.pmprogateway_stripe.php ~L130:
`add_filter('pmpro_required_billing_fields', array('PMProGateway_stripe', 'pmpro_required_billing_fields'));			`
was still running since it's loaded on `init()`, even if the gateway was not Stripe.


sc0ttkclark (approved):
This make sense (after resolving the conflicts, moving it still makes sense)

dparker1005:
Closing as this issue has been resolved.

PR #627, opened 2018-02-09
BUG: Membership level id and name were missing from Billing Failure …

greathmaster:
…emails sent to members from webhook


sc0ttkclark (approved):
Looks good but needs eyes from another team member familiar with this area on whether there's a better solution

services/stripe-webhook.php
```diff
@@ -228,6 +228,7 @@
 			$old_order = getOldOrderFromInvoiceEvent($pmpro_stripe_event);
 			$user_id = $old_order->user_id;
 			$user = get_userdata($user_id);
+			$user->membership_level = pmpro_getMembershipLevelForUser($user_id);
```
dparker1005, line 231:
```suggestion
```
This change is no longer needed as the user's membership level is now being retrieved in the sendBillingFailureEmail() method as a result of this PR:
https://github.com/strangerstudios/paid-memberships-pro/pull/1586

dparker1005 (changes requested):
The user->membership_level line is no longer needed, but the $morder->membership_id change looks good.

sc0ttkclark (approved):

PR #633, opened 2018-02-12
BUG: When using PMPro MMPU memberships were not being expired

greathmaster:


sebkajeka:
I already have a pull request for this issue open and the correct usage of the API

https://github.com/strangerstudios/paid-memberships-pro/pull/619

greathmaster:
@sebkajeka pull request is correct. I'm going to close this one as it actually introduces a bug that only cancels id = 1.

PR #635, opened 2018-02-13
Load print.css on pages that use print feature only

andrewlimaza:
Load the print.css file only on pages that use the print feature.

Possibility would be to add this to the frontend.css since the CSS code is one line.

### DIFF
--- a/includes/init.php
+++ b/includes/init.php
@@ -68,7 +68,12 @@ function pmpro_init()
 			$print_css = get_template_directory_uri() . "/paid-memberships-pro/print.css";
 		else
 			$print_css = plugins_url('css/print.css',dirname(__FILE__) );
-		wp_enqueue_style('pmpro_print', $print_css, array(), PMPRO_VERSION, "print");
+
+		// Only load it on pages that are using the print feature.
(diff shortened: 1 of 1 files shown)


andrewlimaza:
Actually think we should check if `isset` instead?

dcervantes:
**_Isset_** just checks if is it set, it could be anything not null. **_Empty_** checks if the variable is set, and if it is, it checks it for null, "", 0, etc.

Both will solve the problem of load in all pages the print.css.

eighty20results:
First of all, enqueuing resources should follow WordPress best practices (ie. should happen in the [admin|wp]_enqueue_scripts hook) and not be done in the `init` action.

In the WP Codex ([in wp-admin](https://codex.wordpress.org/Plugin_API/Action_Reference/admin_enqueue_scripts) there are documented best practices for adding scripts/styles to specific page(s) as they're being loaded for the backend (use the passed $hook variable to identify the page(s))

For the front end, people typically use `is_page()` and the page ID (rather than the request variable that could be a duplicate from some other plugin that also doesn't use the plugin prefix for it's POST/GET variables).

dparker1005:
Closing in favor of merging print.css into frontend.css:
https://github.com/strangerstudios/paid-memberships-pro/pull/2940

PR #636, opened 2018-02-13
Remove Update button on billing field for some gateways.

andrewlimaza:
This removes the Update/Cancel button on the billing page for payment gateways that do not capture billing details such as PayPal etc.


andrewlimaza:
Not sure if the `before_submit_button` hook should be duplicated/moved for when the buttons are removed for certain payment gateways.

PR #638, opened 2018-02-13
fix conflict with pmpro-email-templates when emails are disabled

mathieuhays:
I've been using `pmpro-email-templates` and I disabled the admin email on checkout.

When successfully registering, the page dies because the hooks handling the emails aren't expecting a potential `false` value.

I noticed the same conflict in `pmpro-register-helper`, I also submitted a pull request there. I haven't checked the other addons.

### DIFF
--- a/classes/class.pmproemail.php
+++ b/classes/class.pmproemail.php
@@ -104,6 +104,11 @@
 			
 			//filters
 			$temail = apply_filters("pmpro_email_filter", $this);		//allows filtering entire email at once
+
+			if ( empty( $temail ) ) {
+				return false;
+			}
+
 			$this->email = apply_filters("pmpro_email_recipient", $temail->email, $this);
 			$this->from = apply_filters("pmpro_email_sender", $temail->from, $this);
 			$this->fromname = apply_filters("pmpro_email_sender_name", $temail->fromname, $this);


ideadude:
Thanks for this. I'm going to merge and just add a comment above it.

ideadude:
As an aside, returning false to disable an email makes some sense. I'm not sure we ever made a formal decision RE allowing that, but we started using that in other add ons so we should support it. It can make it complicated so anything else filtering the whole email has to also check that $email is not empty before attempting to work with it.

FWIW another way to disable emails is to blank out the $email->email (to email) value. Although returning false and exiting the sendEmail method soon after would save some computation. I've run into similar situations trying to disable email at the WP level. WP only has filters to disable email late in the process and kind of has to go through the motions of getting the email ready vs allowing you to turn it off early.

Just thought I'd add that context here in case related issues come up.

mathieuhays:
Actually looking at the code again I wonder if it wouldn't be better to return `true` instead?

Before this patch `false` meant something wrong happened but if we disable the email that's an expected behaviour so to be consistent we should return `true`.

I checked the code and what's returned from `sendEmail` is not used but it could save some pain in the future in case that changes.

PR #647, opened 2018-02-19
Pbrx gutenberg

pbrocks:
Removes bullet from level checkboxes from Gutenberg post-editor. Non-gutenberg screen seem unaffected, as seen in screencast.

[Screencast shows checkbox with/without Gutenberg](https://api.monosnap.com/rpc/file/download?id=RhNMtK6KT2jXT7CTByNBnIb54QR03h)


pbrocks:
I think this is better.

PR #652, opened 2018-03-06
BUG FIX: Didn't let Addon Pkgs, etc set level info

eighty20results:


ideadude:
We really only want to use the pmpro_checkout_level filter for when getting a level at checkout. There is probably another way to fix this in add on packages or here without using that filter.

Can you explain the bug that's happening?

eighty20results:
The Add-on packages aren't getting listed on the account page

ideadude:
Do you think the add on packages should show up under levels? Or at least adjust the pricing there? The pages purchased should show up under member links per the add on code. Invoices should show up as well.

dparker1005:
Closing as there is now a MemberOrder method `getMembershipLevelAtCheckout()` which will get the level object at checkout via `pmpro_getLevelAtCheckout()`, which uses the `pmpro_checkout_level` filter.

PR #653, opened 2018-03-06
Dev pbrx css suggestion

pbrocks:
Push watermark to bottom of admin page


ideadude:
Chat with @kimcoleman about this. It didn't seem to work for me when I checked out your branch or I'm not sure what the change it.

Personally, I don't mind the powered by watermark showing up at the bottom of the content vs the very bottom of the page.

In general, stylistic changes like this should be discussed ahead of time and PR's done against a version branch. Enhancements and Features against a branch like v1.9.5, bug fixes against the dev branch.

pbrocks:
In this case I made a change on my local and thought you might like to see it. I forgot that the DEV branch was actually production. Perhaps it shouldn't have been a PR, or maybe against another branch would have been better. I'll chat with @kimcoleman, but closing for now.

PR #660, opened 2018-03-20
Stripe response handler, change e-mail template call

Square-Lines:
To avoid namespace collision, added a prefix to the Stripe Response
Handler. Addresses issue when other plug-ins are also invoking Stripe with the default handler.

### DIFF
--- a/classes/class.pmproemail.php
+++ b/classes/class.pmproemail.php
@@ -246,9 +246,6 @@
 				else
 					$this->template = "checkout_paid";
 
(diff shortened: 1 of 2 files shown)


ideadude:
Going to close because the PR is old and can't be merged. There is some kind of email template filter change in there too.

But now that we have Stripe stuff in it's own JS file, we should add prefixes there to avoid conflicts. Yeah. I will do it through a different commit.

PR #662, opened 2018-03-22
BUG FIX: PHP Notice from cron.php

eighty20results:


dparker1005:
This PR is being closed as the core fix in this PR's title is avoiding the `$euser` warnings when running crons. A fix for these warnings has been released in a previous version of PMPro.

As for the other changes in this PR, if any are still issues in the most recent version of PMPro, can you please open a separate PR for each fix? This will make it much easier to review each change individually and release the fix promptly.

PR #669, opened 2018-04-12
Update init.php

mjulian87:
Makes the Membership Level column in WordPress dashboard sortable.


ideadude:
Looks good. I'll test and merge into the next update.

PR #680, opened 2018-04-26
BUG FIX: Ignored TrialBillingCycles value

eighty20results:
When scheduling an update to a custom trial payment amount, the value supplied for the # of payment periods and used the Recurring Billing period value.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1551,7 +1551,7 @@
 					'when' => 'payment',
 					'billing_amount' => $order->PaymentAmount,
 					'cycle_period' => $order->BillingPeriod,
-					'cycle_number' => $order->BillingFrequency
+					'cycle_number' => $order->TrialBillingCycles
 				);
 
 				//now amount to equal the trial #s


ideadude:
There is something else going on here. The cycle_number value that is stored in the update should be the cycle_number for the subscription AFTER the update is processed... meaning it should be correct to set it to $order->BillingFrequency.

The $order->TrialBillingCycles is how many cycles the trial amount should be charged for. So this means we should set the update to fire AFTER that many trial cycles. We didn't code this. Note the comment "This will force TrialBillingCycles > 1 to act as if they were 1" which is stating the limitation/bug here.

We can do better. We could figure out the date when X trial billing cycles passed and set the update to run then. Or have a new update trigger that keeps track of cycles/payments.

eighty20results:
I'm closing this PR as it's not doing what I wanted it to. Am thinking about how to resolve the issue with the Trial Billing Cycles not being used to set _when_ the scheduled update is supposed to happen.

eighty20results:
Closed: Won't change the fact that the number of periods a Trial Period payment amount is supposed to last is forced to 1.

PR #690, opened 2018-05-17
PayPal Failed Payment

mjulian87:
It appears that PayPal send a 'recurring_payment_skipped' IPN rather than 'recurring_payment_failed'.


ideadude:
Merged in, but I kept the check for status failed on the recurring payment transactions. I'm not sure if that is possible, but if so, we would handle it that way.

I also am spelling out $last_subscription_order for those vars now.

And thinking we should look into the "express_checkout" IPN transaction type. Can we check for these with payment status "Failed" and cancel PPE memberships that appear fine at checkout. I think PayPal may send these before or in place of "recurring_payment_profile_cancel"

PR #692, opened 2018-05-23
Gutenberg

dparker1005:
Implement checkout-button block


blocks/membership/index.js
```diff
@@ -32,6 +33,35 @@ const {
     InnerBlocks,
 } = wp.editor;
 
+var all_levels = [{ value: 0, label: "Non-Members" }];
+
+function get_ajax_url() {
+ var i = window.location.href.indexOf("/wp-admin/post.php?");
```
eighty20results: Assuming that wp-admin is at /wp-admin/ puts you at odds with, for instance the Protect WP Admin plugin which allows setting a custom slug - See: https://plugins.trac.wordpress.org/browser/protect-wp-admin/trunk/protect-wp-admin.php#L121. 

How about using `wp_localize_script()` and create a `wp_admin_post` variable using `get_admin_url( 'post.php' )` as the `indexOf()` search?

Then you should be able to use `window.location.href.indexOf( <someprefix>.wp_admin_post + "?" );` and avoid issues with fixed URIs.

PR #694, opened 2018-05-27
Updated German (de_DE) translations

mattes3:
Needed to update the translations because the PHP source files contained so many changed strings. Enjoy!


ideadude:
Do you also help out over here on GlotPress at wordpress.org: https://translate.wordpress.org/locale/de/default/wp-plugins/paid-memberships-pro

PR #720, opened 2018-07-11
Ensure that a TML action exists before attempting to redirect

### LINKED ISSUE #719: PMPro with TML 7 interferes with post passwords
TML used to handle all actions. However, it no longer does. So, `wp-login.php` is still needed to process the `postpass` action. PMPro's login redirect stops this from happening. Expect a PR soon!

jfarthing84:
This fix also has the added benefit of redirecting the requsted
action to the proper TML action.

Fixes #719

### DIFF
--- a/includes/login.php
+++ b/includes/login.php
@@ -50,7 +50,9 @@ add_filter('wp_signup_location', 'pmpro_wp_signup_location');
 
 //redirect from default login pages to PMPro
 function pmpro_login_head()
-{		
+{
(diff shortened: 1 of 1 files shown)


jfarthing84:
@ideadude This is a pretty critical issue - we are getting support requests for this every day. Can we expedite this one?

andrewlimaza:
@jfarthing84 I am bumping this thread internally with Jason.

PR #730, opened 2018-07-26
convert 'the_content' to use wpautop rather

andrewlimaza:
Adjust the confirmation and level description to filter on wpautop instead of the_content. Some page builders were breaking because of multiple 'the_content' calls.


eighty20results:
I believe committing this breaks support for shortcodes in level descriptions, etc...

PR #733, opened 2018-07-27
Content Filter Priority Settings

andrewlimaza:
Enhancement: Allow users to easily set the filter priority number from "Memberships" > "Advanced Settings".


ideadude:
I rethinking this change since we might bundle special code for builder compatibility.

PR #739, opened 2018-08-11
Fix markup of page title

JoeCohen:
Page title doesn't render correctly.
Add space between hash sign and text of page title


ideadude:
Thank you!

PR #745, opened 2018-08-30
True in strict conditions on sanitize_user

dcervantes:
If there are character like '+' in the username wordpress will remove that character in the insert_user.


ideadude:
Seems like WP core and other plugins with registration use this strict option. Nice catch. Thanks for the PR.

PR #746, opened 2018-09-04
Fix 2checkout pmpro_after_checkout_update_consent

cpbotha:
Specify priority and number of arguments when pmpro_after_checkout_update_consent is installed as action. When this is not done, 2checkout flow errors with:

```
Fatal error: Uncaught ArgumentCountError: Too few arguments to function pmpro_after_checkout_update_consent(),
1 passed in /home/cpbotha/webapps/wp_vxuni/wp-includes/class-wp-hook.php on line 288 and
exactly 2 expected in /home/cpbotha/webapps/wp_vxuni/wp-content/plugins/paid-memberships-pro/includes/privacy.php:483
``


ideadude:
Thank you!

PR #758, opened 2018-10-07
Added Kenyan Shilling to currency list

urandu:
Added Kenyan Shilling to currency list. This will enable easier integration to mobile payment solutions in Kenya such as M-pesa

### DIFF
--- a/includes/currencies.php
+++ b/includes/currencies.php
@@ -54,6 +54,7 @@
 				'position' => 'right',
 				'decimals' => 0,
 				),
+			'KES' => __('Kenyan Shilling', 'paid-memberships-pro' ),
 			'MYR' => __('Malaysian Ringgits', 'paid-memberships-pro' ),
 			'MXN' => __('Mexican Peso (&#36;)', 'paid-memberships-pro' ),
 			'NGN' => __('Nigerian Naira (&#8358;)', 'paid-memberships-pro' ),


pbrocks:
Hi there,

Did you know that we make PMPro extensible so you won't need to wait for this to be reviewed. You can use the `pmpro_currencies` filter right now, and what's even better is that you can be more specific with how the currency displays.  Looks like KES normally appears to the right of the value, so we adjust a parameter in the array.

```php
/**
 * pmpro_add_kes_to_currency_array  This function adds KES to currency array as well as the values of the keys to the `$pmpro_currencies` array.
 *
 * @param  array $pmpro_currencies The currency array created by PMPro
 * @return array                   The adjusted currency array
 */
function pmpro_add_kes_to_currency_array( $pmpro_currencies ) {
	$pmpro_currencies['KES'] = array(
		'name'  => __( 'Kenyan Shilling', 'paid-memberships-pro' ),
		'decimals' => '0',
		'thousands_separator' => '&nbsp;',
		'decimal_separator' => ',',
		'symbol' => '&nbsp;KSh',
		'position' => 'right',
	);
	return $pmpro_currencies;
}
add_filter( 'pmpro_currencies', 'pmpro_add_kes_to_currency_array' );
```

Apologies that I know nothing more than what I found online, so the above may be incorrect. But you can use this in your code with the appropriate corrections.

urandu:
@pbrocks Thank you for the comment, i'm currently using the filter you have mentioned. It would be great however for the currency to be added in the default currency list (just in case someone is interested in using pmpro within the country and they`d like their currency to be KES without using my plugin)

PR #761, opened 2018-10-09
Enhancement: pmpro_is_page utility function

mathieuhays:
I found myself needing this function recently so I assume it could be useful for other people too :-)

Not sure if the naming is ideal. I was tempted to call it `is_pmpro_page` which would make more sense but it would break the convention followed for function names (starting with `pmpro_`).


mathieuhays:
> The is_page() function uses the array values by default (no need for the extra array_values() function call overhead, nor the pmpro_is_page() call itself).

Didn't know that, thanks for mentioning it.



> I'm not sure how that utility code helps improve code readability/improves functionality?

I usually avoid using plugin variables directly because their structure may change in future updates. I prefer using documented APIs hence this function for detecting whether a page is handled by Paid Membership Pro or not. For example I prefer using `is_page()` instead of `global $wp_query; $wp_query->is_page()`, both are valid but the function call is more practical.

ideadude:
Other plugins like bbpress, buddypress, and woocommerce have similar functions. We will take a look at this.

sc0ttkclark (approved):

sc0ttkclark (approved):

includes/functions.php
```diff
@@ -3403,6 +3403,17 @@ function pmpro_cleanup_memberships_users_table() {
 	$wpdb->query( $sqlQuery );
 }
 
+/**
+ * Whether the current page is a PMPRO page or not.
+ *
```
sc0ttkclark, line 3408:
```suggestion
 *
 * @since TBD
 *
```

sc0ttkclark (approved):

ideadude:
Couple thoughts.

A common issue when checking something like `if ( is_page( $pmpro_pages['levels'] ) )` is that if the levels page is undefined it will return true for any page because is_page( 0 ) assumes you're checking for any page.

I am not sure if the is_page() function does something similar for `is_page( array( 'levels'=>null, ... ) )`. Let's check. If it does, we need to strip out the empty pages from the array before checking.

I also wonder if we should check pmpro_is_checkout or otherwise check for PMPro blocks/shortcodes on the page and consider pmpro_is_page() true e.g. if there is a checkout block on a page that is not the specified checkout page.

It would also be nice to address this issue at the same time. (https://github.com/strangerstudios/paid-memberships-pro/issues/1958) Basically allow us to pass in a page key like 'levels' and call pmpro_is_page( 'levels' ) instead of `! empty( $pmpro_pages['levels'] && is_page( $pmpro_pages['levels'] ) )`

dparker1005:
Hi there @mathieuhays, thank you again for submitting these changes.

As @ideadude mentioned, the ideal solution is one that takes the level slug as a parameter and streamlines the check for if we are on a specific PMPro page. This is something that we may work on in the future, but I am going to close this PR for now as we would likely want to rebuild this behavior from the ground up.

PR #765, opened 2018-10-14
BUG FIX: Still showing Stripe ver for other GW

eighty20results:
Not hiding the Stripe version when displaying settings for gateways other than Stripe.com (thought there already was a bug fix for this, but it's still being displayed in master & dev branches)


andrewlimaza:
Just found an issue on that commit, seemed to have copied the logical code twice.

PR #766, opened 2018-10-14
BUG FIX: Correct Braintree plans flagged as faulty

eighty20results:


ideadude:
Merged the one pulled against dev.

PR #770, opened 2018-10-18
ENHANCEMENT: Wrap Checkout Button in <div>

eighty20results:
So it's easier to center or move the button to wherever you need it to be with CSS.


kimcoleman:
If we wrap in a div it will force `display: block`. Could we instead wrap in a span - maybe with the classname matching the shortcode:

`<span class="pmpro_checkout_button">`

kimcoleman:
Did this with the span in a separate PR for the button via shortcode or block method.

https://github.com/strangerstudios/paid-memberships-pro/pull/1415

PR #792, opened 2018-11-09
BUG FIX: remove Stripe billing limit warnings on membership levels page

positivejam:
When using Stripe, there used to be text that said you couldn't create a level with billing cycle limits. That warning was removed in commit 649bf536e9ffafd71ec0acef95ed789255a19425.

However, if billing limits are set on the membership level, an error message is displayed on the admin levels page as well as the admin page for the individual level in question.

Assuming the Stripe integration really does support billing cycle limits now, this warning is incorrect and can be removed.

### DIFF
--- a/adminpages/functions.php
+++ b/adminpages/functions.php
@@ -22,43 +22,10 @@
 	Checks if PMPro settings are complete or if there are any errors.
 	
 	Stripe currently does not support:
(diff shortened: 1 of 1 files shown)


ideadude:
Truth. Thanks.

positivejam:
Sure thing.

PR #799, opened 2018-11-12
Adding additional quick search links when no members are found in members list search

kimcoleman:
Adding links to filter search by other filters or search all users when no members are found in a Members List search.

### DIFF
--- a/adminpages/memberslist.php
+++ b/adminpages/memberslist.php
@@ -264,13 +264,18 @@
 				?>
 				<tr>
 					<td colspan="9">
-						<p><?php _e( 'No members found.', 'paid-memberships-pro' ); ?><?php if ( $l ) { ?> <a href="?page=pmpro-memberslist&s=<?php echo esc_attr($s);?>"><?php _e( 'Search all levels', 'paid-memberships-pro' );?></a><?php } ?></p>
+						<p>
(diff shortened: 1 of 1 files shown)


ideadude:
2 notes. These are fixes the old code needed as well.

1. Use esc_url instead of esc_attr for the $s var. We can actually just move the esc_url to wrap the whole thing, since we can't trust admin_url to be escaped and need to escape the whole thing.

2. Use add_query_arg to build the URLs vs ?var=value/etc. Note, we have not been good about doing this in the past, but since this is new code...

So...

`admin_url( 'admin.php?page=pmpro-memberslist&l=cancelled&s=' . esc_attr( $s ) )`

becomes...

`esc_url( add_query_arg( array( 'page' => 'pmpro-memberslist', 'l' => 'cancelled', 's' => $s ) ) );`

kimcoleman:
Is it OK that when I use your code example it is a URL like:

/wp-admin/admin.php?l=cancelled&page=pmpro-memberslist&s=Jason

....seems to be "alphabetizing" the attributes?

kimcoleman:
Do you have a suggested improvement for the users.php search link:

`<?php echo admin_url( 'users.php?s=' . esc_attr( $s ) ); ?>`

PR #812, opened 2018-11-15
Updated Credit card (Front & Back) as well as Printer Icons to .svg f…

travislima:
…iles.


ideadude:
We should figure out if we want to remove the .gif files now. Or wait for a while in case others are using them.

kimcoleman:
I would like to keep them for awhile just in case. We should also run a search to confirm if any of our own add ons rely on the .gif images then update.

PR #820, opened 2018-11-21
Enhancement: Show Order Status on Account Page.

andrewlimaza:
Enhancement: Show order "status" on account page under "Past Invoices".

Inlcudes 'refunded' orders.

### DIFF
--- a/shortcodes/pmpro_account.php
+++ b/shortcodes/pmpro_account.php
@@ -31,7 +31,7 @@ function pmpro_shortcode_account($atts, $content=null, $code="")
 		$ssorder->getLastMemberOrder();
 		$mylevels = pmpro_getMembershipLevelsForUser();
 		$pmpro_levels = pmpro_getAllLevels(false, true); // just to be sure - include only the ones that allow signups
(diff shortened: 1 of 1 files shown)


ideadude:
This looks good, but we probably want to massage the status here a bit, mostly to avoid showing "cancelled" as a status here. (We are trying to remove the cancelled order status in the 2.0 release also.)

Even "success" is an odd term for end user customers. I think we need to check the status and set a $display_status. This also allows translators to overwrite these.

So

```
if ( in_array( $invoice->status, array( '', 'success', 'cancelled' ) ) ) {
    $display_status = __( 'Paid', 'paid-memberships-pro' );
} elseif ( $invoice->status == 'pending' ) {
    // Some Add Ons set status to pending.
    $display_status = __( 'Pending', 'paid-memberships-pro' );
} elseif ( $invoice->status == 'refunded' ) {
    $display_status = __( 'Refunded', 'paid-memberships-pro' );
}
```

eighty20results:
Can be dangerous to assume that `''` === null or empty(), so it's probably more future proof to use:

`if ( empty( $invoice->status ) || in_array( $invoice->status, array( 'success', 'cancelled' ) ) ) {`

PR #825, opened 2018-11-23
Default checkout level when admin selects "view page"

andrewlimaza:
Enhancement: Append first available level ID to 'View Page' checkout button when viewing from admin dashboard.

Filter: 'pmpro_default_view_checkout_level' takes membership level ID which will allow user's to adjust the level ID that is shown by default.


andrewlimaza:
Redoing this PR, there's a better place to handle this which will improve compatibility with page builders etc.

PR #827, opened 2018-11-26
Enhancement: Default checkout level

andrewlimaza:
Default checkout level if no level is passed in, or if the post_meta for default level doesn't exist.

Filter Added: "pmpro_default_level" which passes the default level ID for checkout - this will help developer's filter this without setting the post meta but set it more dynamically.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1937,9 +1937,9 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
 			$all_levels = pmpro_getAllLevels( false, false );
 
 			// Get lowest level ID.
-			$default_level =  min( $all_levels );
+			$default_level =  min( array_keys( $all_levels ) );
 
-			$level_id = apply_filters( 'pmpro_default_level', intval( $default_level->id ) );
+			$level_id = apply_filters( 'pmpro_default_level', intval( $default_level ) );
 			
 			// Bail back to levels page if level ID is empty or less than 1.
 			if ( empty( $level_id ) || $level_id < 1 || ! is_int( $level_id ) ) {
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -1929,7 +1929,23 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
 
 	// no level, check for a default level in the custom fields for this post
 	if ( empty( $level_id ) && ! empty( $post ) ) {
+
 		$level_id = get_post_meta( $post->ID, 'pmpro_default_level', true );
+
+		// if still empty, let's try get first available level.
+		if ( empty( $level_id ) ) {
+			$all_levels = pmpro_getAllLevels( false, false );
```
ideadude: This returns an array of objects, not just level IDs.

includes/functions.php
```diff
@@ -1929,7 +1929,23 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
...
+
+			// Get lowest level ID.
+			$default_level =  min( $all_levels );
```
ideadude: $default_level = min( array_keys( $all_levels ) );

includes/functions.php
```diff
@@ -1929,7 +1929,23 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
...
+
+			$level_id = apply_filters( 'pmpro_default_level', intval( $default_level->id ) );
```
ideadude: Then need to edit this to just pass $default_level if we use array_keys above.

ideadude:
Hopefully the review/comments make sense on this. I messed up with the GitHub UI a bit.

In any case, I am not sure that using the php function min() on the associative array of level objects is going to have the desired behavior. Maybe PHP in this case sorts the objects by the array key or the first value in the object (which would be the id), but even if that does happen, it's too obscure to trust in this case.

Instead let's use array_keys() to get the keys/level ids from the array of levels and get the min from that.

Then later on when you use $default_level->id, you'll need to just use $default_level.

Maybe change the var name to $default_level_id to make it more clear.

Thanks!

PR #829, opened 2018-11-27
ENHANCEMENT: Updated Braintree/PHP Library to 3.36.0

greathmaster:


ideadude:
Gonna merge now, but we need to test extensively before pushing v2.0 out.

Issue #498, opened 2017-09-08
Discount Code placement under CVV


contemplate:
Hey @strangerstudios so I guess I never noticed this until now but I see the Discount Code box at checkout on every membership site I've built below the CVV.

<img width="564" alt="screen shot 2017-09-08 at 8 22 09 am" src="https://user-images.githubusercontent.com/1093666/30211865-9643f4ac-9470-11e7-97a8-635e935d9702.png">

Not sure if it's always been there but we have more and more customers mistakenly hitting the "Apply" button after filling in their credit card info rather than the submit button.

I can hide this field through CSS but could the placement of this be re-thought?  or have it hidden by default like the one at the top?

strangerstudios:
That's a good point. We've been looking at this during a UI update and thinking about making that a text link instead of a button. That might help.

Better ways to choose if this is displayed etc could be helpful for other reasons.

contemplate:
Excellent.  Yes text link or something else would be great.  The only button after the CC info should be the Submit button.

strangerstudios:
Another thing we should do is show an error ("Please enter a discount code to apply.") when the apply button is clicked and the discount code is empty. This will make it more clear that the apply here refers to the discount code.

kimcoleman:
I'm going to create a new issue for potentially adding the error message ("Please enter a discount code to apply.") when the apply button is clicked and the discount code is empty.

PR #830, opened 2018-11-27
Updating appearance and error message of the Discount Code "Apply" button

kimcoleman:
Updated "button" to appear as a link

Updating error message when "Apply" button is clicked but no code was given.

Resolves issue https://github.com/strangerstudios/paid-memberships-pro/issues/498

### DIFF
--- a/css/frontend.css
+++ b/css/frontend.css
@@ -75,6 +75,29 @@ fieldset[disabled] .pmpro_btn {
 .pmpro_btn.pmpro_btn-submit-checkout {
 	width: auto;
 }
+
+input[type="button"]#discount_code_button,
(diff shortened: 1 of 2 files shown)


ideadude:
> Updated "button" to appear as a link

This is fine.

> Updating error message when "Apply" button is clicked but no code was given.

This is messy because the old code was bad. We had the same JS/jQuery.ajax written out twice when we could have abstracted it to once. Now we have it 3 times, which is no bueno. We mention "invite" button instead of "apply" in a comment. We are working to move this JS code into .js files anyway.

Let's look at this together.

kimcoleman:
Just removed that addition al JS and made a small CSS improvement.

PR #834, opened 2018-11-29
Dev checkout js

pbrocks:
This move js out of checkout page and into its own file to be enqueued

### DIFF
--- a/css/checkout-page.css
+++ b/css/checkout-page.css
@@ -1,0 +1,76 @@
+	.dialog-box {
+		border: 1rem solid lightgray;
+		background-color: #fff;
+		color: #444;
+		width: 100%;
+	    position: absolute;
+	    left: 25%;
+	    z-index: 33;
+	    bottom: 4%;
+	    display: none;
+	}
+
+	.box {
+		color: #fff;
+		border-radius: 5px;
+		padding: .2rem 1rem;
+		font-size: 90%;
+
+	}
+
+	.box .box {
+		color: #444;
+	}
+	.d{
+		display: grid;
+		grid-gap: .3rem;
+		grid-template-columns: 5fr 3fr;
+	}
+	.top.box.e {
+		color: red;
+	}
+	.e {
+		text-align: center;
+		grid-column: 1 / 3;
+	}
+
+	.f {
+		grid-column: 1;
+		grid-row: 2;
+	}
+
+	.g {
+		grid-column: 2;
+		grid-row: 2;
+	}  
+#small-x-close {
+	float: right;
+	background: gray;
+	padding: 1px 11px;
+	border-radius: 50%;
+	color: white;
+	}  
+#small-x-close:hover {
+	cursor: pointer;
+}
+
+#close-button {
+	text-align: center;
+}
+@-webkit-keyframes fadeIn {
+	to {
+		opacity: 1;
+	}
+}
+
+@keyframes fadeIn {
+	to {
+		opacity: 1;
+	}
+}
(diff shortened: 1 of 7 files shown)


includes/card-cvv.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+?>
+	<div id="cvv-window" class="fade-in dialog-box">
+		<div id="small-x-close">x</div>
+		<div class="box d">
+			<div class="top box e">
+				<h3>Locating your Credit Card CVV/CVV2 Security Code</h3>
```
ideadude: Let's localize the strings in this file.

includes/init.php
```diff
@@ -112,6 +112,43 @@ function pmpro_init()
...
+		'checkout-page',
+		'checkout_page_object',
+		array(
+			'checkout_page_ajaxurl'   => admin_url( 'admin-ajax.php' ),
+			'checkout_page_nonce'     => wp_create_nonce( 'checkout-page-nonce' ),
+			'applydiscountcode'		  => apply_filters( 'pmpro_ajax_timeout', 5000, 'applydiscountcode' ),
+		)
+	);
+	wp_register_style( 'checkout-page', plugins_url( '/css/checkout-page.css', __DIR__ ),  time() );
+	if ( is_page( $pmpro_pages['checkout'] ) ) {
```
ideadude: This is unreliable since you can have the checkout shortcode on multiple pages.

Instead, let's enqueue the script from within preheaders/checkout.php, which should work and will make sure that the code is only enqueued on checkout pages.

css/checkout-page.css
```diff
@@ -0,0 +1,76 @@
+	.dialog-box {
```
ideadude: Is this for the modal? If so, let's remove the file and not enqueue it. Any CSS needed can go into frontend.css. Let's have one frontend css file only.

js/other-discount-code-button.js
```diff
@@ -0,0 +1,65 @@
+jQuedry(document).ready(function($){
```
ideadude: Why is this a separate file?

I was looking at the applydiscountcode JS code earlier and realize that we should be able to abstract this to a function so we don't have so much redundant code.

pages/checkout.php
```diff
@@ -1,689 +1,634 @@
-<?php
```
ideadude: changes the whole page because of the tab and/or spacing updates. We need a WPCS pass first so we can better merge this. I will not merge a PR that replaces the whole file. Too likely we miss something with so many updates going on.

includes/card-cvv.php
pbrocks: Deleting this file, since we're removing this from checkout page.

includes/init.php
pbrocks: Fair enough. We can register here and enqueue from `preheaders`

css/checkout-page.css
pbrocks: Removing

js/other-discount-code-button.js
pbrocks: this needs to be deleted. Shouldn't have been here.

pages/checkout.php
pbrocks: This may be fallout from merging with v2. I didn't replace the whole page when coding.

pbrocks:
Closing this one opting for [cleaner checkout.php commits on #836](https://github.com/strangerstudios/paid-memberships-pro/pull/836)

Issue #292, opened 2016-02-15
Pass levels to email


curtismchale:
So if one wants to customize an email for the purchase of a specific level they can't. At least I haven't seen this yet.

Really emails should get the level that is being purchased so that we could conditionally change out the email content (or subject) based on the level purchased.

strangerstudios:
While there isn't a level value associated with the email object, we do store 'membership_id' in the data array for each of the default PMPro email templates. You can also use $data['user_id'] to look things up via the user id. Or check for the membership level associated with the invoice object for paid levels.

Here is an example:
https://gist.github.com/strangerstudios/09c576fef8eb888bbb1a

I think you're a paying member, so if you need a more specific solution for this, post there with info about your levels and content, and we can give you a better gist.

This post has a gist that adds the level confirmation message (set on the edit level page in the dashboard) to emails: http://www.paidmembershipspro.com/add-custom-level-confirmation-message-checkout-e-mails/

strangerstudios:
Something quick that could help support this is a feature in core to include the level confirmation message in the email. We even added a checkbox at some point to the edit level page under the confirmation message to "Also include in the confirmation email" but never wired it up. For reference, here is a gist to include the confirmation message in the confirmation email (vs just the confirmation page).

https://gist.github.com/scottsousa/6700307

kimcoleman:
This PR adds an option to include the level confirmation message in email: https://github.com/strangerstudios/paid-memberships-pro/issues/292

I think an enhancement to the Email Templates Admin Editor would be a better place to offer level-specific emails. Could work similar to Nav Menus Add On where you check a box on the level to optionally create custom templates for each and then the dropdown to select a template on "Email Templates" admin page would include custom emails for each level.

kimcoleman:
I've added an issue on the Email Templates Admin Editor repository to add level-specific emails. See: https://github.com/strangerstudios/pmpro-email-templates/issues/10

PR #839, opened 2018-12-01
Adding level meta option to include confirmation message in email

kimcoleman:
Updates the email class and email templates to respect the level setting and include the confirmation message if set.


kimcoleman:
Resolved part of issue https://github.com/strangerstudios/paid-memberships-pro/issues/292

PR #848, opened 2018-12-07
Dev pointers v2

pbrocks:
I think it'd be good to put a link in the pointer, but adding the query argument didn't seem to work

### DIFF
--- a/includes/adminpages.php
+++ b/includes/adminpages.php
@@ -87,6 +87,7 @@ function pmpro_parent_file( $parent_file ) {
 		$parent_file = 'pmpro-dashboard';
 		$plugin_page = 'pmpro-dashboard';
 		$submenu_file = 'pmpro-membershiplevels';
+		$submenu_file .= '<span id="membershiplevels"></span>';
 	}
 	
 	return $parent_file;
--- a/includes/pointers.php
+++ b/includes/pointers.php
@@ -93,16 +93,16 @@ class PMPro_WP_Pointers {
 		// The query variable is used to determine which pointer to display.
 		// *****************************************************************************************************
 		$tour = array(
-			'settings'      => array(
-				'id'       => '#dashboard_quick_press',
+			'newlevels'      => array(
+				'id'       => '#pmpro-new-levels',
 				'content'  => '<h3>' . __( 'Congratulations!', 'paid-memberships-pro' ) . '</h3>'
 					. '<p><strong>' . __( 'WP Pointers is working properly.', 'paid-memberships-pro' ) . '</strong></p>'
 					. '<p>' . __( 'This pointer is attached to the "Quick Draft" admin widget.', 'paid-memberships-pro' ) . '</p>'
 					. '<p>' . __( 'Our next pointer will take us to the "Settings" admin menu.', 'paid-memberships-pro' ) . '</p>',
 				'button2'  => __( 'Next', 'paid-memberships-pro' ),
-				'function' => 'window.location.href="' . get_admin_url( 'admin.php?page=pmpro-membershiplevels' ) . '"',
+			$function = 'document.location="' . $this->get_admin_url( 'admin.php', 'pmpro-memberslist' ) . '";',
 			),
-			'site_title'       => array(
+			'newsettings'       => array(
 				'id'       => '#pmpro-new-settings',
 				'content'  => '<h3>' . __( 'Moving along to Site Title.', 'paid-memberships-pro' ) . '</h3>'
 				. '<p><strong>' . __( 'Another WP Pointer.', 'paid-memberships-pro' ) . '</strong></p>'
--- a/includes/pointers.php
+++ b/includes/pointers.php
@@ -1,0 +1,261 @@
+<?php
+/**
+ * Notes in a Pointer dialog box for guiding users in the dashboard interface.
+ */
+
+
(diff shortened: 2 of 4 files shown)


includes/pointers.php
```diff
@@ -0,0 +1,112 @@
...
+	if ( ! in_array( 'pmpro_v2_tour', $dismissed_pointers ) ) {
+		build_pointer_script( $id, $options, __( 'Close', 'paid-memberships-pro' ), $button2, $function );
+	}
+}
+
+/**
+ * [build_pointer_script]
+ * 
+ * @param  [type]  $id       
+Thank you for creating with WordPress. Get Version 5.0
```
ideadude: What?

paid-memberships-pro.php
```diff
@@ -32,9 +32,9 @@
 require_once( PMPRO_DIR . '/includes/localization.php' );           // localization functions
 require_once( PMPRO_DIR . '/includes/lib/name-parser.php' );        // parses "Jason Coleman" into firstname=>Jason, lastname=>Coleman
 require_once( PMPRO_DIR . '/includes/functions.php' );              // misc functions used by the plugin
+require_once( PMPRO_DIR . '/includes/pointers.php' );              // misc functions used by the plugin
 require_once( PMPRO_DIR . '/includes/updates.php' );                // database and other updates
 require_once( PMPRO_DIR . '/includes/upgradecheck.php' );           // database and other updates
-require_once( PMPRO_DIR . '/includes/deprecated.php' );              // deprecated hooks and functions
```
ideadude: If you pull recent upstream work, your PR shouldn't try to remove these lines.

includes/adminpages.php
```diff
@@ -87,6 +87,7 @@ function pmpro_parent_file( $parent_file ) {
 		$parent_file = 'pmpro-dashboard';
 		$plugin_page = 'pmpro-dashboard';
 		$submenu_file = 'pmpro-membershiplevels';
+		$submenu_file .= '<span id="membershiplevels"></span>';
```
ideadude: Is this the only way to get an id into the menu? Can you select on something else. I'm surprised this doesn't break the menu when on a sub tab. Did you test that? Maybe we don't need to point exactly to this item if we can point to something close.

includes/pointers.php
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Notes in a Pointer dialog box for guiding users in the dashboard interface.
+ */
+add_action( 'admin_enqueue_scripts', 'pmpro_enqueue_admin_scripts' );
+/**
+ * [pmpro_enqueue_admin_scripts] Enqueue the scripts needed to builder admin pointers.
+ * 
+ * @return void
+ */
+function pmpro_enqueue_admin_scripts() {
```
ideadude: pmpro_enqueue_pointer_scripts

includes/pointers.php
```diff
@@ -0,0 +1,112 @@
...
+	wp_enqueue_script( 'wp-pointer' );
+	// hook the pointer
+	add_action( 'admin_print_footer_scripts', 'prepare_pointer_script' );
+}
+/**
+ * [prepare_pointer_script] Details about PMPro 2.0 that are added to the Admin Pointer
+ * 
+ * @return void
+ */
+function prepare_pointer_script() {
```
ideadude: pmpro_prepare_pointer_scripts

includes/pointers.php
```diff
@@ -0,0 +1,112 @@
...
+Thank you for creating with WordPress. Get Version 5.0
+Notice: Undefined variable: button2 in /app/public/wp-content/plugins/paid-memberships-pro-dev-pointers-v2/includes/pointers.php on line 42
+
+ * @param  [type]  $options  [description]
+ * @param  [type]  $button1  [description]
+ * @param  boolean $button2  [description]
+ * @param  string  $function [description]
+ * @return [type]            [description]
+ */
+function build_pointer_script( $id, $options, $button1, $button2 = false, $function = '' ) {
```
ideadude: pmpro_build_pointer_script

includes/pointers.php
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * [prepare_pointer_script] Details about PMPro 2.0 that are added to the Admin Pointer
+ * 
+ * @return void
+ */
+function prepare_pointer_script() {
+	$show_pointer = true;
+	$file_error   = true;
+
+	$id       = '#toplevel_page_pmpro-dashboard';
```
ideadude: we would put this stuff into an array if/when we have multiple pointers, but okay as is for now

includes/adminpages.php
pbrocks: Actually, we don't need this now, since we're not doing the step thru. I'll close this PR and put together a cleaner one.

includes/pointers.php
pbrocks: Agreed. I thought I had that already.

includes/pointers.php
pbrocks: DOH!

includes/pointers.php
pbrocks: I agree that we want to prefix this, but want to add that since this is going into core, a clone of this function can be used for any Add On that wants a Pointer.

pbrocks:
close for this instead: pointers v2 clean PR #853

PR #861, opened 2018-12-14
Hide 'couponamount' in order edit and order csv export

kimcoleman:
Added filter 'pmpro_orders_show_coupon_amounts' to allow site to reenable these fields.

### DIFF
--- a/adminpages/orders-csv.php
+++ b/adminpages/orders-csv.php
@@ -330,6 +330,14 @@ $default_columns = array(
 	array( "discount_code", "code" )
 );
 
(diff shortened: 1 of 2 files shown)


ideadude:
Could just use the pmpro_order_list_csv_default_columns filter later? Better to have a specific filter for this?

kimcoleman:
Filtering via the `pmpro_order_list_csv_default_columns` would only adjust it for the CSV export, but not on the "Orders" single view in the admin. A separate filter is needed IMHO.

PR #883, opened 2019-01-02
Bug fix: Paypal Express orders stuck in "review"

TravisHardman:
https://www.paidmembershipspro.com/forums/topic/paypal-express-orders-seemingly-complete-but-in-review-status/


ideadude:
Thanks for submitting this so we can look into it further. As you mentioned in the thread, this feels like a band aid and we need to figure out why the order status would be success here and why the subscribe method wouldn't run later on after the PayPal transaction is confirmed to set the status to success.

TravisHardman:
Adding a note here because this fix seems not to have worked after all. We are still experiencing the bug. Surprised that it's not happening to anyone else. Message Jason about it here: https://www.paidmembershipspro.com/forums/topic/paypal-express-orders-seemingly-complete-but-in-review-status/#post-182949

ideadude:
Are you still having this issue? I updated the PMPro forum thread too. I feel that updates since Jan may have affected this.

PR #903, opened 2019-01-14
Order ID Missing in some cases

andrewlimaza:
Bug Fix: Order id missing in some cases.


kimcoleman:
Jason fixed this via a similar method.

PR #913, opened 2019-01-23
Bug Fix Cancel email data.

dcervantes:
Use "display_name" and "user_email" in the array. The cancel email template is using this fields, but the sendCancelEmail method isn't.
To avoid that the email contains !!display_name!! (!!user_email!!) I include those in the data array.


ideadude:
I think it's good to send dupe name and display_name for backwards compatibility. The other emails use name still. I'm going to pull this, then do a bit of cleanup. Thanks.

PR #917, opened 2019-01-28
Fix database upgrade error on localhost environment

codezz:
See error here: http://7thqueen.com/public-files/gabi/Screenshot_2019-01-28_at_11.28.53.png
The recommended pull request should fix it.


codezz:
@ideadude I hope you can review this and included it in the wordpress repo version since PMPRO  cannot be used on localhost

ideadude:
Thank you. This makes sense. We would need a similar update to our code here: https://github.com/strangerstudios/paid-memberships-pro/blob/991993479b6df3217b4b80d2fb63f1fee970e300/adminpages/emailsettings.php#L54

I will pull this in, make the change to emailsettings.php, and include this in the next update that goes out.

PR #919, opened 2019-01-29
Allow dequeueing of jquery.creditCardValidator on billing page

rafecolton:
Move the `wp_euqueue_script` call to before `do_action( 'pmpro_billing_preheader')` so we can use that action to dequeue the script. Use case is replacing the form on the page with a custom form that has its own validation.

**NOTE:** In the checkout preheader, the action already comes after enqueueing the script.

### DIFF
--- a/preheaders/billing.php
+++ b/preheaders/billing.php
@@ -34,11 +34,11 @@ if (empty($user_order->gateway)) {
 // Set the gateway, ideally using the gateway used to pay for the last order (if it exists)
 $gateway = !empty( $user_order->gateway ) ? $user_order->gateway : pmpro_getOption("gateway");
 
-//action to run extra code for gateways/etc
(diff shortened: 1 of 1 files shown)


ideadude:
I can't think of anything that would break by moving the do_action.

However, since the preheader's are run on the wp hook priority 1

https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/init.php#L167

... you could unenqueue it by hooking into wp priority > 1.

`add_action( 'wp', 'my_unenqueue_pmpro_credit_card_validator' );`

And then add that callback to unenqueue the script. Let me know if this works.

rafecolton:
Why not just swap the order to make it consistent with the checkout preheader?

ideadude:
>consistent with the checkout preheader

Makes sense. Merged.

PR #941, opened 2019-03-13
ENHANCEMENT: Cache user levels to reduce MySQL queries

seagyn:
We have a site that is firing off this specific function multiple times (I haven't done a trace but I assume it's because there is a large number of "posts" on the page and it's checking members level for each one).

This is just a quick solution to reduce those queries and is limited to each request if there is no persistent object cache and will be persisted if there is (making future references quicker as well).

I set expiration to one day but it might need to be shorter depending on your opinion of how often levels change.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1082,6 +1082,10 @@ function pmpro_changeMembershipLevel( $level, $user_id = null, $old_level_status
 	// remove cached level
 	global $all_membership_levels;
 	unset( $all_membership_levels[ $user_id ] );
+	
+	// remove levels cache for user
+	$cache_key = 'user_' . $user_id . '_levels';
+	wp_cache_delete( $cache_key, 'pmpro' );
 
 	// update user data and call action
 	pmpro_set_current_user();
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1792,31 +1792,37 @@ function pmpro_getMembershipLevelsForUser( $user_id = null, $include_inactive = 
 	$user_id = intval( $user_id );
 
 	global $wpdb;
-
-	$levels = $wpdb->get_results(
-		"SELECT
-								l.id AS ID,
-								l.id as id,
-								mu.id as subscription_id,
-								l.name,
-								l.description,
-								l.expiration_number,
-								l.expiration_period,
-								mu.initial_payment,
-								mu.billing_amount,
-								mu.cycle_number,
-								mu.cycle_period,
-								mu.billing_limit,
-								mu.trial_amount,
-								mu.trial_limit,
-								mu.code_id as code_id,
-								UNIX_TIMESTAMP(startdate) as startdate,
-								UNIX_TIMESTAMP(enddate) as enddate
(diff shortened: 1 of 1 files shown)


seagyn:
@andrewlimaza has suggested clearing the cache when the user level is updated. This will clear the requirement to set an expiry.

ideadude:
Beautiful. We could do more of this in other places.

We could even just use a static var to cache per page load... using the cache function works and if we update when levels change, we are fairly safe.

I'm going to merge and then just tweak to put pmpro_ prefixes on the cache keys.

ideadude:
I'm also going to experiment with the if ( ! $levels ) test. If a user has 0 levels, this might still run the query every time.

eighty20results:
Save yourself the time and simply check for `empty( $levels )`

ideadude:
Hi Thomas! I want to make sure an empty array() or false isn't seen an assumed the cache didn't exist.

Also TIL/Remembered the cache functions have a domain and pmpro is used as the domain so no prefixes needed for the key.

seagyn:
Hi Jason,

I actually found something really cool. There are two more parameters to
wp_cache_get. The 3rd on is if the local cache should be updated with the
value. I set this to true. The 4th value is a variable passed by reference
if a value was found or not for exactly your query you’ve made now.

I’m happy to make another PR to fix that?

ideadude:
FYI I changed it to check if $levels === false which works.
https://github.com/strangerstudios/paid-memberships-pro/commit/3adaa3d3e31a760c24b7274b6fa3083ee2ecbed9

I'm reading up on local vs persistent caches for the $force param.

The $found param would be useful for that === false check, but since we store no levels as array() the checking for false works here too. The code to return $found and check would be a little too verbose IMO.

In my tests (no object cache setup I think) the cache is clearing on every page load. Which I think is fine for this query that should be relatively quick when not fired hundreds of times (Really surprised I didn't notice this sooner... actually reminds me of a few "performance issues since upgrading to 2.0" type reports). I wonder if/what we should set the expires value in the set cache to. You said "I set expiration to one day but it might need to be shorter depending on your opinion of how often levels change.", but it was blank when I looked and the default is "as long as possible".

ideadude:
Related: https://codex.wordpress.org/Class_Reference/WP_Object_Cache#Persistent_Caching

ideadude:
local = memory
persistent = DB or however the specific caching plugin is setup to store things (e.g. memcache)

I figured, but wanted to check the docs. I might choose something like 1 day in the expires just in case someone's level changes without going through the changeMembershipLevel function and the cache delete you added.

seagyn:
That’s a fair case. There might be a commit which has the 1 day expirey in
it.

Local is more likely the current request cache (which will be in memory)
which may have another value from another request and this will update it.
Actually should look at why exactly it is doing.

The persistent object caches out there are actually pretty useful for sites
running at scale so it might be a cool resource for you to have on your
site.

ideadude:
FWIW We use persistent cache on our site. Just not on my dev site I was testing on at the time.

seagyn:
I meant as a blog post 🙈

Issue #942, opened 2019-03-15
Fix misspelled action hooks (deprecate and correct)


kimcoleman:
In the twocheckout-ins.php file, the word "recurring" is misspelled in two action hooks. These should be moved to the deprecated list and replaced in the handler with the correct spelling.

https://github.com/strangerstudios/paid-memberships-pro/blob/991993479b6df3217b4b80d2fb63f1fee970e300/services/twocheckout-ins.php#L463

and 

https://github.com/strangerstudios/paid-memberships-pro/blob/991993479b6df3217b4b80d2fb63f1fee970e300/services/twocheckout-ins.php#L495

ideadude:
We should at least add a do_action with the correct spelling for each of these.

Then we can in the short term leave the old do_action in place with a comment. We can update our deprecation stuff to handle cases like this and add these to it.

related https://github.com/strangerstudios/paid-memberships-pro/issues/940

PR #943, opened 2019-03-18
Updating action to use $morder; Fixing action misspelling

kimcoleman:
Updated the `pmpro_subscription_recuring_stopped` and `pmpro_subscription_recuring_restarted` actions to filter the $morder (not the $last_order which isn't passed to the function). Added a duplicate action to correct the spelling of "recurring". 

We will need to deprecate this incorrect filter name and remove eventually. See issue: https://github.com/strangerstudios/paid-memberships-pro/issues/942


eighty20results:
Partial overlap with #659

PR #946, opened 2019-03-20
ENHANCEMENT: pmpro_gateway_change_subscription_delay

eighty20results:
Add filter to let programmer change the duration of the plan 'trial' (i.e. time until first payment from the plan). Can be used by the Subscription Delay add-on to ensure billing happens on the same date even when the specified delay value is prior to the current date.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1591,7 +1591,7 @@
                     "amount" => $amount * $currency_unit_multiplier,
                     "interval_count" => $order->BillingFrequency,
                     "interval" => strtolower($order->BillingPeriod),
-                    "trial_period_days" => $trial_period_days,
+                    "trial_period_days" => apply_filters( 'pmpro_gateway_change_subscription_delay', $trial_period_days, $order ), 
                     "name" => $order->membership_name . " for order " . $order->code,
                     "currency" => strtolower($pmpro_currency),
                     "id" => $order->code


ideadude:
Hmm. There is already a filter for the plan:
pmpro_stripe_create_plan_array https://github.com/strangerstudios/paid-memberships-pro/blob/354050b65e175ac99c5f2bd50e8e6c29178240e8/classes/gateways/class.pmprogateway_stripe.php#L1600

And for the subscription start date:
pmpro_profile_start_date https://github.com/strangerstudios/paid-memberships-pro/blob/354050b65e175ac99c5f2bd50e8e6c29178240e8/classes/gateways/class.pmprogateway_stripe.php#L1548

Do you think we need the new filter for clarity? Since we use the Stripe trials to do things like separate out the initial payment, I think it might be more confusing to have a filter for "trail days". A monthly plan with no "trial" should still have 28-31 "trial days" at checkout.

eighty20results:
I saw that, but neither of those are global (applicable in other gateways). The filters are also sorely lacking in terms of useful info to let someone calculate what the actual number of trial days for the plan need to be (if the start date is before the current date, which causes trouble on Stripe). So figured it made sense to have an explicit filter for the "test period" calculation that includes all the member specific order/level info in one go.

ideadude:
We can pass the order to the plan filter. That makes sense. And document this stuff better.

dparker1005:
Closing as the `pmpro_profile_start_date` filter is now included in the `pmpro_calculate_profile_start_date()` function, which is now considered a "standard" across gateways for calculating the profile start date.

PR #952, opened 2019-04-01
Preventing double cancellations of Stripe subscriptions.

messica:
cancelSubscriptionAtGateway() was being called from Stripe's pmpro_checkout_before_processing() code first and then later during pmpro_changeMembershipLevel().

There may be a better way to fix this like enabling and disabling the pmpro_cancel_previous_subscriptions filter before and after membership level change, but this fix patched it on a support user's site.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1827,6 +1827,7 @@
 				$subscription = $this->getSubscription($order);
 
 				if(!empty($subscription) 
+				        && empty( $subscription->canceled_at )
 					&& ( empty( $pmpro_stripe_event ) || empty( $pmpro_stripe_event->type ) || $pmpro_stripe_event->type != 'customer.subscription.deleted' ) )
 				{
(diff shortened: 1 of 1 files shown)


ideadude:
I think this is good.

Checking for $subscription->canceled_at will definitely avoid trying to cancel twice. My first thought however was we should figure out WHY it's trying to cancel twice. Fixing it here might hide an issue elsewhere.

I'm double checking if these new double cancellations have to do with our code that cancels Stripe subs BEFORE checkout... if it is trying to cancel them again after checkout. The related orders should be updated to cancelled status so we would skip them, but maybe that is not happening.

In any case, it's a good idea to check if a sub is cancelled before trying to cancel it. The right place for this would be right before the Stripe API calls in the cancelSubscriptionAtGateway method.

I'm going to close this PR and do something like that. Thanks.

ideadude:
Here's what I did. https://github.com/strangerstudios/paid-memberships-pro/commit/bd65f2fa7fb7c131061d561c16c3f73f58368fec

PR #955, opened 2019-04-04
removed 'membership_code' from attribute.

andrewlimaza:
Bug Fix: Removed the "membership_code" attribute from the `[pmpro_member]` as it was non existent in our getmembershiplevelforuser call.

ENHANCEMENT: added membership_description and membership_confirmation attributes in the shortcode. This may be useful for users.

### DIFF
--- a/shortcodes/pmpro_member.php
+++ b/shortcodes/pmpro_member.php
@@ -26,7 +26,6 @@ function pmpro_member_shortcode($atts, $content=null, $code='')
 		'membership_name',
 		'membership_description',
 		'membership_confirmation',
-		'membership_code_id',
 		'membership_initial_payment',
 		'membership_billing_amount',
 		'membership_cycle_number',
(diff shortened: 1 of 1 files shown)


andrewlimaza:
@kimcoleman If this gets merged in, we would need to update the documentation for this. `membership_code` was not returning anything and is not available in the pmpro_getMembershipLevelForUser function.

ideadude:
It seems like code_id is not a property of the membership_level object either.

We could either account for these in the shortcode as specific requests.

Or we could update the getMembershipLevelForUser() and pmpro_getMembershipLevelsForUser() functions to also get the discount code id and code.

I prefer updating the shortcode, since the discount code is technically applied to the order not the level. So it doesn't make sense to put it on the membership_level object. With the shortcode, the param would basically mean "give me the last discount code this user used at checkout". We could look for their last order and get the code from it. Hmm.

In reality, some people use discount codes basically as a way to segment members within a membership level. So they would want to check for the discount code from the last order with a discount code for that user. That's a more complicated question.

Since they weren't working before and we can't figure out how to make them work in an uncomplicated way. Let's remove both the code and code_id here.

andrewlimaza:
> With the shortcode, the param would basically mean "give me the last discount code this user used at checkout". We could look for their last order and get the code from it. Hmm.

I was thinking about this, but figured out user's may get confused with this.

> Let's remove both the code and code_id here.

Update in-coming.

PR #956, opened 2019-04-08
Add status to invoice page.

andrewlimaza:
Add status to invoice page.


ideadude:
It seems like code_id is not a property of the membership_level object either.

We could either account for these in the shortcode as specific requests.

Or we could update the getMembershipLevelForUser() and pmpro_getMembershipLevelsForUser() functions to also get the discount code id and code.

I prefer updating the shortcode, since the discount code is technically applied to the order not the level. So it doesn't make sense to put it on the membership_level object. With the shortcode, the param would basically mean "give me the last discount code this user used at checkout". We could look for their last order and get the code from it. Hmm.

In reality, some people use discount codes basically as a way to segment members within a membership level. So they would want to check for the discount code from the last order with a discount code for that user. That's a more complicated question.

Since they weren't working before and we can't figure out how to make them work in an uncomplicated way. Let's remove both the code and code_id here.

PR #957, opened 2019-04-08
Fix constant errors.

andrewlimaza:
BUG FIX: general constants that may be declared in other plugins or themes check.


ideadude:
Should probably check if ! defined for the SITENAME constant as well and add a comment saying that we have these unprefixed for reverse compatibility

PR #965, opened 2019-04-26
ENHANCEMENT: Added Level ID's to membership level names in admin

greathmaster:


ideadude:
Do we want to include the level ID here? Is there a neater way of doing it?

kimcoleman:
We could do this as a hidden column and allow a filter to choose default columns instead.

ideadude:
We have an optional membership level id column now on the members list table.

PR #966, opened 2019-04-27
BUG - Fixes issue-918: Lower discount code field not showing for offs…

greathmaster:
…ite and check gateways.


ideadude:
The idea was to have the discount code field near both the total price and the payment fields when they are shown since people may expect that.

If there are no payment fields, I don't feel we need to show a duplicate discount code field. In fact, it will show up very close to the one at the top unless the site is using the add on to show the billing address for free/offsite gateways.

This should be a gist IMO.

sc0ttkclark:
Closing based on previous comment here from Jason.

PR #974, opened 2019-05-08
FEATURE: Checking for valid Stripe Webhook when saving recurring memberships

greathmaster:
PR updates Stripe PHP library to the latest version, and adds Stripe webhook check when saving recurring memberships.


messica:
This is great but I don't think checking the webhook every time a recurring subscription membership level is saved is necessary. This could also work as a button on the Payment Settings page.

ideadude:
Instead of a button, we could run the check when the payment gateway settings page loads and show an error if there is an issue.

Upping the Stripe Library API version is a big deal, but we were going to do that anyway soon.

We can squeeze this into v2.1

Issue #969, opened 2019-05-03
Bug: Stripe subscription updates cancelling subscriptions.


messica:
Subscription updates set to "now" (and maybe others, needs testing) are cancelling subscriptions at Stripe without creating new ones. The new ones are failing to create because the `trial_period_days` is incorrect.

**Support Ticket (Moderators Only)**
https://www.paidmembershipspro.com/forums/topic/urgent-now-updates-causing-issues-in-stripe-cancelling-memberships/

messica:
"T0:0:0" is being added at https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/gateways/class.pmprogateway_stripe.php#L1673 even though the ProfileStartDate already has "T0:0:0" in some cases.

PR #975, opened 2019-05-08
BUG: Now using the correct subscription profile start date for Stripe…

messica:
… subscription updates.

Fixes https://github.com/strangerstudios/paid-memberships-pro/issues/969


messica:
Needs more testing

ideadude:
In what cases was this breaking? Specific versions of PHP?

I think what you did is more clear and shouldn't break in any way that the original wouldn't have been breaking anyway. So I'm going to merge. I'm more interested in the underlying error though to figure out why this fixes it.

PR #998, opened 2019-06-19
Bug Fix: REST API Fatal error

andrewlimaza:
Bug Fix: Fix issue on sites that don't have REST API enabled.

### DIFF
--- a/includes/rest-api.php
+++ b/includes/rest-api.php
@@ -1,56 +1,58 @@
 <?php
-	
-class PMPro_REST_API_Routes extends WP_REST_Controller {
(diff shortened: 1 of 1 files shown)


andrewlimaza:
This just checks to ensure the REST API Class exists first.

ideadude:
I'm not sure you can disable the REST API anymore since core WP functionality ( in the admin at least ) depends on it, but good to do this I guess.

andrewlimaza:
This is to support sites for pre 4.7 (A customer was using WP 4.5) - they have updated which did solve the issue.

Issue #1019, opened 2019-07-19
From Email and From Name are not filtered when 'Only Filter PMPro Emails?' is checked


mjulian87:
The following two filters have no effect as they are never passed to wp_mail if 'Only Filter PMPro Emails?' is checked:
- pmpro_email_sender
- pmpro_email_sender_name

To reproduce:

- Set From Email and From Name in settings
- Check Only Filter PMPro Emails
- Send PMProEmail
- Output : 
  ["From"] => "wordpress@domain"
  ["FromName"] => "WordPress"

mjulian87:
The two filters have no effect at all, regardless of whether 'Only Filter PMPro Emails?' is checked.

pmpro_getOption("from_email") and pmpro_getOption("from_name") work if 'Only Filter PMPro Emails?' is checked, however these don't work either if 'Only Filter PMPro Emails?' is uncheked.

dparker1005:
Resolved here: https://github.com/strangerstudios/paid-memberships-pro/commit/f5f1734b2d5bacad4c0da98d0394ecfa7e905b7e

PR #1021, opened 2019-07-19
Enables pmpro_email_sender and pmpro_email_sender_name filters to work

mjulian87:
### All Submissions:

Previously `pmpro_email_sender` and `pmpro_email_sender_name` filters didn't do anything as their values were never passed to wp_mail. Changing the From Name and From Email worked by using the options set in settings if 'Only Filter PMPro Emails?' was left unchecked as their values were passed straight to the wp_mail filters. With 'Only Filter PMPro Emails?' checked, these values were not passed to wp_mail and the filters did nothing.

Added a new function `pmpro_wp_mail_add_from_filters` to /includes/email.php that adds the filters to wp_mail. Added a wrapper function `pmpro_wp_mail_add_from_filters` to load these filters via the `plugins_loaded` hook (meaning they can be easily removed if needed) if 'Only Filter PMPro Emails?' is unchecked. 

Added a function call to `sendEmail`, after the `pmpro_email_sender` and `pmpro_email_sender_name` filters have been applied, so that the filtered data is actually passed to wp_mail.

Removed `$this->template = apply_filters("pmpro_email_template", $temail->template, $this);` from `sendEmail()` as it didn't do anything.

Closes Issue:   #1019 

The following should now work, whereas previously they didn't.

```
add_filter( 'pmpro_email_sender_name', function() {
	return 'Mickey Mouse';	
});

add_filter( 'pmpro_email_sender', function() {
	return 'mickey@mouse.com';	
});
```

Also, if you check 'Only Filter PMPro Emails?', it should now work, whereas previously it didn't.

### Other information:


ideadude:
This is a good idea. I didn't realize that these filters weren't being used by wp mail. One would assume they would. So it's good to get this working as expected.

A couple things I would look into.

1. Even though $email->template isn't used in the rest of the method after that filter, it is used by other code hooked into the filters that come later. So e.g., someone could be hooking into pmpro_email_body and checking $email->template and maybe expecting an earlier filter to have run on the template.

Could be a very small % of code doing something like that, but we should be careful when removing a filter that's already live.

2. When the filters are added for the from email and name, they override all email sent by wp mail. In a lot of cases, just the next email is the one email being sent, but there might be other email sent afterward that shouldn't be filtered. So we would want to remove the filters after the email is sent to reset the behavior of wp mail for other emails.

Let me know if this makes sense.

mjulian87:
Your points make sense, I have added some notes below. In it's current form, the pull request does not satisfy your second point.

1. All of the functions that call `$this->sendEmail()` appear to already filter `$this->tamplate` before calling `sendEmail()`. So if someone was hooking into `pmpro_email_body` to check the template, `$this->template` should already have the correct value. I can't find a situation where this is not the case.

2. This I had not considered. The issue that I had when coding the pull request was that the from email/name is being filtered from two places. wp_mail and pmpro_email. So to pass the values from the `PMProEmail` Class to wp_mail requires passing new values to filter. How to do this, in a way that allowed the filters to be removed was my issue. I settled on using an action to set the filters, so that if need be, the action could be removed. However this doesn't allow for the filters to be removed once the action has executed.

Because of this issue, the only way I can think of to satisfy your second point would be to filter the name/email just before calling `wp_mail()` and saving the current email/name to a global, then re-filtering after `wp_mail()`.

Not a very elegant solution, but I can't think of another way?

mjulian87:
I've updated the pull request, it now satisfies your second point.

I've added a filter that adds a placeholder to the from name and from email if it's a PMProEmail and then replaces the placeholder with the actual values. After sending the email, the placeholder is removed and the original from name and email are used.

It feels a little hacky, but it's the most elegant solution I could come up with. I used the following code to test and you can see that the first email is filtered and the second is not.

```
add_action( 'phpmailer_init', function( $phpmailer ) {
	
	var_dump( $phpmailer );
});

add_action( 'plugins_loaded', function() {
	
	add_filter( 'pmpro_email_sender', function() {
		
		return 'a@b.com';
	});
	
	add_filter( 'pmpro_email_sender_name', function() {
		
		return 'abc';
	});
	
	$pmproemail = new PMProEmail();
	$pmproemail->sendCheckoutAdminEmail();
	
	wp_mail( 'mickey@mouse.com', 'Email subject test', 'Email body test' );
	
	die;
});
```

I've left `$this->template = apply_filters("pmpro_email_template", $temail->template, $this);` out, but happy to put it back if you think it should stay.

Let me know what you think.

ideadude:
RE #1. I agree. It's very likely no one is using the pmpro_email_template filter in a way where it needs to run in that spot.

RE #2. Adding/removing filters like this is always hacky. I didn't think of this before, but there is probably a simpler solution. We should be able to squeeze the from email and name into the headers passed to wp_mail. If you look at the wp_mail function here, it is extracting the from headers:

https://github.com/WordPress/WordPress/blob/c86ee39ff4c1a79b93c967eb88522f5c09614a28/wp-includes/pluggable.php#L256-L274

I'm a little worried about that "mainly for legacy" comment, but I believe most drop in replacements for wp_mail should/would also honor any from email and name set in the headers.

So basically we can set $temail->headers[] = 'from:name <email>'; after setting the fromname and fromemail. https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/class.pmproemail.php#L115-L116

I don't think we need to check for from: headers before updating, but if it seems likely that people would have set that header earlier, we can account for that. We should probably avoid it for now unless/until someone has that issue.

PR #1036, opened 2019-08-01
moving check for subscription earlier when retrieving subscription from Stripe

LMNTL:
### All Submissions:

* [X] Have you followed the [Contributing guideline](CONTRIBUTING.MD)?
* [X] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

During the getSubscription() function for Stripe, we check whether $this->customer->subscriptions is empty after we call $this->customer->subscriptions->retrieve(), so if the subscription object is empty, it causes an error. This error is handled on PHP <7 due to the enclosing try/catch, but not on PHP 7+. See here: https://www.php.net/manual/en/language.errors.php7.php Need to create separate PR to update all try/catch blocks to handle Throwables.

Forum topic (mods only):
https://www.paidmembershipspro.com/forums/topic/wordpress-notification-experiencing-a-technical-issue/

### Other information:

Test environment: Wordpress 5.2.2, PHP 7.2.15, PMPro 2.0.7

### Changelog entry

BUGFIX: checking for subscription object before retrieving subscriptions by ID when checking at Stripe


ideadude:
Another PR had this change too. But merging to give you credit and fix some whitespace. :)

PR #1048, opened 2019-08-08
Showing admin notice on Delete User Form if any users for deletion have membership.

kimcoleman:
### All Submissions:

This PR adds a notice when admin tries to delete users. It checks if any users for deletion have membership and shows an appropriate message for a single or multiple user deletion informing admin that membership and active subscriptions will be terminated after deletion.

1. Go to Users admin page.
2. Select a single user with no level for deletion. No message should be shown. 
3. Select a single user with a level for deletion. A message should be shown. 
4. Select multiple users, some with a level, for deletion. A different message should be shown. 

### Other information:

### Changelog entry
* ENHANCEMENT: Showing a notice on the Delete User form so admin knows that membership and subscriptions will be cancelled.

### DIFF
--- a/includes/cleanup.php
+++ b/includes/cleanup.php
@@ -21,6 +21,36 @@ function pmpro_delete_user($user_id = NULL)
 add_action('delete_user', 'pmpro_delete_user');
 add_action('wpmu_delete_user', 'pmpro_delete_user');
 
+/**
+ * Show a notice on the Delete User form so admin knows that membership and subscriptions will be cancelled.
+ *
+ * @param WP_User $current_user WP_User object for the current user.
+ * @param int[]   $userids      Array of IDs for users being deleted.
+ */
+function pmpro_delete_user_form_notice( $current_user, $userids ) {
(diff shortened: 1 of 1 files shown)


ideadude:
Neat.

I think that this line

> Deleting a user will also cancel their membership and recurring subscription.

Should be

> Deleting a user will also cancel their membership and any recurring subscriptions at the gateway.

I'm worried a bit about performance. You do break on the first hit (nice!), but checking pmpro_hasMembershipLevel for each user requires a few DB hits. Of course, if the user is a member, we're going to do a similar few DB hits to erase them.

We could instead do one custom DB check like this:

    $sqlQuery = "
    SELECT COUNT(user_id)
    FROM $wpdb->pmpro_memberships_users
    WHERE user_id IN(" . implode( ',', $userids ) . ")
        AND status = 'active'
    GROUP BY user_id;
    ";

Then test if the result is non empty. Bonus is we get a count of how many users are affected, which we could share in the message.

ideadude:
Merging as is and we will address performance if it comes up.

PR #1081, opened 2019-09-04
Fixed entries being skipped in order csv

dparker1005:
### All Submissions:

There were two instances where users were trying to perform large CSV order exports and entries were missing. I believe this line was a typo, but I would like confirmation for other developers to make sure.

Looking at the debug log generated with the old version, the first 2000 orders were being skipped during the export. After this line was changed, all orders were successfully exported. I have, however, noticed some behavior where orders would be exported multiple times.

1. Have orders dataset with more than the iteration limit of orders (defaulted to 2000(
2. Export to CSV
3. Verify that all orders are present in CSV

### Other information:

### Changelog entry

### DIFF
--- a/adminpages/orders-csv.php
+++ b/adminpages/orders-csv.php
@@ -428,7 +428,7 @@ for ( $ic = 1; $ic <= $iterations; $ic ++ ) {
 	}
 
 	//increment starting position
-	if ( $iterations > 1 ) {
+	if ( $ic > 1 ) {
 		$i_start += $max_orders_per_loop;
 	}
 	// get the order list we should process
--- a/adminpages/orders-csv.php
+++ b/adminpages/orders-csv.php
@@ -432,7 +432,7 @@ for ( $ic = 1; $ic <= $iterations; $ic ++ ) {
 		$i_start += $max_orders_per_loop;
 	}
 	// get the order list we should process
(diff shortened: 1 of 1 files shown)


ideadude:
I believe the change you made is needed. As it was coded, it would always skip the first 2000 orders if there were more than 2000 orders.

Note that we have a similar issue in the members list export here: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/adminpages/memberslist-csv.php#L298-L302

The istart stuff is coded slightly differently, but basically the check there needs to test $ic instead of $iterations and only adjust the istart if the current iteration is > 1.

How are the duplicate orders showing up? At the boundaries? So the 2000th order shows up twice? I tried to figure it out from the code, but it seemed to be subtracting 1 as needed. If there are 2000 items per iteration, we want elements 0-1999. I probably missed it though.

dparker1005:
I have just recreated the duplication issue on a local dev site. The site has a total of 262 orders. When I export with the default batch size of 2000, everything works correctly (262 orders exported), but when I decrease the batch size to 100, 323 orders are exported, meaning that there is duplicates.

Looking at this file and the orders in which they were exported, there were order id's 262 down to 1 in order, and then 62 down to 1 in order again. This leads me to believe that the loop is running one iteration too long. I'll update this PR soon with a proposed fix.

dparker1005:
Duplication bug should be fixed with that last commit.

Issue #1136, opened 2019-10-07
BUG: "Now" subscription update creates payment immediately


messica:
**Describe the bug**
Creating a "Now" subscription update cancels the current subscription and creates a new subscription with first payment charged immediately.

**Expected behavior**
Creating a "Now" subscription update should cancel the current subscription and create a new subscription with the same start date as the old subscription.

**Support Ticket (Mods only)**
https://www.paidmembershipspro.com/forums/topic/urgent-now-updates-creating-immediate-duplicate-charges-in-stripe/

ideadude:
Pulled in that PR.

Notes from Jess' PR

> • This creates the new Subscription using the Customer’s default Payment Method instead of a specific Payment Method (because we don’t have payment information available from $_REQUEST during the update). This might cause issues if there is no default Payment Method set. We could fix this by getting the old Payment Method ID and setting it explicitly for the new Subscription.
> • Another way we could fix this is by changing the way Subscription Updates work in general. Stripe’s API does allow us to update Subscriptions so we could actually update it with 1 request instead of cancelling, creating a new Plan, and subscribing to the new plan. This might be the easiest and cleanest way to do Subscription Updates but I’m not sure if that might conflict with other code in another plugin, etc

PR #1137, opened 2019-10-07
BUG FIX: Preserving trials during 'now' Subscription Updates.

messica:
### All Submissions:

* [ ] Have you followed the [Contributing guideline](CONTRIBUTING.MD)?
* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Closes Issue: XXX.

1.
2.
3.

### Other information:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?

### Changelog entry

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1697,6 +1697,7 @@ class PMProGateway_stripe extends PMProGateway {
 		//build order object
 		$update_order = new MemberOrder();
 		$update_order->setGateway( 'stripe' );
+		$update_order->code             = $update_order->getRandomCode();
 		$update_order->user_id          = $user_id;
 		$update_order->membership_id    = $user_level->id;
 		$update_order->membership_name  = $user_level->name;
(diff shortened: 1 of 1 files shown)


messica:
so this PR https://github.com/strangerstudios/paid-memberships-pro/pull/1137 is the easy fix, but there are a few issues and we might want to handle subscription updates differently because of them:
• This creates the new Subscription using the Customer’s default Payment Method instead of a specific Payment Method (because we don’t have payment information available from $_REQUEST during the update). This might cause issues if there is no default Payment Method set. We could fix this by getting the old Payment Method ID and setting it explicitly for the new Subscription.
• Another way we could fix this is by changing the way Subscription Updates work in general. Stripe’s API does allow us to update Subscriptions so we could actually update it with 1 request instead of cancelling, creating a new Plan, and subscribing to the new plan. This might be the easiest and cleanest way to do Subscription Updates but I’m not sure if that might conflict with other code in another plugin, etc

messica:
fixes https://github.com/strangerstudios/paid-memberships-pro/issues/1136

PR #1159, opened 2019-10-21
Improve Block Editor code and webpack.

andrewlimaza:
### All Submissions:

### Other information:

### Changelog entry

* ENHANCEMENT: Fixed webpack to make use of WP Scripts package instead of a custom one.
* ENHANCMENT: Improved block JS and block CSS to be simpler and easier to maintain.
* BUG FIX: Error with webpack causing blocks to error within the new Block Editor.


andrewlimaza:
This removes all blocks CSS files and creates one master CSS file. Blocks all use the same class now as the CSS was redundant and we weren't doing anything 'crazy' with the blocks in terms of styling. 99% of styling was the same between all blocks with an exception of 1-2 blocks. 

This has been catered for now. Looks like a lot of changes, but a lot of files have been removed and optimized.

A lot of code was not being used inside the blocks which was removed to tidy up the code base.

ideadude:
Seems we are missing some bits for the localizing in JS

https://make.wordpress.org/core/2018/11/09/new-javascript-i18n-support-in-wordpress/

Localizing the strings in the JS would make our minimum requirement 5+ unless we detected if the wp_set_script_translations function existed before using it. We are probably already testing for Gutenberg or WP 5.0+ before loading blocks. I don't think any of the localizing in JS in here is done to anything but the block JS.

ideadude:
... and so we should probably wait for a point release before merging in.

PR #1165, opened 2019-10-30
Hide notifications / Rules

andrewlimaza:
Added in show_if/hide_if rules.

* Hide notification if a plugin is active.
* Show a notification if a plugin is a specific version.

TODO: Support multiple plugins.


andrewlimaza:
This includes only show security notification setting and some logic around this.

Screen Shot 2019-10-30 at 12 57 46

PR #1167, opened 2019-11-07
Notifications

andrewlimaza:
Improved code for rules.

Created the following functions:

pmpro_check_active_plugins
pmpro_check_plugin_version

### DIFF
--- a/includes/notifications.php
+++ b/includes/notifications.php
@@ -154,8 +154,9 @@ function pmpro_is_notification_applicable( $notification ) {
 		return false;
 	}
 	
+
 	// Need to finish the below, but just show for now.
-	return true;
+	// return true;
 	
 	// Hide notification by default.
 	$show_notification = false;
@@ -198,6 +199,20 @@ function pmpro_is_notification_applicable( $notification ) {
 		} else {
 			$show_notification = false;
 		}
+
+		// If we're already showing it, just show it.
+		if ( $show_notification ) {
+			return $show_notification;
+		}
+
+		$plugin_slug_and_file = $show_if->plugins_active[0] . '/' . $show_if->plugins_active[0] . ".php";
+
+		if ( is_plugin_active( $plugin_slug_and_file ) ) {
+			$show_notification = true;
+		}
+
+
+
 	}
 	return $show_notification;
 }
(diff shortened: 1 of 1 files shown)


includes/notifications.php
```diff
@@ -262,3 +276,42 @@ function pmpro_footer_link() {
...
+	$pass = false;
+
+	// run check if plugin data is not emptpy.
+	if ( $checks ) {
+		$plugin_file = $checks[0] . '/' . $checks[0] . '.php';
+		$plugin_data = get_plugin_data(  WP_PLUGIN_DIR . '/' . $plugin_file, false, true );
+
+		$conditional_check =  $checks[2] . " " . $checks[1] . " " . $plugin_data['Version'];
+
+		if ( $conditional_check ) {
```
eighty20results: This will never be false (empty) even if the checks array is empty and the $plugin_data['Version'] is empty. It's minimally always a concatenated string of 2 spaces.

includes/notifications.php
```diff
@@ -262,3 +276,42 @@ function pmpro_footer_link() {
...
+/**
+ * Check if a plugin is active with a specific version.
+ * @param array $checks Plugin data to run a check. Includes: [ 0 => plugin-slug, 1 => comparison operator, 2 => version_to_check]
+ */
+function pmpro_check_plugin_version( $checks = NULL ) {
+	// Assume pass is failing right now.
+	$pass = false;
+
+	// run check if plugin data is not emptpy.
+	if ( $checks ) {
```
eighty20results: Even if you're checking that there's something in $checks, the code may still trigger PHP notices since there are no checks for whether $checks[0] and $checks[1] actually exist.

includes/notifications.php
ideadude: I like that you're interested in this. :) Thanks for pointing these out.

This code is very much incomplete and we're abusing PRs to pass the code back and forth between andrew and I in a public way. I'm working on it more today and will reach out to you for a once over if you are interested. We might push it next week.

includes/notifications.php
ideadude: FYI https://github.com/strangerstudios/paid-memberships-pro/commit/f82f458bee963c5367f6a03a982ae7bbe5279e6a

ideadude: Going to make a lot more updates today yet

PR #1193, opened 2019-12-09
Removing deprecated stripe-php function calls

woodwardmatt:
### All Submissions:

This change removed calls to a deprecated function in the Stripe-PHP library.

The deprecated call can be seen here:  https://github.com/strangerstudios/paid-memberships-pro/blob/00e7e8130ad667808c161f0c3964ae732798a3df/classes/gateways/class.pmprogateway_stripe.php#L1940

This function (i.e. Stripe\Customer::invoices()) was deprecated in the Stripe PHP Library in this commit: https://github.com/stripe/stripe-php/commit/8a046cfd119e50a048645f8d4249490a566379bc#diff-814438c52fc09aaad9ad47d5fa40351c

Having queries this with the Stripe support team, the recommended approach is now to use the list function on the Invoice object with the appropriate customer id and status (of open). This approach is documented at:  https://stripe.com/docs/api/invoices/list?lang=php

1. Create a new member & membership
2. Whilst logged in as that member, choose to cancel the membership
3. See error "There has been a critical error on your website." (prior to change, successful cancellation after change.)

### Other information:

### Changelog entry

Remove calls to deprecated stripe-php function Stripe\Customer::invoices() used during the cancellation of a membership.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1937,8 +1937,7 @@ class PMProGateway_stripe extends PMProGateway {
 		$this->getCustomer( $order );
 
 		// Get open invoices.
-		$invoices = $this->customer->invoices();
-		$invoices = $invoices->all();
+		$invoices = Stripe_Invoice::all(['customer' => $this->customer->id, 'status' => 'open'])
 
 		// Found it, cancel it.
 		try {


woodwardmatt:
This relates to a new fatal error at: https://github.com/strangerstudios/paid-memberships-pro/issues/1194

ideadude:
Thanks again for finding this.

We are using a slightly older version of the PHP API that does still have that method in the Customer class: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/lib/Stripe/lib/Customer.php#L82-L93

If you are having this issue on a site, it might be because another plugin or bit of code loaded the newer Stripe API before PMPro did. If the API is already loaded, we try to use it, but obviously there can be issues if the versions don't match.

You can maybe solve that conflict on your site or use this patch if it works for you.

I'd like to update to the latest API and adopt this new method, but we'd also have to update the whole API, which might impact other things. So we'll need to do further testing.

We generally don't update the API until we "need to" for something in our code, but it's good in general to be on the latest version to avoid things like this. So it's good to upgrade.

At the same time, upgrading to a version that has removed these deprecated methods could cause other issues. Even if we aren't using those old methods, maybe another plugin using Stripe and expecting an older API would. Still, in general, Stripe is not developing their API to be backwards compatible to avoid integration conflicts in WP plugins.

TL;DR; We need to update the entire Stripe API to the latest version and research and test to make sure that doesn't cause any other issues. Then it would make sense to apply this patch.

woodwardmatt:
> If you are having this issue on a site, it might be because another plugin or bit of code loaded the newer Stripe API before PMPro did. If the API is already loaded, we try to use it, but obviously there can be issues if the versions don't match.

You're quite right, I have a version in the root of my site that I recently updated to the latest version. I can see that you've got the older stripe-php included with that call still in it, so it must be my root version overriding it - thanks for the heads up :)  Totally missed that dynamic in the hunt for resolving it yesterday :)

ideadude:
I'm going to close this PR. When we update the Stripe Library included with PMPro (not sure when that will happen, but eventually it will) this will take care of itself.

ideadude:
I reopened this and set it to merge into our new stripe-lib branch.

We should do some more testing if there are other cases like this where the SDK method changed.

Maybe: $customer->subscriptions...
https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/gateways/class.pmprogateway_stripe.php#L1430

Or $customer->invoice_settings...
https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/gateways/class.pmprogateway_stripe.php#L1867

Issue #1174, opened 2019-11-18
WP timezone changing expiration months


dparker1005:
**Describe the bug**
Changing the timezone in WordPress general settings seems to be able to offset PMPro expiration dates by a month.

With quick testing, this is what I've seen:
American timezone --> European timezone : Expiration dates are a month sooner
European timezone --> American timezone : Expiration dates are a month later

**To Reproduce**
1. Make a note of one of your user's expiration dates
2. Go to WordPress general settings
3. Change timezone (not sure exactly which ones cause this error, may need some trial and error)
4. Check if user's expiration date has been modified by a month

**Expected behavior**
Expiration date should not change based on the timezone

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only Paid Memberships Pro plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Memberlite](https://wordpress.org/themes/memberlite/).
- [x] I can reproduce this bug consistently using the steps above.

**Admins**
Tickets that may have this issue (both of these have seen the month be modified upon saving the user's "Edit User" page after setting an expiration date):
https://www.paidmembershipspro.com/forums/topic/issue-when-creating-a-user-and-adding-a-membership-level-and-mailchimp/
https://www.paidmembershipspro.com/forums/topic/expires-date-going-haywire/

dparker1005:
WP 5.3 changelog blog post: https://wordpress.org/support/topic/read-this-first-wordpress-5-3-master-list/#post-12124062

`Dates or times being incorrect, permalinks being off by a day, other weirdness related to “time”

WordPress 5.3 received a significant upgrade to its date and time handling code. The previous code dated all the way back to over 10 years ago, when PHP 4 was still supported. This was a significant chunk of code scattered all through the core of WordPress, and it was very slow and a drag on performance.

With 5.3, most of this code has been updated or replaced with saner PHP 5.6 compatible code. A lot of the legacy has been removed. However, some plugins and themes may have relied on the side effects from how this code previously operated.

For this reason, some plugins or themes may be doing incorrect things. One such thing would be to call the date_default_timezone_set() function in PHP. WordPress works by setting the default timezone to UTC and then performing its own calculations to adjust times. Setting the default timezone incorrectly to anything else will result in these calculations being incorrect. WordPress will be overcompensating the timezone adjustments.

[...]

PR #1217, opened 2020-01-16
Membership profile: fix month calculation in expiry dropdown #1174

arnalyse:
### All Submissions:

This commit uses the in WordPress 5.3 newly introduced wp_timezone and wp_date functions, to calculate the timestamp for the first of the month while respecting the timezone from WordPress' settings.

The main difference is, that the timestamp created is fed with the wp_timezone, and thus the timezone isn't subtracted or added later on. 

Furthermore, this removes the – now legacy from 5.3 on – `date_i18n()` call.

Closes Issue: 1174.

1. Set WP timezone to somewhere in Europe (this causes the issue on my personal setup)
2. Ensure that months when setting an expiration date begin with January

### Other information:

### Changelog entry

> Membership profile: fix month calculation in expiry dropdown

### DIFF
--- a/includes/profile.php
+++ b/includes/profile.php
@@ -131,8 +131,10 @@ function pmpro_membership_level_profile_fields($user)
 						<?php
 							for($i = 1; $i < 13; $i++)
 							{
+								$first_of_month = date_create($i . "/1/" . $current_year, wp_timezone());
+                                				$option_month = wp_date( "M",  $first_of_month->getTimestamp());
 							?>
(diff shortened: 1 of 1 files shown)


ideadude:
Thanks for this work. There is more work we want to do in general to get date and time math working properly in the plugin.

Since the wp_date function is only available in WP 5.3+, we want to make sure that we roll this out in a way that will give everyone time to upgrade properly. There are still quite a few sites running PMPro on older versions of WP and PHP.

ideadude:
Thanks again for the PR. This should be addressed by other commits related to our timestamp fixes in general.

PR #1219, opened 2020-01-20
Stripe Billing Page Fix for Non-recurring members.

andrewlimaza:
* Added pmpro_are_levels_recurring.
* Adjusted conditions to load stripe.js on billing page based on user's level status.


ideadude:
Going to adjust how the caching works, but pulling for now.

PR #1244, opened 2020-02-21
Adding Advanced Setting to Hide Admin Bar.

kimcoleman:
### All Submissions:

Adds new setting to hide admin bar from non-administrator users. Users can still access this setting on the "Edit Profile" admin page. We may want to consider adding the CSS to hide that row.

### Changelog entry
ENHANCEMENT: Adding an Advanced Setting to "Hide the Toolbar from non-Administrator users.".

### DIFF
--- a/adminpages/advancedsettings.php
+++ b/adminpages/advancedsettings.php
@@ -44,6 +44,9 @@
 		//tos
 		pmpro_setOption("tospage");
 
+		//toolbar
+		pmpro_setOption( 'hide_toolbar' );
+
 		//footer link
 		pmpro_setOption("hide_footer_link");
 
@@ -80,6 +83,8 @@
 	$recaptcha_privatekey = pmpro_getOption("recaptcha_privatekey");
 
 	$tospage = pmpro_getOption("tospage");
+
+	$hide_toolbar = pmpro_getOption( 'hide_toolbar' );
 
 	$hide_footer_link = pmpro_getOption("hide_footer_link");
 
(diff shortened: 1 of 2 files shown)


tictag:
There is an issue where plugins that hide the admin dashboard (I assume this to be the same as 'admin bar') conflict with the Membership Manager Role Add On in that if the plugin is active, a user with the Membership Manager role assigned cannot access PMPro functions because the admin dashboard is hidden from them. Ticket [https://www.paidmembershipspro.com/forums/topic/help-membership-manager-add-on-incompatible-with-restricted-admin-dashboard/] discusses this incompatibility.

(1)  Would this PR also hide the admin dashboard from users with the Membership Manager role?

(2)  Would this 'additional setting' effectively replace the need for the plugin (e.g. Drew Jaynes' Remove Dashboard Access plugin)?

kimcoleman:
This update will not block admin dashboard access for any user role. This update is only to hide the "Toolbar". You can read about the toolbar filter we are using in the WordPress developer docs here: https://developer.wordpress.org/reference/functions/show_admin_bar/

We do not plan to block dashboard access for any user roles in core PMPro at this time, so using the "Remove Dashboard Access" plugin may still be needed if you have replicated all features your default site role needs with a frontend features.

tictag:
Ahh, OK, the 'admin toolbar' is the horizontal page element containing icons for various plugins etc.

Not sure why PMPro would have a feature to modify this but I'm sure you have your reasons.

As the PMPro 'Membership' menu is also available via this toolbar, you might want to make sure the 'Membership' menu remains available if the Membership Manager Add On is installed.

PR #1290, opened 2020-04-03
Change wp_login_url to use pmpro_login_url

andrewlimaza:
### All Submissions:

* Change !!login_link!! to use pmpro_login_url() instead for email class.f accepted.


ideadude:
Some of the old calls also pass a redirect URL to redirect back to the account page. I suppose that's the default for login now, should we use `pmpro_login_url( '', pmpro_url( 'account' ) )` for these? Maybe it doesn't matter?

andrewlimaza:
`pmpro_login_url` forces our URL to redirect to the account page as of V2.3 and is filterable if needed? Let me know your thoughts on this.

Code ref: https://github.com/strangerstudios/paid-memberships-pro/blob/v2.3/includes/login.php#L90-L101

PR #1342, opened 2020-05-02
PMPRO_IPN_DEBUG logic refactoring

mircobabini:
### All Submissions:

- Some refactoring;
- Logic fix: should avoid counterintuitive interpretation of false;
- Filter on ipn_logfile: I want to keep the history, can't log into pmpro folder;

### Other information:

### DIFF
--- a/services/ipnhandler.php
+++ b/services/ipnhandler.php
@@ -378,7 +378,7 @@ function pmpro_ipnExit() {
 			} elseif ( is_email( PMPRO_IPN_DEBUG ) ) {
 				//email to specified address
 				wp_mail( PMPRO_IPN_DEBUG, get_option( "blogname" ) . " IPN Log", nl2br( $logstr ) );							
-			} elseif ( !!PMPRO_IPN_DEBUG ){
+			} else {
 				//email to admin
 				wp_mail( get_option( "admin_email" ), get_option( "blogname" ) . " IPN Log", nl2br( $logstr ) );							
 			}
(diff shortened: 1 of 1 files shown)


services/ipnhandler.php
```diff
@@ -361,22 +361,27 @@ function pmpro_ipnExit() {
...
+			} elseif ( PMPRO_IPN_DEBUG === "log" ) {
+				//file
+				$logfile = apply_filters( 'pmpro_ipn_logfile', dirname( __FILE__ ) . "/../logs/ipn.txt" );
+				$loghandle = fopen( $logfile, "a+" );
+				fwrite( $loghandle, $logstr );
+				fclose( $loghandle );
+			} elseif ( is_email( PMPRO_IPN_DEBUG ) ) {
+				//email to specified address
+				wp_mail( PMPRO_IPN_DEBUG, get_option( "blogname" ) . " IPN Log", nl2br( $logstr ) );							
+			} elseif ( !!PMPRO_IPN_DEBUG ){
```
ideadude, line 381: Why !!PMPRO_IPN_DEBUG instead of just else {} or elseif ( PMPRO_IPN_DEBUG === true )?

mircobabini: Well, I think "PMPRO_IPN_DEBUG === true" can be ok. Actually !!PMPRO_IPN_DEBUG is a cast to boolean so it's the same thing.

ideadude: I like using an else {}

PMPRO_IPN_DEBUG is defined, it's not 'log' or an email address, let's assume it's true or 1 or yes and people meant for an email to be sent.

Can you update? Then I'll merge in.

mircobabini: Done!

PR #1344, opened 2020-05-04
Updating the advanced settings page to remove message about "EasyAdsense"

kimcoleman:
### All Submissions:

This PR removes the line "Ads from the following plugins will be automatically turned off: Easy Adsense" since this is no longer an active plugin and may be misleading as there are newer plugins that we do not integrate with that are called "Easy Google Adsense" and "Easy Peasey Adsense".

The function itself still checks for the presence of the `EzAdSense` class. We should look to deprecate this integration and consider other new ad plugins we can have built-in filters for.  

### Other information:

### Changelog entry
* BUG FIX/ENHANCEMENT: Removed text on Advanced Settings for "Hide Ads" section where it may be misleading that we integrate with newer Adsense plugins (Easy Adsense is no longer in development).
 
### DIFF
--- a/adminpages/advancedsettings.php
+++ b/adminpages/advancedsettings.php
@@ -341,12 +341,11 @@
 				<tr id="hideads_explanation" <?php if($hideads < 2) { ?>style="display: none;"<?php } ?>>
 					<th scope="row" valign="top">&nbsp;</th>
 					<td>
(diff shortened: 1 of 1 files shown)


ideadude:
While we are at it, should we add a function_exists check to our sample code there?

PR #1358, opened 2020-06-01
Adding the current user's payment method used for subscription to Billing Information page

kimcoleman:
Now showing the payment method to be used or just updated so the users know that their payment method update was successful.

There is a filter `pmpro_billing_show_payment_method` to disable the display for sites using a service like Churnbuster that can get out of sync with the user's usermeta values for card type, account number, and expiration date.

Users previously had no way to see what card they were paying with or whether the update was successful after update.

### All Submissions:

### Other information:

### Changelog entry

BUG FIX/ENHANCEMENT: Added a notice of the payment method being used for subscription on the Membership Billing page.

### DIFF
--- a/pages/billing.php
+++ b/pages/billing.php
@@ -79,12 +79,12 @@
 			<?php
 				$pmpro_billing_show_payment_method = apply_filters( 'pmpro_billing_show_payment_method'
 					, true);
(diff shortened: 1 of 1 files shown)


ideadude:
Going to merge and then escape. Even though we're pretty certain of these values, we should always escape.

Issue #1130, opened 2019-10-01
Support MAXFAILEDPAYMENTS for PayPal Gateways.


andrewlimaza:
**Is your feature request related to a problem? Please describe.**
We should consider handling failed payment limits in core.

**Describe the solution you'd like**
Users would like to cancel user's membership levels if the user's recurring payments failed. We can  maybe set this to 2-3 failed payments and filterable?

**Describe alternatives you've considered**
Failed Payment Limit Add On.

**Additional context**
https://developer.paypal.com/docs/classic/express-checkout/integration-guide/ECRecurringPayments/#maximum-number-of-failed-payments

mircobabini:
Commit to set MAXFAILEDPAYMENTS=1

0620227

Done.

PR #1385, opened 2020-06-30
Fix total amount of recurring payment order from PayPal Express

mircobabini:
### All Submissions:

Check the mc_gross values before the amount value. This is because when a recurring payment comes with a cumulative price (because of one or more previous missed payments) the amount value is the recurring amount, but the mc_gross value is the total, comprehensive of the previous missed payments. This is the real transaction price.

Check this up: [gist](https://gist.github.com/mircobabini/9a02f185e895d4e999371d179ed7ec0b)

1. You have to generate a recurring payment transaction for a cumulative amount. Or you can just trust me :)

### Other information:

### DIFF
--- a/services/ipnhandler.php
+++ b/services/ipnhandler.php
@@ -723,6 +723,10 @@ function pmpro_ipnSaveOrder( $txn_id, $last_order ) {
 			//check for tax
 			if ( isset( $_POST['tax'] ) && ! empty( $_POST['tax'] ) ) {
 				$morder->tax = (float) $_POST['tax'];
+				if ( isset( $_POST['amount'] ) && ! empty( $_POST['amount'] ) && $morder->InitialPayment > (float) $_POST['amount'] ) {
+					$morder->tax *= (float) $morder->InitialPayment / (float) $_POST['amount'];
+				}
+
 				$morder->total = $morder->InitialPayment;	//so tax isn't added into the subtotal again
 				$morder->subtotal = $morder->total - $morder->tax;
 			}
--- a/services/ipnhandler.php
+++ b/services/ipnhandler.php
@@ -709,12 +709,12 @@ function pmpro_ipnSaveOrder( $txn_id, $last_order ) {
 		//set amount based on which PayPal type
 		if ( false !== stripos( $last_order->gateway, "paypal" ) ) {
 
-			if ( isset( $_POST['amount'] ) && ! empty( $_POST['amount'] ) ) {
-				$morder->InitialPayment = $_POST['amount'];    //not the initial payment, but the class is expecting that
-				$morder->PaymentAmount  = $_POST['amount'];
-			} elseif ( isset( $_POST['mc_gross'] ) && ! empty( $_POST['mc_gross'] ) ) {
+			if ( isset( $_POST['mc_gross'] ) && ! empty( $_POST['mc_gross'] ) ) {
 				$morder->InitialPayment = $_POST['mc_gross'];    //not the initial payment, but the class is expecting that
 				$morder->PaymentAmount  = $_POST['mc_gross'];
+			} elseif ( isset( $_POST['amount'] ) && ! empty( $_POST['amount'] ) ) {
+				$morder->InitialPayment = $_POST['amount'];    //not the initial payment, but the class is expecting that
+				$morder->PaymentAmount  = $_POST['amount'];
 			} elseif ( isset( $_POST['payment_gross'] )  && ! empty( $_POST['payment_gross' ] ) ) {
 				$morder->InitialPayment = $_POST['payment_gross'];    //not the initial payment, but the class is expecting that
 				$morder->PaymentAmount  = $_POST['payment_gross'];


ideadude:
I have to do some more thinking about this one.

Some time way back when, we switched to using amount instead of mc_gross. I don't recall why we did that.

https://github.com/strangerstudios/paid-memberships-pro/commit/e5b2f441ec98bd5ab5267b36caee6a5f74e3217b#diff-ee3ac4366bc5b03163bdb17970a26a4d

(That's a bit complicated commit that wasn't broken down.)

Maybe mc_gross is sometimes not passed? If so, we could check for it first and fallback on amount instead. Maybe there were cases where "credit" was used to pay for the order and it made it seem like the users paid $0 or a lower amount than the full amount they were supposed to pay.

This is an interesting case where you want the order or to match up with what happens at PayPal really, including additional payments or discounts/etc, but you also want the order to match what people expect based on how the subscription was set up.

mircobabini:
> Some time way back when, we switched to using amount instead of mc_gross. I don't recall why we did that.

Checked the diff... yeah, don't know. Maybe sometimes is not passed, yes, maybe.

> If so, we could check for it first and fallback on amount instead

That's right. And that's what you were already doing since that piece of code is unique for all the \*paypal* gateways, I just had to change the order. Instead of looking for the amount first, you have to look for the gross and then fallback on the amount if eventually missing.

> This is an interesting case where you want the order or to match up with what happens at PayPal really

I must disagree. Not an interesting case at all, it's a consequence of a pmpro bug imho. Do you remember that you do not pass the parameter to tell PayPal Express to STOP the subscription if there's no enough money to renew it? This is causing 2 veeeery bad situations:
1. Users with no money at the beginning, subscription created with status = error. When some day they re-charge their PayPal account, they got charged of all the unpaid (even after years);
1. Users with no money at the renewal, they will still access the membership (!!). Yeah, for free! And when soem day they re-charge their PayPal account, they got charged of all the unpaid.

This is what made me go crazy in the last weeks. And that's why we have the issue above.
If you skip a month, then recharge later, then PayPal charges you 2 months in a single time you must log the correct price.
If you don't, you will have 2 related issues at least:
- wrong reports in pmpro
- wrong invoice in pmpro

Then:
> but you also want the order to match what people expect based on how the subscription was set up

Yes and no. If you really want the order to match what people expect, just remove users from membership and stop their subscription at gateway, when they have no money. There is no service that permits you to keep your membership even if you don't pay.

If we fix that bug, we won't have to worry about double/triple payments anymore.

mircobabini:
I'm referring to this one #1130: MAXFAILEDPAYMENTS

I think the Difficult should be set to Easy and the Impact to High :)
Just a matter of adding a filter and an options.

That's it!

ideadude:
Thanks for the info, the help, and your patience. This won't make the next update, but I will try to address it soon.

You can run your sites off your branch and merge dev or master into it to get our updates.

ideadude:
I think we're going to miss getting this into another update, but I will mark it as "blocking next release" after this push.

I think it makes sense to merge as is, but it also is tied up in other things we're working on and thinking about. In general, we want to do work to make sure that the data in the PMPro orders is good enough to "calculate your taxes by". In the past, we would say "the gateway is the source of truth" and that makes sense since so many things can run through the gateway that we don't know about... refunds, extra purchases, cases like this one. But it's a bit of a cop out. We need our data to be as tight as possible with the gateway and having the order reflect the actual amount charged (duh) makes sense.

Another related things we've been thinking through is basically avoiding this case where PayPal is carrying over payments. It's a nice feature, but confusing to maintain through code and confusing for users. (Why was I charged double?) I think we might set MAXFAILEDPAYMENTS to 1 and basically cancel the subscription right away on the first failed payment. Users will get an email asking them to check out again.

In our experience, the automatic retry attempts rarely work. Users can update their billing, but that requires their input and isn't so different form just checking out again. Some other addons like the proration one or cancel on next payment date get confused by these users who missed a payment and in the limbo of not being cancelled yet while payments are retried. So just cancelling on the date of the expected payment would avoid some of that confusion, including cases this PR tries to address.

Let me know your thoughts on switching to MAXFAILEDPAYMENTS=1 here in PayPal, and in general. Do you feel that would be an issue on your site?

mircobabini:
Short answer is no, not an issue at all. I think you should just change MAXFAILEDPAYMENTS=1 as soon as possible (#1130); it's 100% the best way to move on. 

In fact, I had to write custom code to force the gateway cancellation on first payment failed and it works good on my platform. The users are less confused, sometimes they just ask "why my subscription is not working anymore?" then I check the gateway and boom, their credit card are just empty. They will check out again and we are done, simple as that. Most of the time they sort it out by themselves.

I totally agree, thanks for setting it as "blocking next release".

ideadude:
Commit to set MAXFAILEDPAYMENTS=1

https://github.com/strangerstudios/paid-memberships-pro/commit/06202279f02c98b121a393bd9c9556db6c8eabd5

This is supposed to tell PayPal to cancel a subscription on the first failed payment. While looking into this, I saw all the old comments on forums about how PayPal sometimes ignores this parameter, so we will see how it works in practice for new subscriptions.

Will take another look at the mc_gross fix/PR before next release.

mircobabini:
> Some time way back when, we switched to using amount instead of mc_gross. I don't recall why we did that.
> 
> [e5b2f44#diff-ee3ac4366bc5b03163bdb17970a26a4d](https://github.com/strangerstudios/paid-memberships-pro/commit/e5b2f441ec98bd5ab5267b36caee6a5f74e3217b#diff-ee3ac4366bc5b03163bdb17970a26a4d)
> 
> (That's a bit complicated commit that wasn't broken down.)
> 
> Maybe mc_gross is sometimes not passed? If so, we could check for it first and fallback on amount instead.

Correct, the `mc_gross` is sometimes not passed. Please check [Table 1 "Summary of recurring payment variables"](https://developer.paypal.com/docs/api-basics/notifications/ipn/IPNandPDTVariables/).

amount is passed for _Profile created message_ and _Recurring payment message_ as well.
mc_gross is passed for _Recurring payment message_ only.

But as reported in that page:

> Full amount of the customer's payment, before transaction fee is subtracted. Equivalent to payment_gross for USD payments. If this amount is negative, it signifies a refund or reversal, and either of those payment statuses can be for the full or partial amount of the original transaction.

In conclusion:
- _mc_gross_ is the only one which consider an eventual sum of transactions because of missing credit in previous cycles;
- we want to use _mc_gross_, falling back on _amount_.

Hope this can help to close this one.

ideadude:
Confirmed all of the above and merging in for next release. Sorry this one took so long.

PR #1388, opened 2020-07-02
Now using CONVERT_TZ for start and enddates

dparker1005:
### All Submissions:

Now using CONVERT_TZ SQL function when getting `startdate` and `enddate` from database. Query should now get the time from the DB in local time, meaning that no additional conversions are required. Fixes issues were SQL time settings could cause off-by-one error when showing enddate on profile page.

### Other information:

### Changelog entry

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1836,8 +1836,8 @@ function pmpro_getMembershipLevelForUser( $user_id = null, $force = false ) {
 				mu.trial_amount,
 				mu.trial_limit,
 				mu.code_id as code_id,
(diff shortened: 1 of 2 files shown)


ideadude:
GREAT.

Going to merge this.

Note that we are kicking down the road the issue of what to do when user's change the WP timezone. In some cases, they would want the dates to adjust for the new timezone and they may or may not do that here and throughout PMPro.

We want to adjust any other time we use UNIX_TIMESTAMP in a MySQL query to also call CONVERT_TZ.

PR #1419, opened 2020-07-24
Removing unused JS from select field.

kimcoleman:
Unused JS function was causing a console error on the admin discount codes page.


ideadude:
handled in another commit

PR #1426, opened 2020-07-31
Adding a hook to the member profile edit form tag.

kimcoleman:
New hook pmpro_member_profile_edit_form_tag replicates the core WP hook here: https://developer.wordpress.org/reference/hooks/user_edit_form_tag/

This hook is used by Register Helper here: https://github.com/strangerstudios/pmpro-register-helper/blob/b35bb8b2e87acab6d0276e4c1a7f743703613f2f/pmpro-register-helper.php#L1064-L1071

We should test file uploads in pmpro without that form type on the file upload - they may work as is. Most avatar upload systems need this encoding on the form so that we can support a plugin like Simple Local Avatars on our frontend form.

### DIFF
--- a/includes/profile.php
+++ b/includes/profile.php
@@ -480,7 +480,16 @@ function pmpro_member_profile_edit_form() {
 	}
 	?>
 	<div class="<?php echo pmpro_get_element_class( 'pmpro_member_profile_edit_wrap' ); ?>">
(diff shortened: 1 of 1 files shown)


kimcoleman:
Here's an example recipe using the new filter to add today's date before the output of the content: https://gist.github.com/kimcoleman/e2c7d0b709ab6b9bb3d3c86421a7f2b8

PR #1467, opened 2020-08-27
Fixed issue where confirmation message wouldnt be sent in checkout email

dparker1005:
### All Submissions:

We recently made a change to hide the confirmation message from the checkout level so that users couldn't possibly see confirmation message until they have that level. This caused an issue where the confirmation wouldn't be shown in the confirmation email.

This change adds the confirmation back into the level object after the user's membership level is successfully changed during checkout, fixing the email issue and potentially other issues with the confirmation message.

Check box on edit level page to send confirmation message in email, and then check out for that level with the default PMPro email templates. See that the confirmation is shown with this fix, but not without.

### Other information:

### Changelog entry

### DIFF
--- a/preheaders/checkout.php
+++ b/preheaders/checkout.php
@@ -641,6 +641,9 @@ if ( ! empty( $pmpro_confirmed ) ) {
 
 		if ( pmpro_changeMembershipLevel( $custom_level, $user_id, 'changed' ) ) {
 			//we're good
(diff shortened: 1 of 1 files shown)


ideadude:
Instead of this, I'm going to move the logic to hide the confirmation message into the REST API endpoint instead of always doing it in the function call. Makes more sense that way I think.

PR #1485, opened 2020-10-01
applydiscountcode.php mmpu compatibility

dparker1005:
### All Submissions:

Fix issues in MMPU where correct price would may be shown after discount code is applied and where discount code may not apply successfully if it doesn't work for all membership levels:
https://github.com/strangerstudios/pmpro-multiple-memberships-per-user/issues/47

### Other information:

^ Want to do more testing.

### Changelog entry

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1697,7 +1697,7 @@ function pmpro_checkDiscountCode( $code, $level_id = null, $return_errors = fals
 	 *
 	 * @param bool $okay true if code check is okay or false if there was an error
 	 * @param object $dbcode Object containing code data from the database row
-	 * @param int $level_id ID of the level the user is checking out for.
+	 * @param int|array $level_id ID of the level the user is checking out for.
 	 * @param string $code Discount code string.
 	 *
 	 * @return mixed $okay true if okay, false or error message string if not okay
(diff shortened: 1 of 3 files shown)


dparker1005:
This PR is not ready to merge. While using MMPU, the discount code now applies OK even if it doesn't apply to all levels at checkout, but checkout does not complete successfully unless the discount code applies to the "main" checkout level. Here are the lines that would need a fix:
- https://github.com/strangerstudios/paid-memberships-pro/blob/1ade8bdab647f86970fe50b0e9b8f03246d6aaee/preheaders/checkout.php#L607
- https://github.com/strangerstudios/paid-memberships-pro/blob/1ade8bdab647f86970fe50b0e9b8f03246d6aaee/includes/functions.php#L2230

ideadude:
David pointed out we also need to fix preheaders/checkout.php so checkout works.

We can add a check here for the $pmpro_checkout_level_ids globsl that MMPU sets https://github.com/strangerstudios/paid-memberships-pro/blob/dev/preheaders/checkout.php#L607 and pass that or $pmpro_level->id to the code check function.

dparker1005:
The two `pmpro_checkDiscountCode()` calls mentioned previously should now be fixed as per @ideadude 's suggestion.

PR #1502, opened 2020-11-08
Move € symbol to the left for de_* locales

### LINKED ISSUE #1473: Euro currency position for Deutsch locale
**Describe the bug**
A user has suggested that the location of the currency symbol for the Euro when using the Deutsch locale should be on the left. The function `pmpro_euro_position_from_locale` could be extended to support this locale and currency position.

https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/currencies.php#L119-L129

mircobabini:
### All Submissions:

Resolves #1473.

### Other information:

### DIFF
--- a/includes/currencies.php
+++ b/includes/currencies.php
@@ -122,9 +122,13 @@
 	 */
 	function pmpro_euro_position_from_locale($position = 'right') {
 		$locale = get_locale();
+		
 		if(strpos($locale, 'en_') === 0) {
 			$position = 'left';
+		}elseif(strpos($locale, 'de_') === 0) {
+			$position = 'left';
 		}
+		
 		return $position;
 	}
 	


ideadude:
I noticed the current rules here (http://publications.europa.eu/code/en/en-370303.htm) state:

> Position of the euro sign (€) in amounts
> The euro sign is followed by the amount without space:
> a sum of €30
> NB: The same rule applies in Dutch, Irish and Maltese. In all other official EU languages the order is reversed; the amount is followed by a hard space and the euro sign:
> une somme de 30 €

However, Amazon.de has the euro symbol showing up after the amount. I'd really like there to be a consensus when I make changes like this, otherwise people will reach out and want the symbol on the OTHER SIDE again.

I don't know how we should define consensus. It would be nice if we could outsource these decisions and piggy back on the decisions of a larger open source project like WooCommerce or even a non-WP project that deals with currencies.

Any ideas?

mircobabini:
Nice job. I think we followed the same path (Quora > Wikipedia > European Union Commission). By the way, seems like the logic follows the old currency active before the euro introduction which is relatively new.

At this point, I totally agree. Was out of knowledge about this, just thought it was a bug but it's not. Woocommerce should let the user choose where to put the currency symbol as far as I know; but really don't know how the multi-currency feature works. I'll check it out soon.

I just found something very interesting.
1. Everyone is saying that the $ sign must be before the value.
1. Just found a guy telling that in canada, you put the $ sign after the value.
1. Looking deeper I found [this FastSpring article](https://fastspring.com/blog/how-to-format-30-currencies-from-countries-all-over-the-world/) and I think this one is pretty much like a bible... just because this is what they do.

> Canada does it both ways. In English Canada, the dollar symbol goes to the left of the amount, while in French Canada, the dollar sign goes to the right.

By the way, code talking now. This pull request is just to follow the same code-style of what was already implemented into PMPro, but since there is no real solid logic in place, I think we should switch to something more solid first.

Maybe we can just work on `global $pmpro_currencies` (just thinking out loud). Let's take the Canada example above. Can we add that logic in that structure? Maybe we should use the locales instead of the currency ISO 4217 code.

I'll make some other research. Speak soon.

mircobabini:
WooCommerce settings: https://i.imgur.com/eZJSuaLl.png
=> can just set a global value for all currencies

Multi Currency for WooCommerce (the best, they say): https://i.imgur.com/ztFt8oS.png
=> foreach currency, same settings as above

Another multi-currency: https://i.imgur.com/CgnlYwz.png
=> same as above

Can't find a very complete solution. I think that we should:
1) Define standards for most diffused currencies
2) Define overrides for well known specific countries differences ($ for Canada, EUR for Dutch, Irish and Maltese)

These filterable overrides should be locale-based.

sc0ttkclark:
@ideadude Did we sort out a solution for this?

dparker1005:
I am going to close this PR for now as there has not been discussion here in 5+ years. For the time being, the best way to customize how a currency is displayed on your site is by using the `pmpro_currencies` and `pmpro_euro_position` filters.

PR #1509, opened 2020-11-10
Extend order code possibilities to base-36 (0-9A-Z)

mircobabini:
### All Submissions:

Extend order code possibilities to base-36 (0-9A-Z), to add even more entropy and less collisions.

### Other information:

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -837,9 +837,15 @@
 			$count++;
 
 			while( empty( $code ) ) {
+				// create a unique scramble which uses a base-36 representation
 				$scramble = md5( AUTH_KEY . microtime() . SECURE_AUTH_KEY . $count );
+				$scramble = base_convert( $scramble, 16, 36 );
+
+				// create the code stripping out the extra chars
 				$code = substr( $scramble, 0, 10 );
 				$code = apply_filters( 'pmpro_random_code', $code, $this );	//filter
+
+				// ensure the uniqueness
 				$check = $wpdb->get_var( "SELECT id FROM $wpdb->pmpro_membership_orders WHERE code = '$code' LIMIT 1" );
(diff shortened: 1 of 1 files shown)


ideadude:
Do you get extra letters in practice? Maybe this comment is out of date, but it seems that sometimes large numbers (> 32 bits) wouldn't really convert: https://www.php.net/manual/en/function.base-convert.php#109660

I think 16^10 gives us over a trillion possible codes already. It also avoids letters like L and O which make the numbers less human-readable.

I appreciate the contributions coming in form large scale sites like the one you run. If there really are still collisions after the last update, we should do *something* to combat that.

mircobabini:
With the last update we have fixed the issue of two orders trying to generate a scramble at the same `time()` breaking the site, this is confirmed. ✅

Collisions, meaning some scramble in the future which will result in the same order code, is certainly rare but can always exist in theory. By the way there's the "while exists" loop for this scope, to avoid issues. ✅

I just found something very interesting.
1. Everyone is saying that the $ sign must be before the value.
1. Just found a guy telling that in canada, you put the $ sign after the value.
1. Looking deeper I found [this FastSpring article](https://fastspring.com/blog/how-to-format-30-currencies-from-countries-all-over-the-world/) and I think this one is pretty much like a bible... just because this is what they do.

> Canada does it both ways. In English Canada, the dollar symbol goes to the left of the amount, while in French Canada, the dollar sign goes to the right.

By the way, code talking now. This pull request is just to follow the same code-style of what was already implemented into PMPro, but since there is no real solid logic in place, I think we should switch to something more solid first.

Maybe we can just work on `global $pmpro_currencies` (just thinking out loud). Let's take the Canada example above. Can we add that logic in that structure? Maybe we should use the locales instead of the currency ISO 4217 code.

I'll make some other research. Speak soon.

PR #1510, opened 2020-11-11
Fixing 2Checkout Discount Codes User Count and Custom Expiration Date (enddate) Settings

karambk:
### All Submissions:

Resolves The following problems regarding discount codes and 2checkout :
1- Increase user count when discount code is used
2- Reads the new enddate after the discount code is applied

1. You need to have a Live fully functional 2checkout account integrated with pmpro
2. Create a Discount Code, edit the expiry date to a custom value different from the original membership
3. Test the Discount Code and you should see the increase of the user count that used this coupon and the custom expiry date should be applied

### Other information:

### Changelog entry

### DIFF
--- a/classes/gateways/class.pmprogateway_twocheckout.php
+++ b/classes/gateways/class.pmprogateway_twocheckout.php
@@ -217,7 +217,7 @@
 		 */
 		static function pmpro_checkout_before_change_membership_level($user_id, $morder)
 		{
-			global $wpdb, $discount_code_id;
+			global $wpdb;
 
 			//if no order, no need to pay
 			if(empty($morder))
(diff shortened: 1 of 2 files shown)


ideadude:
Thank you so much for this PR. Merging it in now. Will do some testing and include in the next update.

karambk:
> Thank you so much for this PR. Merging it in now. Will do some testing and include in the next update.

Great thank you so much, We've been testing this fix for quite sometime now, I can confirm it is perfectly working on our side.

karambk:
@ideadude Hello
I recommend a review for possible inconsistencies with 2checkout integration and discount codes

As some days ago we had a single strange case, and we couldn't replicate it

The case was a student was supposed to register for a full year after using a discount code

But instead the user registered with the original membership date which is a month and payed the discount code price.. 

Just wanted to inform you with this.. Although in theory there doesn't seem a way for a problem to happen after this fix, but it seems yet, the discount code end date might not be pulled correctly for some reason...

karambk:
@ideadude Hello again,

The problem happened once more, although there have been many many students registering without problems, but this problem happen once in a while..

To explain more what happens, here's a mysql table comparing healthy ones with non-health student
You can clearly see that there is no Start Data, nor Initial fees.. that also appear in the members list in pmpro itself..
mysql-error

Kind Regards,
Karam

karambk:
@ideadude 
So we identified the problem, all the members imported with CSV who doesn't contain startdate, has this problem when they EXTEND the membership... the scary thing is that even if I fixed the startdate for an individual, he's broken, more investigation is being done... 2checkout's removal of sandbox is really bad

karambk:
@ideadude 
Alright problem identified and reported here :
https://www.paidmembershipspro.com/forums/topic/mysql-start-date/

It has nothing to do with mysql

Simply, a person cannot extend his membership with a discount code and capture the correct end date

PR #1513, opened 2020-11-17
Conditionally load billing fields filter for Stripe Gateway

mircobabini:
### All Submissions:

I must admit I spent a while to find this one; I'm using PayPal Express only and have Stripe disabled. But Stripe has few lines of codes still running around, these are just few. These lines are disabling required billing fields on checkout even when just PayPal Express is selected. Well, PayPal Express disables these fields as well, but when you try to remove PayPal Express hooks you'll see that there are also that function hooked to `pmpro_required_billing_fields`.

PayPal Express gateway is implemented with this if condition as well.

### Other information:

I made tests with PayPal Express only. Tests should be run with Stripe as well.


ideadude:
Nice catch. I believe the code in the Stripe class' pmpro_required_billing_fields method was also causing issues when using a gateway like Authorize.net. The CVV field was not showing up as required, leading to odd error messages.

I tested this with our add PayPal Express add on, to make sure it doesn't get confused. We sometimes want to check the pmpro_gateway option in addition to using the pmpro_getGateway function, which checks the param/session/etc to see which gateway was chosen at checkout. In this case, it's fine as you coded it.

I'm going to merge this and push out in a point release right away.

PR #1538, opened 2020-12-26
Add missing pmpro_doing_webhook action

mircobabini:
### All Submissions:

This is very useful to me. I'm doing some debug to enhance the behaviour of the orders in status error from PayPal Express. At this time, I'm using the pmpro_ipn_validate to detect we are running an ipnrequest, which is the only action running every time an ipn requests comes, even if /services/ipnhandler.php is called directly (no wp ajax).

Then, I think this action can be useful to avoid hacks on hacks.

### Other information:

### DIFF
--- a/services/ipnhandler.php
+++ b/services/ipnhandler.php
@@ -23,6 +23,8 @@ if ( ! pmpro_ipnValidate() ) {
 	pmpro_ipnExit();
 }
 
+do_action( 'pmpro_doing_webhook' );
+
 //assign posted variables to local variables
 $txn_type               = pmpro_getParam( "txn_type", "POST" );
 $subscr_id              = pmpro_getParam( "subscr_id", "POST" );


mircobabini:
For instance this is how [I extend ipn to manage refunds](https://gist.github.com/mircobabini/c81da00ceab5b6bad0473c1d353dc609/4c0f47f2dffc0a3f4a773cde214c81c54b11a63d). And this is how it should be with that action: [changes](https://gist.github.com/mircobabini/c81da00ceab5b6bad0473c1d353dc609/revisions?diff=split).

ideadude:
I agree with the need for something like this. Let's have a chat sometime to figure out a solution that works for your use case and is done in a way general across all of the PMPro gateway webhooks.

In general, it would be nice to standardize the webhooks. We don't have to go that far. But we can e.g. make a function that we'd call at the top of every webhook that both defines the PMPRO_DOING_WEBHOOK constant and also runs a hook.

ideadude:
Did it this way: https://github.com/strangerstudios/paid-memberships-pro/commit/2fc9358e5c79e502c4c3add30032cfdb99b48b15

One change vs yours is that the hook is firing before the pmpro_ipnValidate() function is called/checked. We could update pmpro_ipnValidate() to cache the results. Then you could call pmpro_ipnValidate() in your callbacks without making extra calls to PayPal.

PR #1567, opened 2021-02-04
Display Country on it's own line

kimwhite:
Added a <be> tag to the Billing Address.
Added to invoice and confirmation pages.

The country should have its own line.


kimwhite:
COUNTRY... not County

kimwhite:
Perhaps we should add classes to these?  So the user can modify as needed?
Zip Code, Country?

PR #1571, opened 2021-02-12
Add filter for subscriptions created w/o subscribe

ermGit:
The "subscribe" function has a filter (pmpro_stripe_create_subscription_array) for the parameters that it passes when it creates the subscription.  However, the standard Checkout page does not use that route through the code.  The Checkout page uses a route that goes through the "create_subscription" function.  This pull request adds a filter (pmpro_stripe_create_subscription_params_array) to the "Stripe_Subscription::create($params) function call inside the "create_subscription" function.  Thereby allowing more flexibility in how the subscription is setup.

### All Submissions:

* [X] Have you followed the [Contributing guideline](https://github.com/strangerstudios/paid-memberships-pro/blob/master/.github/CONTRIBUTING.md)?

The "subscribe" function has a filter (pmpro_stripe_create_subscription_array) for the parameters that it passes when it creates the subscription.  However, the standard Checkout page does not use that route through the code.  The Checkout page uses a route that goes through the "create_subscription" function.  This change adds a filter (pmpro_stripe_create_subscription_params_array) to the "Stripe_Subscription::create($params) function call inside the "create_subscription" function.  Thereby allowing more flexibility in how the subscription is setup.

Resolves XXX.

1.  Buy a Subscription with the Checkout page
2. Now write a function that hooks into the new "pmpro_stripe_create_subscription_params_array" filter
3. Buy another Subscription with the Checkout page

### Other information:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?

### Changelog entry
The "subscribe" function has a filter (pmpro_stripe_create_subscription_array) for the parameters that it passes when it creates the subscription.  However, the standard Checkout page does not use that route through the code.  The Checkout page uses a route that goes through the "create_subscription" function.  This change adds a filter (pmpro_stripe_create_subscription_params_array) to the "Stripe_Subscription::create($params) function call inside the "create_subscription" function.  Thereby allowing more flexibility in how the subscription is setup.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -2978,7 +2978,7 @@ class PMProGateway_stripe extends PMProGateway {
 					'pending_setup_intent.payment_method',
 				),
 			);
-			$order->subscription = Stripe_Subscription::create( $params );
+			$order->subscription = Stripe_Subscription::create( apply_filters( 'pmpro_stripe_create_subscription_params_array', $params ) );
 		} catch ( Stripe\Error\Base $e ) {
 			$order->error = $e->getMessage();
 			return false;


ideadude:
Hi, thanks for this PR. I agree that we need a filter when we create the subscription on line 2981.

I'm about to push out a point release and may not be able to squeeze this in.

We have the pmpro_stripe_create_subscription_array filter on line 2157 and should maybe reuse that name. Or better, we need to standardize the filter names here and when/how we create the subscriptions. So I'd like to take a look to see if that is possible without too much refactoring before adding this filter.

We should be able to get to it before the next release.

Keep you site patched and stay tuned in case we rename filters when we merge this in. Thanks.

mircobabini:
@ideadude this way? https://github.com/strangerstudios/paid-memberships-pro/compare/dev...mircobabini:patch-11

ideadude:
I merged Mirco's version which includes the filter. Thanks for your work on this ermGit. This code will be in the next release going out soon.

PR #1576, opened 2021-02-18
Allow translation of Stripe's hosted fields.

andrewlimaza:
ENHANCEMENT: Automatically set the locale of the Stripe checkout fields based on the user's browser locale. Improves translations.


andrewlimaza:
Stripe doc reference - https://stripe.com/docs/js/appendix/supported_locales

PR #1591, opened 2021-03-06
Always fire pmpro_update_order/pmpro_updated_order

mircobabini:
### All Submissions:

I was debugging on the order status flow in case of issues with PayPal Express. After a while, I was able to recall that the status can change even when the `MemberOrder::saveOrder()` method is not called, because of `updateStatus`. This PR adds the update/updated actions every time runs an update query on `$wpdb->pmpro_membership_orders`.

⚠️ Also noticed there is a useless updateStatus [here](https://github.com/strangerstudios/paid-memberships-pro/blob/d0a252dfa967dca080f95cedacb1982410ff6847/classes/class.memberorder.php#L940) because it's already done by the early query above, to avoid race conditions.
⚠️ Last but not least, [this note](https://github.com/strangerstudios/paid-memberships-pro/blob/d0a252dfa967dca080f95cedacb1982410ff6847/classes/class.memberorder.php#L957) should be updated: it says the truth, but it does not consider the early query above, which was probably added later.

By the way, after that early query, we should just try to update the Gateway. No more updateStatus, so I would delete this one: https://github.com/strangerstudios/paid-memberships-pro/blob/d0a252dfa967dca080f95cedacb1982410ff6847/classes/class.memberorder.php#L940

No related issue to resolve, but very useful to debug.

### Other information:

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -616,8 +616,11 @@
 			global $wpdb;
 			$this->sqlQuery = "UPDATE $wpdb->pmpro_membership_orders SET timestamp = '" . $date . "' WHERE id = '" . $this->id . "' LIMIT 1";
 
+			do_action('pmpro_update_order', $this);
 			if($wpdb->query($this->sqlQuery) !== "false") {
 				$this->timestamp = strtotime( $date );
+				do_action('pmpro_updated_order', $this);
+				
 				return $this->getMemberOrderByID($this->id);
 			} else {
 				return false;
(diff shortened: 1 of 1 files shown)


classes/class.memberorder.php
```diff
@@ -863,12 +866,17 @@ function updateStatus($newstatus)
 			if(empty($this->id))
 				return false;
 
-			$this->status = $newstatus;
 			$this->sqlQuery = "UPDATE $wpdb->pmpro_membership_orders SET status = '" . esc_sql($newstatus) . "' WHERE id = '" . $this->id . "' LIMIT 1";
-			if($wpdb->query($this->sqlQuery) !== false)
+			
+			do_action('pmpro_update_order', $this);
+			if($wpdb->query($this->sqlQuery) !== false){
+				$this->status = $newstatus;
```
mircobabini, line 873: To be consistent, `$this->status` should be updated with the new value only if the query was successful.

mircobabini:
Especially useful to activate real-time backups from BlogVault every time the order status is changed. Work in progress: https://gist.github.com/mircobabini/ea948d4da9f2c395463b813197fa8559/revisions#diff-5ef215fac6a87ad440be3691abb144cb3653ce1dd94ebd70b092ec6b7bcab9c3R712-R731

sc0ttkclark (approved):
I'm approving this with the expectation that we will revisit Level / Order CRUD operations and abstract them all into a proper ORM

PR #1598, opened 2021-03-10
PayPal Express: avoid double processing of the same review request

mircobabini:
### All Submissions:

When something goes wrong with PayPal Express, the subscribe() method returns an error, but the user will stay on that page. URL is something like /checkout/?level=1&review=X&token=EC-Y&PayerID=Z. If the user hits F5, the order (which status already changed to review or error) got set to review once again. So we just lose the progress information and since a level agreement is already set for that order, we are unable to restore it but manually. This PR is to avoid double processing the same review request.

1. Force an order to give error during subscribe()
2. Try to reload the error page
3. See the error (A successful Billing Agreement has already been created for this token) and check the order status (review).

### Other information:

### Changelog entry

Avoid PayPal double processing of the same review request.

### DIFF
--- a/classes/gateways/class.pmprogateway_paypalexpress.php
+++ b/classes/gateways/class.pmprogateway_paypalexpress.php
@@ -322,22 +322,28 @@
 
 				$morder = new MemberOrder();
 				$morder->getMemberOrderByPayPalToken(sanitize_text_field($_REQUEST['token']));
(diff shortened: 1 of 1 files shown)


ideadude:
Thanks for this and the other PRs.

I'm going to merge this. I might change the error message "... was already processed." Sounds like the order went through successfully, which may not be the case. We can check the status. If it's error or review, show something more appropriate.

PR #1599, opened 2021-03-10
Fix the error handling in case of ACK=FAILURE during subscribe for PayPal Express

mircobabini:
### All Submissions:

Since saveOrder() is never called, the status = error does nothing and the order has been left in review status waiting for an IPN response with Initial Payment = Failure. ~~We already know that something goes wrong and we don't want to leave orders in review status. So, let's updateStatus( 'error' ).~~ **Wrong!** the order must stay in review status. Errors with the API doesn't mean that the status should be set to error (which means "not paid"). Errors with the API can be anything.

Then, stop the execution of the review process which leads the user to the confirmation message; can't confirm something that is not working.

1. Force return error during subscribe() for PayPal Express
2. Reload the page, so you'll get the error message "A successful Billing Agreement has already been created for this token"
3. You'll see the unwanted confirmation message, while the order is still in status=review.

### Other information:

### Changelog entry

Fix the error handling in case of ACK=FAILURE during subscribe for PayPal Express

### DIFF
--- a/classes/gateways/class.pmprogateway_paypalexpress.php
+++ b/classes/gateways/class.pmprogateway_paypalexpress.php
@@ -684,7 +684,7 @@
 
 		function subscribe(&$order)
 		{
-			global $pmpro_currency;
+			global $pmpro_currency, $pmpro_review;
 
 			if(empty($order->code))
 				$order->code = $order->getRandomCode();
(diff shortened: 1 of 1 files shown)


mircobabini:
When PayPal API returns an error, we don't want to set the order's status to error.

It doesn't make any sense. There was an error on the communication, can't assume the subscription was wrong or anything. status=error, for us, means "subscription not created because of initial payment failure".

**Then this PR which transform `$this->status = "error"` to `$order->updateStatus( 'error' )` is essentially wrong.**

`$this->status = "error"` is wrong as well and must be removed everywhere. Searching the whole code base seems like `$this->status = "error"` was around in Stripe Gateway and Braintree Gateway as well, but commented at some point.

It's still around in:
- PayPal Standard: is anyone still using it?
- PayPal Express: gonna fix it
- Authorize.net: don't know it
- Payflow Pro: don't know it
- CyberSource: only during `::subscribe()`
- 2Checkout: only during `::cancel()`

I searched all around for if statement like `status == "error"` and similar (`===`, yoda conditions, singular quotes, ...) and I've found nothing. So:
1. Setting the status and not saving te order is weird; we must save the order or NOT set the status. Saving the order does not make any sense, so we don't want to set the status.
2. It can probably be removed safely from all the gateways.

I'm gonna removing it by PayPal Express right now to complete the tests.

PR #1600, opened 2021-03-10
Handle missing PROFILESTATUS and fix api error handling

### LINKED ISSUE #1590: PayPal Express should handle errors on subscription creation
**Spoiler: probably found a solution 🎉**
tldr; Found that there is a difference between a CreateRecurringPaymentProfile success and an insuccess. In the first case, there is [PROFILESTATUS] => ActiveProfile; when it fails instead, even if ACK=SUCCESS, PROFILESTATUS is completely missing. [Fix idea](https://github.com/strangerstudios/paid-memberships-pro/compare/dev...mircobabini:patch-14).

-----

**Describe the bug**
When the users gets back from the gateway we know if the initial payment was successful or not. This is true for some gateways, but it's not always true for PayPal Express. Sometimes PayPal Express returns not enough informations to say "yes, the order was really successful". PMPro handles this uncertainty with this flow:
1. order in status token is created before the gateway call
1. order status becomes review and suddenly success, when gateway returns to the site
1. order status becomes error when on the ipn call reporting "Initial Payment = Failed"

**Consequences**

Confusion: the customer receive a success email, confirming its membership. He logs in, but after a minute or two he is not able to continue on site; he is redirected to the subscription page once again... not clear at all.

Invoices add-on: legal invoices got generated when order status is set to success, and should be invalidated when the order status becomes error. This is not possible, because every legal document emitted is emitted; can't just be invalidated. In fact, it should not be emitted at all. As already discussed, this is probably acceptable in US... but it's causing issues in EU, expecially in Italy because of e-invoice which must be precise, fast and essential to be considered valid and to avoid fines.

Gifts add-on: gift coupon code are sent when the order status is set to success, but they should be revoken after 2 minutes. This is impossible and confusing.

**Ideas**
1. The review status should be kept for a while; we can wait for the coming soon ipn response, which will change the order status to success or error. Meanwhile, we can set up an option to decide if the customer should be able to use the membership area or if he should see an informative message of the idle moment. Please note that there are possibilities for the ipn response to never come. These are rare cases I was able to solve using a cron job; every night, I'm checking the orders in review status, older than 24 hours (the day before). If there are some, I call the getSubscriptionStatus gateway feature and I'm 100% able to see if the initial payment was a success or an error. Tricky, but at least I was able to remove the orders in review status, which is not very informative later on.
1. We can use the getSubscriptionStatus gateway feature to better understand what's going on.
1. We can investigate and see if there is something helpful to avoid issues, with no additional requests.  **This is what I've tried to do in the following analysis.**

**The flow**
### Status = token

Thanks to the *pmpro_added_order* action I was able to detect the exact moment the MemberOrder was created, inserted into the database in status=token. This happens during the checkout page submit (submit-checkout=1), then the user is redirected to PayPal Express.

### Status = review

This is happening when the user get back from PayPal Express. The current request is similar to the checkout page with some additional params: &review={PMPRO_ORDER_CODE}&token={PAYPAL_TOKEN_EC-XXXXXXXXXXXXX}&PayerID={PAYPAL_PAYER_IDENTIFIER}.

Thanks to getExpressCheckoutDetails() we know from PayPal API[*GetExpressCheckoutDetails*] if the checkout was ACK=SUCCESS (causing order status = review) or ACK=SUCCESSWITHWARNING (causing order status = error). [https://github.com/mircobabini/paid-memberships-pro/blob/60ab16975ad75cdb78340512a32390f1468dc1aa/classes/gateways/class.pmprogateway_paypalexpress.php#L328](https://github.com/mircobabini/paid-memberships-pro/blob/60ab16975ad75cdb78340512a32390f1468dc1aa/classes/gateways/class.pmprogateway_paypalexpress.php#L328)

### Status = success

During the same routine as above, if PayPal Express is set to skip confirmation, $order→confirm() is called; this one is responsible of →subscribe(), running the PayPal API[*CreateRecurringPaymentsProfile*] to set up the recurrent payment: ACK=SUCCESS (causing order status = success) or ACK=SUCCESSWITHWARNING (causing order status = error). [https://github.com/mircobabini/paid-memberships-pro/blob/60ab16975ad75cdb78340512a32390f1468dc1aa/classes/gateways/class.pmprogateway_paypalexpress.php#L383](https://github.com/mircobabini/paid-memberships-pro/blob/60ab16975ad75cdb78340512a32390f1468dc1aa/classes/gateways/class.pmprogateway_paypalexpress.php#L383)

🔜: this is where we should fix the payment_transaction_id: [https://github.com/mircobabini/paid-memberships-pro/blob/60ab16975ad75cdb78340512a32390f1468dc1aa/classes/gateways/class.pmprogateway_paypalexpress.php#L747-L748](https://github.com/mircobabini/paid-memberships-pro/blob/60ab16975ad75cdb78340512a32390f1468dc1aa/classes/gateways/class.pmprogateway_paypalexpress.php#L747-L748). Must be collected somehow. Weird thing is that PayPal docs says we should have a TRANSACTIONID but it's missing. Let's see if PayPal Checkout can do something more. Related issue: #1077.

💡 Found that there is a difference between a CreateRecurringPaymentProfile success and an insuccess. In the first case, there is [PROFILESTATUS] => ActiveProfile; when it fails instead, even if ACK=SUCCESS, PROFILESTATUS is completely missing. This can be used to avoid setting the order status=success here: [https://github.com/strangerstudios/paid-memberships-pro/blob/d0a252dfa967dca080f95cedacb1982410ff6847/classes/gateways/class.pmprogateway_paypalexpress.php#L745-L753](https://github.com/strangerstudios/paid-memberships-pro/blob/d0a252dfa967dca080f95cedacb1982410ff6847/classes/gateways/class.pmprogateway_paypalexpress.php#L745-L753)

[This is the proof of concept fix](https://github.com/strangerstudios/paid-memberships-pro/compare/dev...mircobabini:patch-14). Must be completed, of course subscribe() must then return false at least... and must check the flow.

### Status = error

After a while, the ipn response comes from PayPal with txn_type = recurring_payment_profile_cancel, initial_payment_status = Failed. This was already subject of a recent (2020) fix about setting order status to error, instead of cancel, to avoid misunderstanding and not considering errored orders in analytics totals.

---

Personal note: this was debugged on order_id = 91424 (first data).
Then the missing PROFILESTATUS field is confirmed in dozens of orders on March 6, 2021.

mircobabini:
### All Submissions:

This one is the last of a series. There are three commits included:

1. Handle missing PROFILESTATUS, save the orders status=error and stop the review process.
2. Don't set status=error when just an API issue is detected.
3. Stop the review process when an API issue is detected during Gateway->subscribe().

Resolves #1590.

PayPal must send you back to the site as it was a success, but with no PROFILESTATUS when doing ->subscribe(). I've still found no way to replicate the issue, but I'm testing it on a live website on which occurs 20-30 times in a day. It's working.

### Other information:

### DIFF
--- a/classes/gateways/class.pmprogateway_paypalexpress.php
+++ b/classes/gateways/class.pmprogateway_paypalexpress.php
@@ -618,7 +618,6 @@
 
 				return true;
 			} else  {
-				//$order->status = "error";
 				$order->errorcode = $this->httpParsedResponseAr['L_ERRORCODE0'];
 				$order->error = urldecode($this->httpParsedResponseAr['L_LONGMESSAGE0']);
 				$order->shorterror = urldecode($this->httpParsedResponseAr['L_SHORTMESSAGE0']);
--- a/classes/gateways/class.pmprogateway_paypalexpress.php
+++ b/classes/gateways/class.pmprogateway_paypalexpress.php
@@ -778,6 +778,9 @@
 					return false;
 				}
 			} else  {
+				// stop processing the review request on checkout page
+				$pmpro_review = false;
+
 				//$order->status = "error";
 				$order->errorcode = $this->httpParsedResponseAr['L_ERRORCODE0'];
 				$order->error = urldecode($this->httpParsedResponseAr['L_LONGMESSAGE0']);
--- a/classes/gateways/class.pmprogateway_paypalexpress.php
+++ b/classes/gateways/class.pmprogateway_paypalexpress.php
@@ -585,7 +585,7 @@
 
 				//exit('SetExpressCheckout Completed Successfully: '.print_r($this->httpParsedResponseAr, true));
 			} else  {
-				$order->status = "error";
+				//$order->status = "error";
 				$order->errorcode = $this->httpParsedResponseAr['L_ERRORCODE0'];
 				$order->error = urldecode($this->httpParsedResponseAr['L_LONGMESSAGE0']);
 				$order->shorterror = urldecode($this->httpParsedResponseAr['L_SHORTMESSAGE0']);
@@ -619,7 +619,7 @@
 
 				return true;
 			} else  {
-				$order->status = "error";
+				//$order->status = "error";
 				$order->errorcode = $this->httpParsedResponseAr['L_ERRORCODE0'];
 				$order->error = urldecode($this->httpParsedResponseAr['L_LONGMESSAGE0']);
 				$order->shorterror = urldecode($this->httpParsedResponseAr['L_SHORTMESSAGE0']);
@@ -673,7 +673,7 @@
 
 				return true;
 			} else  {
-				$order->status = "error";
+				//$order->status = "error";
 				$order->errorcode = $this->httpParsedResponseAr['L_ERRORCODE0'];
 				$order->error = urldecode($this->httpParsedResponseAr['L_LONGMESSAGE0']);
 				$order->shorterror = urldecode($this->httpParsedResponseAr['L_SHORTMESSAGE0']);
@@ -778,7 +778,7 @@
 					return false;
 				}
 			} else  {
-				$order->status = "error";
+				//$order->status = "error";
 				$order->errorcode = $this->httpParsedResponseAr['L_ERRORCODE0'];
 				$order->error = urldecode($this->httpParsedResponseAr['L_LONGMESSAGE0']);
 				$order->shorterror = urldecode($this->httpParsedResponseAr['L_SHORTMESSAGE0']);
@@ -816,7 +816,7 @@
 			if("SUCCESS" == strtoupper($this->httpParsedResponseAr["ACK"]) || "SUCCESSWITHWARNING" == strtoupper($this->httpParsedResponseAr["ACK"])) {
 				return true;
 			} else {
-				$order->status = "error";
+				//$order->status = "error";
 				$order->errorcode = $this->httpParsedResponseAr['L_ERRORCODE0'];
 				$order->error = urldecode($this->httpParsedResponseAr['L_LONGMESSAGE0']) . ". " . __("Please contact the site owner or cancel your subscription from within PayPal to make sure you are not charged going forward.", 'paid-memberships-pro' );
 				$order->shorterror = urldecode($this->httpParsedResponseAr['L_SHORTMESSAGE0']);
@@ -844,7 +844,7 @@
 			}
 			else
 			{
-				$order->status = "error";
+				//$order->status = "error";
 				$order->errorcode = $this->httpParsedResponseAr['L_ERRORCODE0'];
 				$order->error = urldecode($this->httpParsedResponseAr['L_LONGMESSAGE0']);
 				$order->shorterror = urldecode($this->httpParsedResponseAr['L_SHORTMESSAGE0']);
--- a/classes/gateways/class.pmprogateway_paypalexpress.php
+++ b/classes/gateways/class.pmprogateway_paypalexpress.php
@@ -779,7 +779,6 @@
 				// stop processing the review request on checkout page
 				$pmpro_review = false;
 
-				//$order->status = "error";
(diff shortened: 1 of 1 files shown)


mircobabini:
Just updated the _Changes proposed in this Pull Request_ paragraph.

This one is ready to be merged.

mircobabini:
@ideadude just did a triple check with a few other PayPal Express implementations found around and still working like a charm on my production website after a month.

classes/gateways/class.pmprogateway_paypalexpress.php
```diff
@@ -585,7 +585,7 @@ function setExpressCheckout(&$order)
 
 				//exit('SetExpressCheckout Completed Successfully: '.print_r($this->httpParsedResponseAr, true));
 			} else  {
-				$order->status = "error";
+				//$order->status = "error";
```
ideadude, line 588:
```suggestion
				//$order->status = "error";
```
```suggestion
```

ideadude: Let's just remove this instead of commenting them out.

classes/gateways/class.pmprogateway_paypalexpress.php
```diff
@@ -619,7 +619,7 @@ function getExpressCheckoutDetails(&$order)
 
 				return true;
 			} else  {
-				$order->status = "error";
+				//$order->status = "error";
```
ideadude, line 622:
```suggestion
				//$order->status = "error";
```
```suggestion
```

classes/gateways/class.pmprogateway_paypalexpress.php
```diff
@@ -673,7 +673,7 @@ function charge(&$order)
...
```
ideadude, line 676:
```suggestion
				//$order->status = "error";
```
```suggestion
```

classes/gateways/class.pmprogateway_paypalexpress.php
```diff
@@ -743,16 +743,45 @@ function subscribe(&$order)
...
+					$order->saveOrder();
+
+					return false;
+				}
 			} else  {
-				$order->status = "error";
+				// stop processing the review request on checkout page
+				$pmpro_review = false;
+
+				//$order->status = "error";
```
ideadude, line 784:
```suggestion
				//$order->status = "error";
```
```suggestion
```

classes/gateways/class.pmprogateway_paypalexpress.php
```diff
@@ -790,7 +819,7 @@ function cancelSubscriptionAtGateway(&$order) {
 			if("SUCCESS" == strtoupper($this->httpParsedResponseAr["ACK"]) || "SUCCESSWITHWARNING" == strtoupper($this->httpParsedResponseAr["ACK"])) {
 				return true;
 			} else {
-				$order->status = "error";
+				//$order->status = "error";
```
ideadude, line 822:
```suggestion
				//$order->status = "error";
```
```suggestion
```

classes/gateways/class.pmprogateway_paypalexpress.php
```diff
@@ -818,7 +847,7 @@ function getSubscriptionStatus(&$order)
 			}
 			else
 			{
-				$order->status = "error";
+				//$order->status = "error";
```
ideadude, line 850:
```suggestion
				//$order->status = "error";
```
```suggestion
```

ideadude (commented):
Let's remove these `$order->status = 'error';` lines  instead of commenting them out.

classes/gateways/class.pmprogateway_paypalexpress.php
```diff
@@ -684,7 +684,7 @@ function charge(&$order)
 
 		function subscribe(&$order)
 		{
-			global $pmpro_currency;
+			global $pmpro_currency, $pmpro_review;
 
 			if(empty($order->code))
 				$order->code = $order->getRandomCode();
```
sc0ttkclark, line 690:
```suggestion
```

PR #1614, opened 2021-04-06
Bug Fix: Remove undefined variable warning at checkout

andrewlimaza:
BUG FIX: Fixes an undefined variable issue at checkout.


andrewlimaza:
Ignore this, I pulled from upstream dev but change wasn't there already?

PR #1618, opened 2021-04-09
Initial steps toward a return function for invoice/order prices parts

kimcoleman:
### All Submissions:

This PR adds a function to return a complex price with all the separate "parts" along with a new filter `pmpro_display_price_parts` to insert additional price data, such as the tax information that our forthcoming AvaTax Integration will add. This could also be used to show price data like a Donation amount, variable price-adjusting field through Register Helper, and more.

1. Create an order that has a subtotal, tax, and total.
2. Check the Membership Confirmation page, Invoice page, Order Print and Email (from admin).

### Other information:

### Changelog entry
* ENHANCEMENT: Added function to return a complete price block with breakdown of price parts for an order/invoice total.
* ENHANCEMENT: Added filter `pmpro_display_price_parts` to allow custom code and other plugins to insert parts into a displayed price total. 

### DIFF
--- a/adminpages/templates/orders-email.php
+++ b/adminpages/templates/orders-email.php
@@ -20,7 +20,7 @@
 	</tr>
 	<?php if(!empty($order->billing->name)): ?>
 		<tr>
-			<td>
+			<td style="padding-bottom:10px;">
 				<strong><?php _e( 'Bill to:', 'paid-memberships-pro' ); ?></strong><br>
 				<?php
 					echo pmpro_formatAddress(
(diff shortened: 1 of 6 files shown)


kimcoleman:
Adding files soon for an additional filter (now a filter for with and without the total so you can sneak things in between the two). Also, the price parts is now a filterable array of labels and values, making it easier to format in custom code like on the orders-email where it was expecting to draw table.

`function my_pmpro_get_price_parts( $pmpro_display_price_parts, $format ) {
	$pmpro_display_price_parts['GST'] = array (
		'label' => 'GST',
		'value' => pmpro_escape_price( pmpro_formatPrice( 22 ) ),
	);
	return $pmpro_display_price_parts;
}
add_filter( 'pmpro_get_price_parts', 'my_pmpro_get_price_parts', 10, 2 );
`

ideadude:
This is good. We were checking $order->total != '0.00' before, but we probably need to update that to account for different currency formats. Maybe we need a helper function or method on the order class to tell if a price is "free".

When a lot of subtotals get displayed, say 3 different taxes and maybe a donation, subtotal, etc, it would help to have different formatting for the different "levels" or "style" of price.

We can add a "style" or "class" element to the array here: https://github.com/strangerstudios/paid-memberships-pro/pull/1618/files#diff-1c08b27bcf43583c75cbc442bc401638be3427586fe544f3389e22c9979fa1afR2777

And then have a few different styles for items in the list. Maybe "total", "subtotal", and "default" or something like that.

ideadude:
To get different styles for each row in the price list, we could either add a class element to the array as suggested above, or we could add css classes to the wrapping <tr> with the array key for that price part. So something like:

<tr class="pmpro-price-part-subtotal" ...></tr>

kimcoleman:
PR updated to leverage the pmpro_element_class and the $key of the price part to wrap with a filterable and unique CSS selector per price part:

Screen Shot 2021-04-20 at 11 00 03 AM

PR #1642, opened 2021-05-03
BUG FIX: Fixed warning when all_levels were empty

andrewlimaza:
BUG FIX: Fixed issue where the default_level was calling min() on an empty variable. Improve the logic.

### All Submissions:

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2285,9 +2285,13 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
 		if ( empty( $level_id ) ) {
 			$all_levels = pmpro_getAllLevels( false, false );
 
-			// Get lowest level ID.
-			$default_level =  min( array_keys( $all_levels ) );
-
+			if ( ! empty( $all_levels ) ) {
+				// Get lowest level ID.
+				$default_level =  min( array_keys( $all_levels ) );
+			} else {
+				$default_level = null;
+			}
+			
 			$level_id = apply_filters( 'pmpro_default_level', intval( $default_level ) );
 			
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -2285,9 +2285,13 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
 		if ( empty( $level_id ) ) {
 			$all_levels = pmpro_getAllLevels( false, false );
 
-			// Get lowest level ID.
-			$default_level =  min( array_keys( $all_levels ) );
-
+			if ( ! empty( $all_levels ) ) {
+				// Get lowest level ID.
+				$default_level =  min( array_keys( $all_levels ) );
```
sc0ttkclark, line 2290: Should the intval be called anymore below, should it be called here?

includes/functions.php
```diff
@@ -2285,9 +2285,13 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
...
+			} else {
+				$default_level = null;
```
sc0ttkclark, line 2292: Should this just be 0?

sc0ttkclark (commented):
Just some feedback

includes/functions.php
ideadude: Nice catch. I'll move the intval up.

ideadude: Actually, I might leave it where it is. As far as I know, nothing relies on that default_level value being null, 0, or false specifically, just that it's empty. For the sake of making the smallest possible change here unless we have a reason to, I'm going to keep the intval where it is.

PR #1650, opened 2021-05-10
BUG FIX: Return null if user already has level.

andrewlimaza:
BUG FIX: Fixes an issue where level changed emails were sending out to admins when updating a user profile without changing their level.

### All Submissions:

### Other information:

### Changelog entry

BUG FIX: Fixes an issue where level changed emails were sending out to admins when updating a user profile without changing their level.


ideadude:
Thanks for this. We used to return false here when a user's level didn't change because they already had it. I think returning true, when it works, false, when there is an error, and returning null (nothing) when we don't do anything makes sense and would fix the issues that came up when we switched to returning true here.

PR #1651, opened 2021-05-11
BUG FIX: Adjusted deprecated function

andrewlimaza:
BUG FIX: Use wp.passwordStrength.userInputDisallowedList() instead of deprecated wp.passwordStrength.userInputBlacklist() function.

This has been deprecated since WordPress 5.5.0

### DIFF
--- a/js/pmpro-login.js
+++ b/js/pmpro-login.js
@@ -12,7 +12,13 @@ jQuery(document).ready(function(){
 		
 		var strength;		
 		if ( pass1 != '' ) {
-			strength = wp.passwordStrength.meter( pass1, wp.passwordStrength.userInputDisallowedList(), pass1 );
+			// Call the disallowed list method corresponding to appropriate WP version.
+			const disallowedList = ( 'function' == typeof wp.passwordStrength.userInputDisallowedList )
+				? wp.passwordStrength.userInputDisallowedList()
+				: wp.passwordStrength. userInputBlacklist();
+
+			strength = wp.passwordStrength.meter( pass1, disallowedList, pass1 );
+
 		} else {
 			strength = -1;
 		}
--- a/js/pmpro-login.js
+++ b/js/pmpro-login.js
@@ -12,7 +12,7 @@ jQuery(document).ready(function(){
 		
 		var strength;		
 		if ( pass1 != '' ) {
-			strength = wp.passwordStrength.meter( pass1, wp.passwordStrength.userInputBlacklist(), pass1 );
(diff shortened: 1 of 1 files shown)


ideadude:
I like this. But I believe userInputDisallowedList was added in WP 5.5, so would break backwards compatibility.

It doesn't make sense here to test and use different versions of the function.

We need a plan for which version of PMPro we will merge this into where we are ready to drop support for pre-WP 5.5.

js/pmpro-login.js
```diff
@@ -12,7 +12,7 @@ jQuery(document).ready(function(){
 		
 		var strength;		
 		if ( pass1 != '' ) {
-			strength = wp.passwordStrength.meter( pass1, wp.passwordStrength.userInputBlacklist(), pass1 );
+			strength = wp.passwordStrength.meter( pass1, wp.passwordStrength.userInputDisallowedList(), pass1 );
```
sc0ttkclark, line 15:
```suggestion
			// Call the disallowed list method corresponding to appropriate WP version.
			const disallowedList = ( 'function' == typeof wp.passwordStrength.userInputDisallowedList )
				? wp.passwordStrength.userInputDisallowedList()
				: wp.passwordStrength. userInputBlacklist();

			strength = wp.passwordStrength.meter( pass1, disallowedList, pass1 );

```

sc0ttkclark: @ideadude What about this?

ideadude: At first I thought something like this "doesn't make sense"... adding too much tech debt and kind of highlighting the terms we want to avoid... but it's growing on me as I look at it. It works. I like it.

js/pmpro-login.js
```diff
@@ -12,7 +12,13 @@ jQuery(document).ready(function(){
 		
 		var strength;		
 		if ( pass1 != '' ) {
-			strength = wp.passwordStrength.meter( pass1, wp.passwordStrength.userInputBlacklist(), pass1 );
+			// Call the disallowed list method corresponding to appropriate WP version.
+			const disallowedList = ( 'function' == typeof wp.passwordStrength.userInputDisallowedList )
+				? wp.passwordStrength.userInputDisallowedList()
+				: wp.passwordStrength. userInputBlacklist();
```
sc0ttkclark, line 18: @ideadude A space snuck in here :(

sc0ttkclark: Fixed https://github.com/strangerstudios/paid-memberships-pro/commit/a54ca035d8d1a644ab0863087ecb88fc6b11e700

PR #1654, opened 2021-05-12
Sets keys in the array returned from pmpro_sort_levels_by_order() to correspond to the level ids. Some code expects that for level arrays.

ideadude:
Before this fix, we were resetting the array keys in the array passed out of this function. And so on the level page, we would set the $pmpro_levels global var to use the ordered (and re-keyed) array. This caused issues later when the pmpro_getLevel() function expected the $pmpro_levels array to have keys corresponding to the level ids. $pmpro_levels[1] may or may not be level 1.

We thought about changing the $pmpro_getLevel function to not have that expectation, but that's a bigger change and it is useful to have the level arrays keyed this way. It's worth trying to maintain IMO.

For now, this one commit should fix the issues that have come up since the introduction of the pmpro_sort_levels_by_order() function.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2374,9 +2374,9 @@ function pmpro_sort_levels_by_order( $pmpro_levels ) {
 	foreach ( $sort_order as $level_id ) {
 		foreach ( $pmpro_levels as $key => $level ) {
 			if ( ! empty ( $level->id ) && $level_id == $level->id ) {
-				$reordered_levels[] = $pmpro_levels[$key];
+				$reordered_levels[$level_id] = $pmpro_levels[$key];
 			} elseif ( ! empty( $level ) && is_string( $level ) && $level_id == $level ) {
-				$reordered_levels[] = $pmpro_levels[$key];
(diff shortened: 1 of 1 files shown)


kimcoleman:
Tested this update and it fixes the issue on the levels page for what I was trying to do when I discovered it.

To test, without this code try using the following in a shortcode block on the Membership Levels page:
`[membership level="0"][pmpro_advanced_levels levels="1"][/membership]
[membership level="1"][pmpro_advanced_levels levels="2,3,4,5,6"][/membership]`

Without this updated code, a user with level 1 would not see the contents of the second membership shortcode.

kimcoleman (approved):
Tested this update and it fixes the issue on the levels page for what I was trying to do when I discovered it.

To test, without this code try using the following in a shortcode block on the Membership Levels page:
`[membership level="0"][pmpro_advanced_levels levels="1"][/membership]
[membership level="1"][pmpro_advanced_levels levels="2,3,4,5,6"][/membership]`

Without this updated code, a user with level 1 would not see the contents of the second membership shortcode.

PR #1683, opened 2021-06-29
Email templates merge

andrewlimaza:
Merging Email Templates into core.

There's a lot of changes in this PR, but here's the gist of it:

* Moved Email Templates to Memberships > Settings > Email Settings
* Added in recurring billing email template by default.
* Added email templates via PHP strings vs HTML files (helps with i18n and stability).
* Filters are kept intact for backwards compatibility.
* Can remove email templates (HTML files from core after this is merged in).
* Adjusted the logic in figuring out what email template to send (improved).


mircobabini:
I have a nice custom setup for email templates on my site:

1. Added some custom templates
1. Disabled some templates
1. Customized almost all of them and translated to it_IT
1. Using [WP HTML Mail for template design](https://it.wordpress.org/plugins/wp-html-mail/)

Are the changes included into this PR to be considered as braking changes?
Is there any chance to make it 100% backward compatible?

I'll make some tests on this, but I suppose we have enough time for it before the 2.6 release.

ideadude:
Hi, Mirco. This update is meant to be 100% backwards compatible with anyone who was using custom .html template files, the old filters, or the separate email templates add on. Let us know what if anything breaks with this version of things. Thanks.

mircobabini:
It works perfectly.

PR #1688, opened 2021-07-09
Fix for other Elementor restriction plugins

codezz:
This logic overrides other restriction plugins by setting the render to be always true even though there is no PMPRO restriction setting

### All Submissions:

It breaks other Elementor restriction logic plugins

1. Install Visibility Logic for Elementor 
2. Apply a restriction to a section
3. You will see it is always visible due to the logic in PMPRO not taking into account the filter variable.

### Other information:

### Changelog entry

> Make sure PMPRO restrict logic does not to override other plugins restriction logic.


ideadude:
Thank you for catching this and for the fix. I fixed it in a slightly different way in this commit here: https://github.com/strangerstudios/paid-memberships-pro/commit/95d8c321a968c106e2931da65cde83564258ca84

This is going out in the next point release and should be compatible with your fix.

PR #1732, opened 2021-08-18
Replace !empty() check with isset() for decimal option

### LINKED ISSUE #1731: Function pmpro_round_price does not set decimals correctly if currency decimals is zero
**Describe the bug**
When using `pmpro_round_price` the price is not rounded correctly when the specified currency has `0` (zero) decimals.

**To Reproduce**
1. Create a custom function that hooks into the `pmpro_checkout_level` filter
2. add `var_dump( pmpro_round_price( 575.34, 'JPY' ) );`
3. Test on PMPro checkout page
4. See error: returns `float 575.34` instead of the expected 575 as Japanese Yen has 0 (zero) decimals.

**Expected behavior**
Prices are correctly rounded according to specified decimals, as per global variable `$pmpro_currencies`, for the currency.
Instead of `empty` we could consider using `isset` to check if the variable contains a decimal value here:
https://github.com/strangerstudios/paid-memberships-pro/blob/8ef6b647db1992ba4d41c3913672b7bba021f8e4/includes/functions.php#L3077

sc0ttkclark:
### All Submissions:

Resolves #1731.

1. Have a currency set that uses 0 decimals
2. Attempt to round price
3. See the rounded price has decimals still set

### Other information:

### Changelog entry

> Fix problem where `pmpro_round_price()` would not take into account currencies with decimals set to 0.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -3073,7 +3073,7 @@ function pmpro_round_price( $price, $currency = '' ) {
 	}
 
 	if ( ! empty( $pmpro_currencies[ $currency ] )
-		&& is_array( $pmpro_currencies[ $pmpro_currency ] )
+		&& is_array( $pmpro_currencies[ $currency ] )
 		&& isset( $pmpro_currencies[ $currency ]['decimals'] ) ) {
 		$decimals = intval( $pmpro_currencies[ $currency ]['decimals'] );
 	}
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Noting: This was the only place in the file and the whole plugin where this logic was using `! empty()` instead of `isset()`

dparker1005 (approved):

includes/functions.php
```diff
@@ -3074,7 +3074,7 @@ function pmpro_round_price( $price, $currency = '' ) {
 
 	if ( ! empty( $pmpro_currencies[ $currency ] )
 		&& is_array( $pmpro_currencies[ $pmpro_currency ] )
```
sc0ttkclark, line 3076:
```suggestion
		&& is_array( $pmpro_currencies[ $currency ] )
```

PR #1736, opened 2021-08-19
Remove check for condition which always 'true' because of prev check

freax:
### All Submissions:

- [x] Have you followed the [Contributing guideline](https://github.com/strangerstudios/paid-memberships-pro/blob/master/.github/CONTRIBUTING.md)?
- [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/strangerstudios/paid-memberships-pro/pulls) for the same update/change?

Condition is always 'true' because 'empty( $addon['License'] ) || $addon['License'] == 'wordpress.org'' is already 'false' at this point (see **line 117**).

### Other information:

- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you successfully run tests with your changes locally?

### Changelog entry

Remove unused condition check.

### DIFF
--- a/includes/addons.php
+++ b/includes/addons.php
@@ -130,7 +130,7 @@ function pmpro_update_plugins_filter( $value ) {
 		}
 
 		// compare versions
-		if ( ! empty( $addon['License'] ) && version_compare( $plugin_data['Version'], $addon['Version'], '<' ) ) {
+		if ( version_compare( $plugin_data['Version'], $addon['Version'], '<' ) ) {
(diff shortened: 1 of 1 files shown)


ideadude:
Nice catch. Merging in.

Not sure if `$addon['Version']` could be unset here and throw a warning, but we will address that if it comes up. I believe if 'License' is set, then 'Version' is set too.

PR #1756, opened 2021-08-31
Keep filters on admin item delete (orders, discounts)

mircobabini:
### All Submissions:

When deleting an order, the current filters (order status, search, ...) are lost. This PR, after cleaning up the REQUEST_URI from the 'delete' key, permits the usage of add_query_arg to keep the filters.

A future enhancement would be to handle the request differently, the way the whole wordpress' core does.
After the 'delete' action is handled, we should redirect to 'deleted' => 1.

Reference: https://github.com/WordPress/WordPress/blob/master/wp-admin/edit.php#L457

1. Create two orders
2. Filter in some way
3. Delete an order

### Other information:

### DIFF
--- a/adminpages/discountcodes.php
+++ b/adminpages/discountcodes.php
@@ -363,6 +363,9 @@
 	}
 
 	require_once(dirname(__FILE__) . "/admin_header.php");
+
+    // Clean-up request uri from in-page actions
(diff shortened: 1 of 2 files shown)


sc0ttkclark (changes requested):
Let's not modify `$_SERVER['REQUEST_URI']` directly here, it should just set it to a variable and then reference that variable

dparker1005:
Closing as this has been fixed during the migration to the current Orders list table.

PR #1759, opened 2021-08-31
Order's deletion message similar to discount's deletion message

mircobabini:
Discount deletion message says which discount has just been deleted.

This was not the same for order's deletion. This is the fix.


sc0ttkclark (commented):
Looks good, just needs decision on whether we should use Order ID or the Order Code to display here.

mircobabini:
> Looks good, just needs decision on whether we should use Order ID or the Order Code to display here.

I think we must lean towards the most readable way, which is the order code.

PR #1775, opened 2021-09-13
Better style Order History & Membership Levels History when empty

mircobabini:
### All Submissions:

From:
Screenshot 2021-09-13 at 10 55 36

To:
Screenshot 2021-09-13 at 10 55 15

For _Order History_ and _Membership Level History_ too.

### Other information:

### Changelog entry

> ...: Better style Order History & Membership Levels History when empty

### DIFF
--- a/includes/profile.php
+++ b/includes/profile.php
@@ -481,9 +481,15 @@ function pmpro_membership_history_profile_fields( $user ) {
 			?>
 			</tbody>
 			</table>
-			<?php } else { 
-				esc_html_e( 'No membership orders found.', 'paid-memberships-pro' );
-			} ?>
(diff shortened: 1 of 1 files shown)


sc0ttkclark (changes requested):
Should this be a <p> tag instead of a table? It seems like this way may be the wrong way to go at solving the style problem. cc @kimcoleman

ideadude:
I guess the idea is to have this look the same as if it were a table. I like how this PR does that.

If we used different markup, we'd have to use CSS to make it look similar. I think it's okay to be a little un-semantic here to keep the output consistent. Will default to Kim's opinion.

PR #1781, opened 2021-09-16
Change status to receive new membership information

andrewlimaza:
ENHANCEMENT: Add more membership statuses to trigger on when polling for membership information. This is useful for Zapier Application to trigger on more than just active accounts. This will help with cancelled and expired level changes too.

### All Submissions:

### DIFF
--- a/includes/rest-api.php
+++ b/includes/rest-api.php
@@ -201,6 +201,9 @@ if ( class_exists( 'WP_REST_Controller' ) ) {
 				array(
 					'methods'  => WP_REST_Server::READABLE,
 					'callback' => array( $this, 'pmpro_rest_api_recent_memberships' ),
+					'args'	=> array(
+						'status' => array(),
+					),
 					'permission_callback' => array( $this, 'pmpro_rest_api_get_permissions_check' ),
 				)
 		));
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Lets add a new REST API parameter to support choosing which status to filter by (one comma separated list). We can default to showing these that you've set, but it would be good to allow to show only active.

sc0ttkclark (changes requested):
Let's make this filterable via REST API param, default can be these new statuses, but someone may want to just filter by active.

andrewlimaza:
I've made the changes to support the parameter 'level_status' which can take single or comma-separated values.

I also added in the filter 'pmpro_rest_recent_member_default_level_status' should you not be able to add in a parameter during the request (PMPro Zapier application doesn't support params for the trigger that runs on this route). I will look at adding this in the future to PMPro Zapier application when triggering data on this event.

Let me know your thoughts. Will move it to draft too for the time being.

andrewlimaza:
I've made the decision to keep backwards compatibility. This route is used solely by the Paid Memberships Pro - Zapier application.

This defaults to standard functionality of 'active' membership statuses only, unless the level_status parameter is supplied. An update will be coming to the Zapier Application to support input and support this parameter in a future release. This will keep current zaps running on Zapier 1.0.0 unaltered as there's quite a lot of data being run through PMPro Zapier Application.

I've left it as a draft for now, but I'm happy with this. Let me know your thoughts.

sc0ttkclark (approved):

PR #1790, opened 2021-09-24
Cleaned up Stripe payment option fields

dparker1005:
### All Submissions:

Simplified code in `pmpro_payment_option_fields()` Stripe method and refactored duplicate code RE Connect to it's own function. 

### Other information:

### Changelog entry


dparker1005:
This is the commit that this PR is conflicting with:
https://github.com/strangerstudios/paid-memberships-pro/commit/4472a172fae416e033e597e7687cf68c5aa8f642#diff-531a755f394c6d86b8be7ba048311f27917d349b6d95072eb8f6a27ef93d69e8

If we want to use the inline JS solution in that commit to show the legacy fields, then we need to undo some of the work in this PR. Otherwise, we can implement similar functionality in this PR to replace the functionality in that commit, such as setting a URL parameter that causes `show_legacy_keys_settings()` to return `true`.

dparker1005:
Opted to keep Jason's change and undo the work I did there. So now this PR is just refactoring the duplicated live/sandbox credentials into a single function that is called twice.

PR #1793, opened 2021-09-27
Add missing Payment Type for gateways that did not set them

sc0ttkclark:
### All Submissions:

Resolves the problem where Payment Type is empty for many of our gateways while it's set for 2Checkout and others.

The gateways missing Payment Type were:

* Authorize.net
* Braintree Payments
* CyberSource
* Payflow Pro/PayPal Pro
* PayPal Website Payments Pro
* Stripe

1. Make a payment on one of the affected gateways
2. Look at the Payment Type set on the order
3. It should be set as the correct payment type matching that gateway

### Other information:

### Changelog entry

> Fixed Payment Type value for new orders so they get set to the gateway as expected for all gateways instead of just a few of them.

### DIFF
--- a/classes/gateways/class.pmprogateway_authorizenet.php
+++ b/classes/gateways/class.pmprogateway_authorizenet.php
@@ -43,7 +43,7 @@ class PMProGateway_authorizenet extends PMProGateway
 		$morder->CVV2 = $authorizenet_cvv;
 		return $morder;
 	}
-	
+
 	/**
 	 * Make sure this gateway is in the gateways list
 	 *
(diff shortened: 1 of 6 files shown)


ideadude:
Missed 2.6.3. I/we need to figure out what the 'payment type' field is actually capturing and is this the data we want to put into that column. We already capture the gateway somewhere else.

I think this was set up for the Website Payments Pro gateway where the payment type could be credit card or paypal. We should consider using it for Stripe to note if Google or Apple pay is used/etc (although as implemented now, we're basically just getting a credit card handoff from those services).

PR #1799, opened 2021-10-01
Broke Stripe getCustomer() into separate methods.

dparker1005:
### All Submissions:

Broke the getCustomer() method in the Stripe gateway class into separate functions and deprecated the original getCustomer() method.

Also moved to temporarily using `setCustomer()` where `getCustomer()` was being used, though that method will be deprecated as well later in the cycle.

Now also sending customer billing address to Stripe and added filters to modify customer data sent to Stripe.

### Changelog entry

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1457,7 +1457,7 @@ class PMProGateway_stripe extends PMProGateway {
 		$amount          = pmpro_round_price( (float) $order->subtotal + (float) $tax );
 
 		//create a customer
-		$result = $this->getCustomer( $order );
+		$result = $this->setCustomer( $order );
 
 		if ( empty( $result ) ) {
 			//failed to create customer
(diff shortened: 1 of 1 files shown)


classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -1517,126 +1517,185 @@ function charge( &$order ) {
...
+	 *
+	 * Should only be called when checkout is being proceesed. Otherwise,
+	 * use update_customer_from_user() method.
+	 *
+	 * @since TBD
+	 *
+	 * @param MemberOrder $order to create/update Stripe customer for.
+	 * @return Stripe_Customer|false
+	 */
+	private function update_customer_from_order( $order ) {
```
ideadude, line 1686: Let's call this update_customer_at_checkout since that's how it's used.
```suggestion
	private function update_customer_at_checkout( $order ) {
```

classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -1661,88 +1721,160 @@ function getCustomer( &$order = false, $force = false ) {
...
+				return $customer;
+			}
+		}
 
-				return $this->customer;
-			} catch ( \Throwable $e ) {
-				// Assume no customer found.
-			} catch ( \Exception $e ) {
-				// Assume no customer found.
+		// The current user does not have an existing customer in Stripe. Create one.
```
ideadude, line 1785:
```suggestion
		// No customer yet. Need to create one.
```

classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -1661,88 +1721,160 @@ function getCustomer( &$order = false, $force = false ) {
...
+		$customer = $this->create_customer( $customer_args );
+		if ( is_string( $customer ) ) {
+			// We were not able to create a new user in Stripe.
+			$order->error      = __( "Error creating customer record with Stripe.", 'paid-memberships-pro' ) . " " . $customer;
+			$order->shorterror = $order->error;
+			return false;
+		}
+
+		if ( empty( $user_id ) ) {
```
ideadude, line 1794:
```suggestion
		// If we don't have a user yet, we need to update their user meta after registration.
		if ( empty( $user_id ) ) {
```

classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -1661,88 +1721,160 @@ function getCustomer( &$order = false, $force = false ) {
...
+			// User not registered yet, queue it up.
```
ideadude, line 1795:
```suggestion
```

classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -1661,88 +1721,160 @@ function getCustomer( &$order = false, $force = false ) {
...
-			} else {
-				//user not registered yet, queue it up
-				global $pmpro_stripe_customer_id;
-				$pmpro_stripe_customer_id = $this->customer->id;
-				if ( ! function_exists( 'pmpro_user_register_stripe_customerid' ) ) {
-					function pmpro_user_register_stripe_customerid( $user_id ) {
-						global $pmpro_stripe_customer_id;
-						update_user_meta( $user_id, "pmpro_stripe_customerid", $pmpro_stripe_customer_id );
-					}
+		// Get data to update customer with.
```
ideadude, line 1843: Let's check if the customer at Stripe already has a billing address, and if so let's not push the billing address we have in user meta to Stripe.

We should bring the `$customer = $this->get_customer_for_user( $user_id );` line from below above this so we can check the values on the customer.

ideadude: Alternatively, let's continue to consider fully synchronizing the customer billing address. If we use a webhook to update the billing address user meta when the address is updated at Stripe, then we're less likely to overwrite something we don't want to. We would need to handle loops.

dparker1005: Applied this suggestion in commit aba4a58, fully synchronizing the customer billing address seems like something that we should keep in mind and talk about across the entirety of PMPro

ideadude:
Merging now. One concern is that the checks for existing addresses check if all address fields are not empty. This might be a problem for addresses with no state or custom checkout fields that are only gathering state and zipcode, etc. In those cases, we might want to update the address even though we only have some address data.

Going to hold off on that for now though.

PR #1801, opened 2021-10-01
Adding deprecation notice to public static webhook methods in Stripe …

dparker1005:
### All Submissions:

Create a new method method_being_called_from_outside_class() in Stripe class, and use it to deprecate some public static methods that should be made private non-static.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -831,9 +831,9 @@ class PMProGateway_stripe extends PMProGateway {
 	 * @deprecated TBD. Only deprecated for public use, will be changed to private nonstatic in a future version.
 	 */
 	static function get_webhooks( $limit = 10 ) {
-		if ( method_being_called_from_outside_class() ) {
-			_deprecated_function( __FUNCTION__, 'TBD' );
-		}
+		// Show deprecation warning if called publically.
+		pmpro_method_should_be_private( 'TBD' );
+
 		if ( ! class_exists( 'Stripe\WebhookEndpoint' ) ) {
 			// Load Stripe library.
 			new PMProGateway_stripe();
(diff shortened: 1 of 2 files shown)


classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -3396,4 +3422,20 @@ private function convert_price_to_unit_amount( $price ) {
...
+	 * Determine whether the calling method was called from
+	 * inside or outside of the Stripe Gateway class.
+	 *
+	 * Useful for preparing to change method visibility from
+	 * public to private.
+	 *
+	 * @return bool
+	 */
+	static private function method_being_called_from_outside_class() {
+		$backtrace = debug_backtrace();
```
sc0ttkclark, line 3435:
```suggestion
		$backtrace = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS );
```

sc0ttkclark: If we don't need arguments we don't need to get them, this just saves a little bit of memory

classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -3396,4 +3422,20 @@ private function convert_price_to_unit_amount( $price ) {
...
+	/**
+	 * Determine whether the calling method was called from
+	 * inside or outside of the Stripe Gateway class.
+	 *
+	 * Useful for preparing to change method visibility from
+	 * public to private.
+	 *
+	 * @return bool
+	 */
+	static private function method_being_called_from_outside_class() {
```
sc0ttkclark, line 3434: In general this is the normal order used everywhere in PHP land and it's best to keep things familiar for devs

```suggestion
	private static function method_being_called_from_outside_class() {
```

sc0ttkclark (approved):

sc0ttkclark (approved):

ideadude:
Maybe let's make this a general function in includes/functions.php so it's available to other classes.

Also include the conditional and deprecation warning inside the new function.

Then we can just add one line to the top of any method we want to check like this:

`pmpro_method_should_be_private()'

And that would do the check and throw the warning.

dparker1005:
Changing to draft until this is reworked

PR #1830, opened 2021-10-27
Added function pmpro_send_200_http_response()

dparker1005:
### All Submissions:

Added function pmpro_send_200_http_response(). Useful to avoid issues like timeouts from gateways during webhook/IPN handlers. Works with Apache and Nginx.

The following code will disable this functionality:
`add_filter( 'pmpro_send_200_http_response', '__return_false' );`

### Other information:

### Changelog entry

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -3791,7 +3791,7 @@ add_filter( 'wp_kses_allowed_html', 'pmpro_kses_allowed_html', 10, 2 );
  *
  * Works with Apache and Nginx.
  *
- * Copied from https://stackoverflow.com/a/42245266
+ * Based on code from https://stackoverflow.com/a/42245266
  *
  * @since TBD.
  */
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -3782,3 +3782,45 @@ function pmpro_kses_allowed_html( $allowed_html, $context ) {
 	return $allowed_html;
 }
 add_filter( 'wp_kses_allowed_html', 'pmpro_kses_allowed_html', 10, 2 );
+
+/**
+ * Send a 200 HTTP reponse without ending PHP execution.
+ *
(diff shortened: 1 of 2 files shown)


recastrodiaz:
I'm a bit late to this PR and I'm sorry if this is obvious to you, but would it make more sense to make this new behaviour opt-in instead of opt-out? I think there's a real chance of events "being lost" by returning 200 OK to Stripe before actually processing the event. For instance, if the database is down or has load issues the event won't be processed and no changes in memberships will take place, yet Stripe won't attempt to redeliver the webhook.

sc0ttkclark:
@recastrodiaz We implemented this for Stripe as an opt-out solution due to the Stripe recommendation that webhooks respond immediately before more complicated logic.

https://stripe.com/docs/webhooks#acknowledge-events-immediately

You can disable this by using this filter if you'd like: `add_filter( 'pmpro_send_200_http_response', '__return_false' );`

We'll look into other potential improvements to our handling here as we continue to expand Stripe. It's possible we can address the potential DB/load issues with additional logic for a better failsafe event handling across all of our gateways in the future.

recastrodiaz:
Thanks for the quick reply @sc0ttkclark. I can see why you made this opt-out.

IMO It's a weird recommendation from Stripe as they do retry timeouts even though the docs say otherwise (currently after 1h).

> Built-in retries

> Stripe webhooks have built-in retry methods for 3xx, 4xx, or 5xx response status codes. If Stripe doesn’t quickly receive a 2xx response status code for an event, we mark the event as failed and stop trying to send it to your endpoint. After multiple days, we email you about the misconfigured endpoint, and automatically disable it soon after if you haven’t addressed it.

A sample retry from Stripe after a timeout webhook event:

Capture d’écran 2021-11-03 à 18 38 57

Thanks again for the quick reply.

PR #1856, opened 2021-11-12
Paid Memberships Pro 2.6.5

sc0ttkclark:
Release PR


.gitignore
```diff
@@ -22,6 +22,7 @@ package-lock.json
 # composer
 vendor
 composer.js
+composer.json
```
sc0ttkclark, line 25: I don't think this should be here, maybe in .gitattributes for export ignore? cc @ideadude

sc0ttkclark: Moved this to .gitattributes

PR #1862, opened 2021-11-17
Cleaned up REST API For Zapier

andrewlimaza:
Cleaned up REST API for Zapier requirements.

Added in function `pmpro_format_date_iso8601` to return a ISO-8601 date format which is required by Zapier triggers.

### All Submissions:

1. Test this with v1.0.2 Zapier native PMPro integration. (Zapier 1.0.2 won't work until you merge this PR in as it changes data structure).

### Other information:

### Changelog entry

Improve the REST API to support Zapier native requirements.

### DIFF
--- a/includes/rest-api.php
+++ b/includes/rest-api.php
@@ -726,7 +726,6 @@ if ( class_exists( 'WP_REST_Controller' ) ) {
 				$members_limit = 1;
 			}
 
-			$response_type = isset( $params['response_type'] ) ? sanitize_text_field( $params['response_type'] ) : null;
 			$limit = apply_filters( 'pmpro_trigger_recent_members_limit', $members_limit );
 
 			// Grab the useful information.
@@ -734,7 +733,7 @@ if ( class_exists( 'WP_REST_Controller' ) ) {
 
 			$sql = "
 				SELECT
-					`mu`.`user_id`,
+					`mu`.`user_id` as `id`,
 					`u`.`user_email`,
 					`u`.`user_nicename`,
 					`mu`.`membership_id`,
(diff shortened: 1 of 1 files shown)


includes/rest-api.php
```diff
@@ -885,6 +877,16 @@ function pmpro_rest_api_convert_to_array( $string ) {
 			return explode( ',', $string );
 		}
 
+		/** 
+		 * Returns formatted ISO-8601 date (Used for Zapier Native app.)
+		 * @param $date date A valid date value.
+		 * @since TBD
+		 */
+		function pmpro_format_date_iso8601( $date ) {
```
sc0ttkclark, line 885:
```suggestion
		function public pmpro_format_date_iso8601( $date ) {
```

sc0ttkclark (commented):
Looks good, would be good to set visibility on all methods to ensure PHP never complains about that (and coding standards rules)

includes/rest-api.php
```diff
@@ -755,13 +754,10 @@ public function pmpro_rest_api_recent_memberships( $request ) {
 
 			$results = $wpdb->get_results( $wpdb->prepare( $sql, $limit ) );
```
eighty20results, line 755: Since it seems like you're expecting a single row to be returned (based on line 817, for instance), what about being more specific in the DB access code and use $wpdb->get_row( $wpdb->prepare( $sql, $limit ) ); ?

includes/rest-api.php
```diff
@@ -819,12 +814,9 @@ public function pmpro_rest_api_recent_orders( $request ) {
 			
 			$results = $wpdb->get_results( $wpdb->prepare( $sql, $limit ) );
```
eighty20results, line 815: Same as for line 755. Justification: You avoid the need to use the array index to address the correct object and the code becomes a little more readable.

includes/rest-api.php
sc0ttkclark: It could be more than one but initial support is based around the first one for Zapier purposes. The API could still be used to get multiple results by providing the 'limit' parameter.

includes/rest-api.php
sc0ttkclark: I think we'll revisit this area of the code in the future to utilize array_map on pmpro_format_date_iso8601 for addressing the column(s) needing to be formatted.

Issue #1617, opened 2021-04-08
Membership renewal considers month to always be 30 days.


dparker1005:
**Describe the bug**
When a user renews an existing membership, PMPro calculates their new expiration date based off of their previous expiration date:
https://github.com/strangerstudios/paid-memberships-pro/blob/4812f157e759a38b2d22cf56be728cc00592af07/includes/filters.php#L6-L53

This calculation assumes that a month is always 30 days, which is not accurate:
https://github.com/strangerstudios/paid-memberships-pro/blob/4812f157e759a38b2d22cf56be728cc00592af07/includes/filters.php#L40

The current approach is the quickest way to "estimate" the number of days until when the user's new expiration date should be, but the real solution would be to:
1. Get the user's current expiration date
2. Use date math (`strtotime()`) to calculate when their new expiration date should be
3. Use date math to get the number of days til that new expiration date
4. Return the number of days that was just calculated

**Expected behavior**
A user renewing a one month membership should have it end on exactly the same day of the following month.

If they were supposed to end on the 31st of this month and the next month only has 30 days, their new enddate when they renew should be the 30th.

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only Paid Memberships Pro plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Memberlite](https://wordpress.org/themes/memberlite/).
- [x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
<details>
```
Please share non-sensitive information about your hosting environment such as WordPress version, PHP version, Paid Memberships Pro and any related plugins versions.
```
</details>

PR #1864, opened 2021-11-17
Fix/membership renewal considers month to always be 30 days

fisher2470:
### All Submissions:

* [x ] Have you followed the [Contributing guideline](https://github.com/strangerstudios/paid-memberships-pro/blob/master/.github/CONTRIBUTING.md)?
* [ x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)?
* [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/strangerstudios/paid-memberships-pro/pulls) for the same update/change?

Resolves Membership renewal considers month to always be 30 days

Utilizes strtotime(+1 month) function, checks if days of expiration date is greater than the last day of the following month. If the last day of the following month is greater than the expiration date's day, then precedes as normal. Otherwise, the new expiration day is set to the last day of the next month, this date is than subtracted from todays date to produce the number of days in the extension

Attempt to extend a membership by one month

### Other information:

* [ x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ x] Have you successfully run tests with your changes locally?

### Changelog entry

#1617

### DIFF
--- a/includes/filters.php
+++ b/includes/filters.php
@@ -40,9 +40,9 @@ function pmpro_checkout_level_extend_memberships( $level ) {
 				$expiration_day = $expiration_date->format('d');
 				$expiration_month = strtotime('+1 month', $expiration_date->format('n'));
 				if ($expiration_day > cal_days_in_month($expiration_month)) {
-					$additional_month = strtotime('Y-m-t', strtotime("+1 month", $expiration_date));
+					$additional_month = strtotime('Y-m-t', strtotime("+1 month", $expiration_date)) - $todays_date;
 				} else {
-					$additional_month = strtotime('Y-m-d', strtotime("+1 month", $expiration_date));
+					$additional_month = (strtotime('Y-m-d', strtotime("+1 month", $expiration_date))) - $todays_date;
 				}
 				$total_days = $days_left + $level->expiration_number * $additional_month;
 			} elseif ( $level->expiration_period == 'Year' ) {
--- a/includes/filters.php
+++ b/includes/filters.php
@@ -38,13 +38,13 @@ function pmpro_checkout_level_extend_memberships( $level ) {
 				$total_days = $days_left + $level->expiration_number * 7;
 			} elseif ( $level->expiration_period == 'Month' ) {
 				$expiration_day = $expiration_date->format('d');
-				$expiration_month = strtotime('+1 month', $expiration_date->format('n'));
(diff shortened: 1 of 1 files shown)


ideadude:
Hi there. Thanks for looking into this. We can do a better job of accounting for months with different numbers of days. This PR seems to fix things for +1 month, which is the most common, however, we could probably update it further to account for cases where a membership is e.g. "every 3 months".

So where you use strtotime('+1 month'), you could use the $level->expiration_number value instead of 1. Then just add the days to the expiration.

Does that make sense? When that's done, we can test further and include in a release.

fisher2470:
That definitely makes sense, edited the filter to utilize $level->expiration_number instead of 1. 

I noticed the filter pmpro_ipnhandler_level_extend_memberships also uses the "* 30) method for extending memberships, should I add the change to that filter in this pull request as well?

includes/filters.php
```diff
@@ -37,7 +37,14 @@ function pmpro_checkout_level_extend_memberships( $level ) {
 			} elseif ( $level->expiration_period == 'Week' ) {
 				$total_days = $days_left + $level->expiration_number * 7;
 			} elseif ( $level->expiration_period == 'Month' ) {
-				$total_days = $days_left + $level->expiration_number * 30;
+				$expiration_day = $expiration_date->format('d');
+				$expiration_month = strtotime("+{$level->expiration_number} month", $expiration_date->format('n'));
+				if ($expiration_day > cal_days_in_month($expiration_month)) {
+					$additional_time = strtotime('Y-m-t', strtotime("+{$level->expiration_number} month", $expiration_date)) - $todays_date;
+				} else {
+					$additional_time = (strtotime('Y-m-d', strtotime("+{$level->expiration_number}", $expiration_date))) - $todays_date;
```
sc0ttkclark, line 45:
```suggestion
					$additional_time = (strtotime('Y-m-d', strtotime("+{$level->expiration_number} month", $expiration_date))) - $todays_date;
```

sc0ttkclark, line 45: The usage of `Y-m-t` and `Y-m-d` here with the first parameter of strtotime won't work as needed here, perhaps this was meant to be working with `date_i18n()` or `date()` and then back into a `strtotime()`?

`"+{$level->expiration_number}"` should be `"+{$level->expiration_number} month"`

dparker1005:
Closing in favor of #2945 which fixes this same issue and simplifies the rest of the function's logic.

PR #1871, opened 2021-12-01
POC to combine tooltip for renewals + revenue

JarrydLong:
Attempt at combining the tooltip label into one so that you can hover over revenue or renewals and it will show as follows: 

image

### DIFF
--- a/adminpages/reports/sales.php
+++ b/adminpages/reports/sales.php
@@ -304,6 +304,16 @@ function pmpro_report_sales_page()
 	if ( 0 !== $units_in_period ) {
 		$average = $total_in_period / $units_in_period; // Not including this unit.
 	}
+
+	$chart_headers = array(
+		array( "label" => esc_html( $date_function ), "type" => "string" ),
+		array( "label" => __( "Renewals", "paid-memberships-pro" ), "type" => "number" ),
+		array( "role" => "tooltip", "type" => "string", "p" => array( "html" => true ), "strong" => array( "html" => true ) ),
+		array( "label" => esc_html( sprintf( __( "New %s", "paid-memberships-pro" ), ucwords( $type ) ) ), "type" => "number" ),
+		array( "role" => "tooltip", "type" => "string", "p" => array( "html" => true ), "strong" => array( "html" => true ) ),
+		array( "label" => __( "Average*", "paid-memberships-pro" ), "type" => "number" )
+	);
+	
 	?>
 	<form id="posts-filter" method="get" action="">
 	<h1>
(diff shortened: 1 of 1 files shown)


adminpages/reports/sales.php
```diff
@@ -404,17 +404,44 @@ function drawVisualization() {
 
 			var data = google.visualization.arrayToDataTable([
 				[
-					'<?php echo esc_html( $date_function );?>', '<?php _e( 'Renewals', 'paid-memberships-pro' );?>', '<?php echo esc_html( sprintf( __( 'New %s', 'paid-memberships-pro' ), ucwords( $type ) ) );?>', '<?php _e( 'Average*', 'paid-memberships-pro' );?>', 
+					{ 'label': '<?php echo esc_html( $date_function );?>', 'type': 'string' },
```
sc0ttkclark, line 407: The correct way to output dynamic data to JS and prevent single quote / JS breakage is to create the data in PHP and then echo the value as JSON.

The code here looks like it would present a JS break if someone translates text with single quotes in them.

sc0ttkclark: Something like `<?php echo wp_json_encode( $data ); ?>`

adminpages/reports/sales.php
```diff
@@ -403,18 +413,38 @@ function pmpro_ShowMonthOrYear()
 		function drawVisualization() {
 
 			var data = google.visualization.arrayToDataTable([
-				[
-					'<?php echo esc_html( $date_function );?>', '<?php _e( 'Renewals', 'paid-memberships-pro' );?>', '<?php echo esc_html( sprintf( __( 'New %s', 'paid-memberships-pro' ), ucwords( $type ) ) );?>', '<?php _e( 'Average*', 'paid-memberships-pro' );?>', 
-				],
+				<?php echo wp_json_encode( $chart_headers ); ?>,
```
sc0ttkclark, line 416: I love this, it would be fantastic to do the same thing for the data output from $cols foreach below this. Essentially building the array of all data to pass into `arrayToDataTable( ..json encoded data.. )`.

sc0ttkclark (changes requested):
There's still a lot of string echoing here and we ought to deal with it by putting all of the JS var building direct into PHP and then echo the json encoded data into JS.

adminpages/reports/sales.php
JarrydLong: I've pushed some changes to move the PHP out of the JS

adminpages/reports/sales.php
```diff
@@ -304,6 +304,58 @@ function pmpro_report_sales_page()
...
+		}
+
+		$chart_contents[$date][] = esc_html( pmpro_round_price( $value[1] ) );
+		$chart_contents[$date][] = $tooltip_string;
+		$chart_contents[$date][] = esc_html( pmpro_round_price( $value[0] - $value[1] ) );
+		$chart_contents[$date][] = $tooltip_string;
+		$chart_contents[$date][] = esc_html( pmpro_round_price( $average ) );
+
+	}
+	// var_dump($chart_contents);
```
sc0ttkclark, line 357:
```suggestion

	$chart_data = array_merge( [ $chart_headers ], array_values( $chart_contents ) );
```

adminpages/reports/sales.php
```diff
@@ -402,21 +454,10 @@ function pmpro_ShowMonthOrYear()
...
-						} else {
-							echo esc_html( $date );
-						}
-					?>', <?php echo esc_html( pmpro_round_price( $value[1] ) );?>, <?php echo esc_html( pmpro_round_price( $value[0] - $value[1] ) );?>, <?php echo esc_html( pmpro_round_price( $average ) );?>,  ], 
-				<?php } ?>
-			]);
+			var data = google.visualization.arrayToDataTable([ 
+				<?php echo wp_json_encode( $chart_headers ); ?>,
+				<?php foreach( $chart_contents as $cc ) { echo wp_json_encode( $cc ).","; } ?> 
+			] );
```
sc0ttkclark, line 460:
```suggestion
			var data = google.visualization.arrayToDataTable(
				<?php echo wp_json_encode( $chart_data ); ?>
			);
```

JarrydLong: Thanks

kimcoleman:
Closing and going to move forward with this one instead: https://github.com/strangerstudios/paid-memberships-pro/pull/1911

PR #1891, opened 2022-01-04
Some MySQL setups require a primary key for each table. Going to doub…

ideadude:
A user was having bugs caused by these 2 tables not being created on their MySQL setup because the tables did not have primary keys.

So this PR sets the first unique key to a primary key. This table does not use "null" for any of the values in the columns, so this is okay.

I am double checking if we need the second unique key. I believe it's there for performance when querying by category, membership vs membership,category.

### DIFF
--- a/includes/upgradecheck.php
+++ b/includes/upgradecheck.php
@@ -268,6 +268,15 @@ function pmpro_checkForUpgrades()
 		$pmpro_db_version = pmpro_upgrade_2_6();
 		pmpro_setOption( 'db_version', '2.6' );
 	}
+	
+	/**
+	 * Version 2.6.7
+	 * Running pmpro_db_delta to update KEY types in a couple tables.
+	 */
+	 if( $pmpro_db_version < 2.67 ) {
+ 		pmpro_db_delta(); 		
+ 		pmpro_setOption( 'db_version', '2.67' );
+ 	}
 }
 
 function pmpro_db_delta()
@@ -373,7 +382,7 @@ function pmpro_db_delta()
 		  `membership_id` int(11) unsigned NOT NULL,
 		  `category_id` bigint(20) unsigned NOT NULL,
 		  `modified` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
-		  UNIQUE KEY `membership_category` (`membership_id`,`category_id`),
+		  PRIMARY KEY `membership_category` (`membership_id`,`category_id`),
 		  UNIQUE KEY `category_membership` (`category_id`,`membership_id`)
 		);
 	";
@@ -385,7 +394,7 @@ function pmpro_db_delta()
 		  `membership_id` int(11) unsigned NOT NULL,
 		  `page_id` bigint(20) unsigned NOT NULL,
 		  `modified` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
-		  UNIQUE KEY `category_membership` (`page_id`,`membership_id`),
+		  PRIMARY KEY `category_membership` (`page_id`,`membership_id`),
 		  UNIQUE KEY `membership_page` (`membership_id`,`page_id`)
 		);
 	";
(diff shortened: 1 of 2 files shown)


ideadude:
Note, this PR also creates a DB update for 2.6.7, which is the expected release this would be included in.

mircobabini:
> I am double checking if we need the second unique key. I believe it's there for performance when querying by category, membership vs membership,category.

When creating table indexes with 2+ columns you are in fact considering to make SELECT queries for category, membership vs membership,category, but when possible it's better to have just 1 index.

1) Define the index order (ex. category,membership)
2) Make all the select query respect this order when adding WHERE conditions

Having indexes with 2+ columns affects queries with LESS where conditions, but they must be in the correct order.
In other words the index "category, membership" affects the SELECT query where only checking for a specific category.

I usually implement this set up:
- 1 or more 1-column indexes (for general purposes on the table)
- 1 more specific 2+ columns index when performing specific queries

`page_id` is involved in just 4 queries and it's always the first where condition and since PRIMARY key is also UNIQUE by definition, it should be safe to get rid of:
`UNIQUE KEY `membership_page` (`membership_id`,`page_id`)`

Btw the membership_id columns is used to inner join the table, so it must be a KEY.
I would probably go for:

```sql
CREATE TABLE `wp_pmpro_memberships_pages` (
  `membership_id` int(11) unsigned NOT NULL,
  `page_id` int(11) unsigned NOT NULL,
  `modified` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  PRIMARY KEY `category_membership` (`page_id`,`membership_id`),
  KEY `membership_id` (`membership_id`)
);
```

On large databases this should affect (<<) the table size keeping the same performances. But since the logic should be reversed (first you think the indexes, then you write the queries) it's probably useful to make consideration on the other approach (inverting membership_id with page_id).

Useful reading:
https://medium.com/@marcel_domke/mysql-how-to-use-indices-c356b8ebe065

includes/upgradecheck.php
```diff
@@ -268,6 +268,15 @@ function pmpro_checkForUpgrades()
 		$pmpro_db_version = pmpro_upgrade_2_6();
 		pmpro_setOption( 'db_version', '2.6' );
 	}
+	
+	/**
+	 * Version 2.6.7
```
ideadude, line 273:
```suggestion
	 * Version 2.6.8
```

includes/upgradecheck.php
```diff
@@ -268,6 +268,15 @@ function pmpro_checkForUpgrades()
...
+	 * Running pmpro_db_delta to update KEY types in a couple tables.
+	 */
+	 if( $pmpro_db_version < 2.67 ) {
```
ideadude, line 276:
```suggestion
	 if( $pmpro_db_version < 2.68 ) {
```

includes/upgradecheck.php
```diff
@@ -268,6 +268,15 @@ function pmpro_checkForUpgrades()
...
+ 		pmpro_db_delta(); 		
+ 		pmpro_setOption( 'db_version', '2.67' );
```
ideadude, line 278:
```suggestion
 		pmpro_setOption( 'db_version', '2.68' );
```

ideadude (commented):
This will be in version 2.6.8 now, so updating db version numbers.

ideadude:
Nice point RE our own queries always using page_id first. So the membership_id, page_id compound index might not be needed. It does double as an index for just membership_id.

In practice, I don't think these indexes get too large and so it probably isn't a big deal.

It's my understanding that INDEX or KEY here are basically synonyms, although we should double check either works with WP's db delta. It expects certain formats. (https://developer.wordpress.org/reference/functions/dbdelta/#comment-1329)

I also think the UNIQUE part there is redundant because the PRIMARY key is going to keep those columns from have dupe combinations. It shouldn't have an effect in practice. I wonder if it's more confusing to include the UNIQUE that's not needed or exclude it, like "hey this needs to be unique too".

We're splitting hairs a bit which is good. I might push as is and keep this open to debate. Something else I am going to look into test before release (perhaps soon) is whether swapping key/index definitions via db delta on larger tables could cause DB issues.

PR #1901, opened 2022-01-11
Enhancement/decline throttling

ideadude:
Spammers are testing credit cards on our checkout pages. Adding reCAPTCHA to the page helps. Using gateway tools like Stripe's RADAR helps. But still, this activity puts load on the site, incurs charges, and potentially can get your gateway to cancel your account.

This PR adds a new file includes/spam.php with some a new system to help combat that.

The checkout and billing preheaders will now make a call to pmpro_track_spam_activity() whenever they fail in error. If there are more than 10 failed charges, authorizations, subscriptions, or billing updates from the same IP address within a 15 minute period, then the pmpro_required_billing_fields hook is used to set up an error to stop checkouts/updates from happening. This will prevent the spammer from continued spam testing.

There is a hook pmpro_is_spammer passing params $is_spammer and $activity, which can be used to override whether or not the spam preventive measures will happen.

The $activity value is an array of timestamps of when the spam activity occurred. It does not log the type of activity. We could set that up now or in the future.

You can adjust the spam activity limit and time limit by adding these constants to your wp-config.php

`
define( 'PMPRO_SPAM_ACTION_NUM_LIMIT', 10 );    // tracked activity within the timeframe
define( 'PMPRO_SPAM_ACTION_TIME_LIMIT', 900 );  // in seconds
`

The activity is stored in transients tied to the IP address. We want to keep the data footprint small.

Some open questions:
* Do we want to specifically exclude admins from the spam checking?
* Do we want to make this feature a setting on the Payment Settings page or should we always enable it?
* Do we want to do something more drastic if pmpro_is_spammer()? We could filter the shortcodes to hide the forms all together, redirect those users to a specific page, or serve them an invalid HTTP status.
* Are there false positives, DDOS possibilities, or other issues with this system?
* Are there better default values for the activity and time limit? Should they be settings.
* Do we want the lockout period to be separate from the tracking period? Both are 15m now.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -3939,5 +3939,8 @@ function pmpro_get_ip() {
 		return false;
 	}
 	
+	// Sanitize the IP
+	$client_ip = preg_replace( '/[^0-9a-fA-F:., ]/', '', $client_ip );
+	
 	return $client_ip;
 }
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -3874,3 +3874,70 @@ function pmpro_format_date_iso8601( $date ) {
 	$datetime = new DateTime( $date );
 	return $datetime->format( DateTime::ATOM );
 }
+
+/**
+ * Determines the user's actual IP address
+ *
+ * $_SERVER['REMOTE_ADDR'] cannot be used in all cases, such as when the user
+ * is making their request through a proxy, or when the web server is behind
+ * a proxy. In those cases, $_SERVER['REMOTE_ADDR'] is set to the proxy address rather
+ * than the user's actual address.
+ *
+ * Modified from WP_Community_Events::get_unsafe_client_ip() in core WP.
+ * Modified from https://stackoverflow.com/a/2031935/450127, MIT license.
+ * Modified from https://github.com/geertw/php-ip-anonymizer, MIT license.
+ *
+ * SECURITY WARNING: This function is _NOT_ intended to be used in
+ * circumstances where the authenticity of the IP address matters. This does
+ * _NOT_ guarantee that the returned address is valid or accurate, and it can
+ * be easily spoofed.
+ *
+ * @since 2.7
+ *
+ * @return string|false The ip address on success or false on failure.
+ */
+function pmpro_get_ip() {
+	$client_ip = false;
+
+	// In order of preference, with the best ones for this purpose first.
+	// Added some from JetPack's Jetpack_Protect_Module::get_headers()
+	$address_headers = array(
+		'GD_PHP_HANDLER',
+		'HTTP_AKAMAI_ORIGIN_HOP',
+		'HTTP_CF_CONNECTING_IP',
+		'HTTP_CLIENT_IP',
+		'HTTP_FASTLY_CLIENT_IP',
+		'HTTP_FORWARDED',
+		'HTTP_FORWARDED_FOR',
+		'HTTP_INCAP_CLIENT_IP',
+		'HTTP_TRUE_CLIENT_IP',
+		'HTTP_X_CLIENTIP',
+		'HTTP_X_CLUSTER_CLIENT_IP',
+		'HTTP_X_FORWARDED',
+		'HTTP_X_FORWARDED_FOR',
+		'HTTP_X_IP_TRAIL',
+		'HTTP_X_REAL_IP',
+		'HTTP_X_VARNISH',
+		'REMOTE_ADDR',			
+	);
+
+	foreach ( $address_headers as $header ) {
+		if ( array_key_exists( $header, $_SERVER ) ) {
+			/*
+			 * HTTP_X_FORWARDED_FOR can contain a chain of comma-separated
+			 * addresses. The first one is the original client. It can't be
+			 * trusted for authenticity, but we don't need to for this purpose.
+			 */
+			$address_chain = explode( ',', $_SERVER[ $header ] );
+			$client_ip     = trim( $address_chain[0] );
+
+			break;
+		}
+	}
+
+	if ( ! $client_ip ) {
+		return false;
+	}
+	
+	return $client_ip;
+}
--- a/includes/spam.php
+++ b/includes/spam.php
@@ -121,7 +121,6 @@ function pmpro_track_spam_activity( $ip = null ) {
  *
  * @return bool True if the clearing of activity was successful, or false if IP could not be determined.
  */
- */
 function pmpro_clear_spam_activity( $ip = null ) {
     if ( empty( $ip ) ) {
         $ip = pmpro_get_ip();        
@@ -138,6 +137,18 @@ function pmpro_clear_spam_activity( $ip = null ) {
 
 	return true;
 }
+
+/**
+ * Track spam activity when checkouts or billing updates fail.
+ *
+ * @since 2.7
+ * @param MemberOrder $morder The order object used at checkout. We ignore it.
+ */
+function pmpro_track_failed_checkouts_for_spam( $morder ) {
+    pmpro_track_spam_activity();
+}
+add_action( 'pmpro_checkout_processing_failed', 'pmpro_track_failed_checkouts_for_spam' );
+add_action( 'pmpro_update_billing_failed', 'pmpro_track_failed_checkouts_for_spam' );
 
 /**
  * Disable checkout and billing update forms for spammers.
--- a/includes/spam.php
+++ b/includes/spam.php
@@ -22,7 +22,7 @@ function pmpro_is_spammer() {
         $is_spammer = true;
     }
     
-    return apply_filters( 'pmpro_is_spammer', $is_spammer );
+    return apply_filters( 'pmpro_is_spammer', $is_spammer, $activity );
 }
 
 /**
--- a/includes/spam.php
+++ b/includes/spam.php
@@ -150,6 +150,12 @@ function pmpro_clear_spam_activity( $ip = null ) {
  * @param MemberOrder $morder The order object used at checkout. We ignore it.
  */
 function pmpro_track_failed_checkouts_for_spam( $morder ) {
+	// Bail if Spam Protection is disabled.
+	$spamprotection = pmpro_getOption("spamprotection");	
+	if ( empty( $spamprotection ) ) {
+		return;
+	}
+	
 	pmpro_track_spam_activity();
 }
 add_action( 'pmpro_checkout_processing_failed', 'pmpro_track_failed_checkouts_for_spam' );
@@ -171,6 +177,12 @@ add_action( 'pmpro_update_billing_failed', 'pmpro_track_failed_checkouts_for_spa
  * @return array The list of required fields.
  */
 function pmpro_disable_checkout_for_spammers( $required_fields ) {
+	// Bail if Spam Protection is disabled.
+	$spamprotection = pmpro_getOption("spamprotection");	
+	if ( empty( $spamprotection ) ) {
+		return $required_fields;
+	}
+	
 	if ( pmpro_was_checkout_form_submitted() && pmpro_is_spammer() ) {
 		pmpro_setMessage( __( 'Suspicious activity detected. Try again in a few minutes.', 'paid-memberships-pro' ), 'pmpro_error' );
 	}
--- a/includes/spam.php
+++ b/includes/spam.php
@@ -46,7 +46,9 @@ function pmpro_get_spam_activity( $ip = null ) {
     
     $transient_key = 'pmpro_spam_activity_' . $ip;
     $activity = get_transient( $transient_key );    
-    if ( empty( $activity ) ) { $activity = []; }
+	if ( empty( $activity ) || ! is_array( $activity ) ) {
+		$activity = [];
+	}
     
     // Remove old items.
 	$new_activity = [];
--- a/includes/spam.php
+++ b/includes/spam.php
@@ -25,7 +25,15 @@ function pmpro_is_spammer() {
         $is_spammer = true;
     }
     
-    return apply_filters( 'pmpro_is_spammer', $is_spammer, $activity );
+	/**
+	 * Allow filtering whether the current visitor is a spammer.
+	 *
+	 * @since 2.7
+	 *
+	 * @param bool  $is_spammer Whether the current visitor is a spammer.
+	 * @param array $activity   The list of potential spam activity.
+	 */
+	return apply_filters( 'pmpro_is_spammer', $is_spammer, $activity );
 }
 
 /**
--- a/includes/spam.php
+++ b/includes/spam.php
@@ -95,7 +95,7 @@ function pmpro_track_spam_activity( $ip = null ) {
     
     // Save to transient.
     $transient_key = 'pmpro_spam_activity_' . $ip;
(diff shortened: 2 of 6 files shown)


preheaders/billing.php
```diff
@@ -276,7 +276,11 @@
 			
 			do_action( 'pmpro_after_update_billing', $current_user->ID, $morder );
         } else {
-            $pmpro_msg = $morder->error;
+			// Track as spam activity to prevent too many CC attempts.
```
sc0ttkclark, line 279: Can we add a hook here instead perhaps, and have `pmpro_track_spam_activity` trigger on that hook?

Thinking about a hook here because it may be useful to allow people (or us) to implement other solutions to this problem via add ons.

preheaders/checkout.php
```diff
@@ -450,11 +450,14 @@
 						$pmpro_msgt      = "pmpro_success";
 						$pmpro_confirmed = true;
 					} else {
+						// Track as spam activity to prevent too many CC attempts.
```
sc0ttkclark, line 453: Can we add a hook here instead perhaps, and have `pmpro_track_spam_activity` trigger on that hook?

Thinking about a hook here because it may be useful to allow people (or us) to implement other solutions to this problem via add ons.

includes/functions.php
```diff
@@ -3894,3 +3894,73 @@ function pmpro_format_date_iso8601( $date ) {
...
+		'HTTP_INCAP_CLIENT_IP',
+		'HTTP_TRUE_CLIENT_IP',
+		'HTTP_X_CLIENTIP',
+		'HTTP_X_CLUSTER_CLIENT_IP',
+		'HTTP_X_FORWARDED',
+		'HTTP_X_FORWARDED_FOR',
+		'HTTP_X_IP_TRAIL',
+		'HTTP_X_REAL_IP',
+		'HTTP_X_VARNISH',
+		'REMOTE_ADDR',
```
sc0ttkclark, line 3941: Any further thoughts on REMOTE_ADDR based on your twitter thread? Should that be put first?

includes/spam.php
```diff
@@ -0,0 +1,139 @@
+<?php
+/**
+ * Code related to spam detection and prevention.
+ */
+// Constants. Define these in wp-config.php to override.
```
sc0ttkclark, line 5: Do we for sure want to introduce new constants for this or create a function with the value filtered so we can offer snippets?

Asking since snippets seem better organized for customizations and constants end up piling up together in the same spot of wp-config.php

I'm still good with the constant usage here, just asking the Q in case it makes sense one way or the other.

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+if ( ! defined( 'PMPRO_SPAM_ACTION_NUM_LIMIT' ) ) {
+    define( 'PMPRO_SPAM_ACTION_NUM_LIMIT', 10 );
+}
+if ( ! defined( 'PMPRO_SPAM_ACTION_TIME_LIMIT' ) ) {
+    define( 'PMPRO_SPAM_ACTION_TIME_LIMIT', 900 );  // in seconds
+}
+
+/**
+ * Is the current visitor a spammer?
+ * @since 2.7
```
sc0ttkclark, line 15:
```suggestion
 * Determine whether the current visitor a spammer.
 *
 * @since 2.7
 *
 * @return bool Whether the current visitor a spammer.

```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
+<?php
```
sc0ttkclark, line 1: This file is using spaces instead of tabs, was that intentional?

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+ */
+function pmpro_is_spammer() {
+    $is_spammer = false;
+    
+    $activity = pmpro_get_spam_activity();
+    if ( count( $activity ) >= PMPRO_SPAM_ACTION_NUM_LIMIT ) {
+        $is_spammer = true;
+    }
+    
+    return apply_filters( 'pmpro_is_spammer', $is_spammer, $activity );
```
sc0ttkclark, line 25:
```suggestion
	/**
	 * Allow filtering whether the current visitor is a spammer.
	 *
	 * @since 2.7
	 *
	 * @param bool  $is_spammer Whether the current visitor is a spammer.
	 * @param array $activity   The list of potential spam activity.
	 */
	return apply_filters( 'pmpro_is_spammer', $is_spammer, $activity );
```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Is the current visitor a spammer?
+ * @since 2.7
+ */
+function pmpro_is_spammer() {
+    $is_spammer = false;
+    
+    $activity = pmpro_get_spam_activity();
+    if ( count( $activity ) >= PMPRO_SPAM_ACTION_NUM_LIMIT ) {
```
sc0ttkclark, line 21: Exclude `false` if we could not determine IP.

```suggestion
    if ( false !== $activity && count( $activity ) >= PMPRO_SPAM_ACTION_NUM_LIMIT ) {

```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+    
+    return apply_filters( 'pmpro_is_spammer', $is_spammer, $activity );
+}
+
+/**
+ * Get spam activity by IP.
+ * @since 2.7
+ * @param string $ip    The IP address to get activity for.
+ * @return array|False  The array of activity if successful, false if no IP. 
+ */
```
sc0ttkclark, line 33:
```suggestion
/**
 * Get the list of potential spam activity.
 *
 * @since 2.7
 *
 * @param string|null $ip The IP address to get activity for, or leave as null to attempt to determine current IP address.
 *
 * @return array|false The list of potential spam activity if successful, or false if IP could not be determined.
 */
```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+    if ( empty( $ip ) ) {
+        $ip = pmpro_get_ip();        
+    }
+    
+    // If we can't determine the IP, let's bail.
+    if ( empty( $ip ) ) {
+        return false;
+    }
+    
+    $transient_key = 'pmpro_spam_activity_' . $ip;
```
sc0ttkclark, line 44: If $ip is provided, should we sanitize/clean `$ip` in any way before getting transient?

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+    $activity = get_transient( $transient_key );    
+    if ( empty( $activity ) ) { $activity = []; }
```
sc0ttkclark, line 46:
```suggestion
	if ( empty( $activity ) || ! is_array( $activity ) ) {
		$activity = [];
	}

```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+    
+    // Remove old items.
+	$new_activity = [];
+	$now = current_time( 'timestamp', true ); // UTC
+    foreach( $activity as $item ) {
+		if ( $item > $now-( PMPRO_SPAM_ACTION_TIME_LIMIT ) ) {
```
sc0ttkclark, line 52:
```suggestion
		// Determine whether this item is recent enough to include.
		if ( $item > $now-( PMPRO_SPAM_ACTION_TIME_LIMIT ) ) {
```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Disable checkout and billing update forms for spammers.
+ *
+ * We're using the pmpro_required_billing_fields filter because it is
+ * consistently used on the checkout and update billing pages.
+ * The pmpro_setMessage() function sets the $pmpro_msgt value to error,
+ * which stops the forms from working and shows the corresponding error.
+ * We return the $required_fields parameter to keep the filter working.
+ *
+ * @since 2.7
```
sc0ttkclark, line 130:
```suggestion
 * @since 2.7
 *
 * @param array $required_fields The list of required fields.
 *
 * @return array The list of required fields.
```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+ */
+function pmpro_disable_checkout_for_spammers( $required_fields ) {    
+    if ( pmpro_was_checkout_form_submitted() && pmpro_is_spammer() ) {
+        pmpro_setMessage( __( 'Suspicious activity detected. Try again in a few minutes.', 'paid-memberships-pro' ), 'pmpro_error' );        
+    }
+    
+    return $required_fields;
+}
+add_filter( 'pmpro_required_billing_fields', 'pmpro_disable_checkout_for_spammers', 10, 1 );
```
sc0ttkclark, line 139: We don't need to specify priority/args as they both default to 10 and 1

```suggestion
add_filter( 'pmpro_required_billing_fields', 'pmpro_disable_checkout_for_spammers' );
```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+    if ( empty( $ip ) ) {
+        $ip = pmpro_get_ip();        
+    }
+    
+    // If we can't determine the IP, let's bail.
+    if ( empty( $ip ) ) {
+        return false;
+    }
+    
+    $transient_key = 'pmpro_spam_activity_' . $ip;
```
sc0ttkclark, line 116: If $ip is provided, should we sanitize/clean `$ip` in any way before getting transient?

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+    
+    delete_transient( $transient_key );
```
sc0ttkclark, line 118:
```suggestion
	delete_transient( $transient_key );

	return true;
```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Clears all stored spam activity for an IP address.
+ * Note that the pmpro_get_spam_activity function clears out old values
+ * automatically, and this should only be used to completely clear the activity.
+ *
+ * @since 2.7
+ * @param string $ip    The IP address to clear activity for.
+ * @return bool         True if successful. False if we couldn't determine the IP.
```
sc0ttkclark, line 104:
```suggestion
 * @since 2.7
 *
 * @param string|null $ip The IP address to clear activity for, or leave as null to attempt to determine current IP address.
 *
 * @return bool True if the clearing of activity was successful, or false if IP could not be determined.
 */
```

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+    array_unshift( $activity, $now );
+    
+    // If we have more than the limit, don't bother storing them.
+    if ( count( $activity ) > PMPRO_SPAM_ACTION_NUM_LIMIT ) {
+        rsort( $activity );
+        $activity = array_slice( $activity, 0, PMPRO_SPAM_ACTION_NUM_LIMIT );
+    }
+    
+    // Save to transient.
+    $transient_key = 'pmpro_spam_activity_' . $ip;
```
sc0ttkclark, line 91: If $ip is provided, should we sanitize/clean `$ip` in any way before getting transient?

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Track spam activity.
+ * When we hit a certain number, the spam flag will trigger.
+ * For now we are only tracking credit card declines their timestamps.
+ * IP address isn't a perfect way to track this, but it's the best we have.
+ *
+ * @since 2.7
+ * @param string $ip    The IP address to track activity for.
+ * @return bool         True if successful. False if we couldn't determine the IP.
```
sc0ttkclark, line 68:
```suggestion
 * @since 2.7
 *
 * @param string|null $ip The IP address to track activity for, or leave as null to attempt to determine current IP address.
 *
 * @return bool True if the tracking of activity was successful, or false if IP could not be determined.
```

sc0ttkclark (commented):
Looks awesome, I added a few Qs/notes and suggested a few phpdoc improvements here.

includes/spam.php
```diff
@@ -0,0 +1,139 @@
...
+    
+    // If we have more than the limit, don't bother storing them.
+    if ( count( $activity ) > PMPRO_SPAM_ACTION_NUM_LIMIT ) {
+        rsort( $activity );
+        $activity = array_slice( $activity, 0, PMPRO_SPAM_ACTION_NUM_LIMIT );
+    }
+    
+    // Save to transient.
+    $transient_key = 'pmpro_spam_activity_' . $ip;
+    set_transient( $transient_key, $activity );
```
sc0ttkclark, line 92: If we intend to keep the footprint small, this will work against that. We need to set an expiration for this transient. WP will autoload all options in the DB which have autoload set to 'yes' and non-expiring transients get that set to 'yes'.

This should likely be set to `PMPRO_SPAM_ACTION_TIME_LIMIT`. If it has expired, then all of the activity was at/beyond the time limit intended and it still continues to function as we want.

```suggestion
    set_transient( $transient_key, $activity, (int) PMPRO_SPAM_ACTION_TIME_LIMIT );
```

preheaders/billing.php
ideadude: I like it. We can add a pmpro_update_billing_failed hook here, and a similar pmpro_checkout_process_failed on the checkout page then hook into them.

The pmpro_track_spam_activity function can be called with no params so it's easy to hook up.

preheaders/checkout.php
ideadude: Yeah. I like pmpro_checkout_process_failed. Probably pass the order into it.

includes/functions.php
ideadude: In theory, I like the idea that $_SERVER['REMOTE_ADDR'] is meant to be the client's IP address and servers should just fix their configuration. In practice, this isn't going to happen and we need to do what OUR users expect, which is figure out the IP address as well as possible from the data given.

Given that, in practice, all of these other places for the IP trump the REMOTE_ADDR on servers. I went with some combination of the JetPack and Community Events implementations. I think it's most important that REMOTE_ADDR comes last, as that's the one that is most often incorrect in practice.

includes/spam.php
ideadude: Earlier those constants were used in both the get and update functions, so I needed them defined outside of the functions. But in the current implementation, they are only used in one function each. So making them params could work. I like it.

Then you could unhook our functions and add your own call backs that called the same functions, but with the parameters you wanted.

I still have a feeling that these 2 numbers might be used in different contexts as we build on this. But we can code this how it's best for now and cross that bridge when we get there. Constants are difficult to set up and manage and should be avoided if possible.

includes/spam.php
ideadude: No. Probably happened when copying some of this code from a different branch on GitHub or something. I like tabs 4 spaces long. I can lint when we're all done.

includes/spam.php
ideadude: Yeah, let's sanitize the IP. This is what we use in the get ip function, what WP core does in some other spots.

`$client_ip = preg_replace( '/[^0-9a-fA-F:., ]/', '', $client_ip );`

dparker1005:
Looks great! Only question, looks like this behavior is currently defaulted to enabled. Is this what we want or did I misread the code?

includes/spam.php
ideadude: The constants are used across different functions, so leaving as constants for now. May refactor later.

PR #1919, opened 2022-01-23
Remove unused condition on sending mail

freax:
### All Submissions:

No need to check on returning value of `wp_mail`.

### Other information:

### Changelog entry

No need to check on returning value of `wp_mail`, `wp_mail` returns boolean by itself.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,7 +1,8 @@
 == Changelog ==
 = 2.7.2 - 2022-01-17 =
-* BUG FIX/ENHANCEMENT: Improved error messages when creating subscriptions with Stripe.
+* BUG FIX/ENHANCEMENT: Improved error messages when creating subscriptions with Stripe. (@dparker1005)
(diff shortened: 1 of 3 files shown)


sc0ttkclark (approved):
Great code clean up

ideadude:
Thanks for the effort here.

This PR has started a discussion on our end about how to handle PRs that only refactor and unrequested PRs in general. We appreciate the help and getting new developers into the fold, but also these kinds of innocuous, seemingly obvious, changes have caused issues in the past and we'd like to avoid that.

In this case, I confirmed in the WP docs and code that the wp_mail function does indeed only return true or false. It won't return null, WP_Error or anything else. So I'm pretty certain we can safely merge this in.

Still, it has taken some time, we're not always so sure there is no negative effect in our complicated codebase, and if we make changes to the email class or functions in the future, we might have to totally rewrite this stuff anyway. So we want to limit refactor efforts to areas that either enhance how developers work with our code or otherwise set the stage for future work.

I'm going to merge this in and push it out in the next release. Thanks again for the contribution. This long comment here is meant for us internally to figure out our processes for PRs like this and how we can update our contributing guidelines.

Issue #1922, opened 2022-01-24
Precision configuration is not accounted for when using floats and displaying/sending them to APIs


sc0ttkclark:
**Describe the bug**
PHP has quite a nasty challenge to workaround in regards to float values.

* [JSON encode float issues](https://stackoverflow.com/questions/42981409/php7-1-json-encode-float-issue)
* [PHP floating point issue with precision workaround](https://stackoverflow.com/questions/14587290/can-i-rely-on-php-php-ini-precision-workaround-for-floating-point-issue)
* [WooCommerce precision](https://wordpress.org/support/topic/implications-of-changing-serialize_precision-to-1/)
* [WooCommerce rounding issues](https://github.com/woocommerce/woocommerce/issues/24975) (this is not related but still good to reference as we adjust how we handle precision)

The problem is very very complicated. Many hosts have many varied precision configurations. Some sites have overrides for these due to other issues encountered and recommendations on changing the setting.

The issue seems to go away after testing with precision 11 / serialize_precision 11 but there may still be other cases because of the challenging aspect of the issue: It doesn't happen for every float and has the word "EDGE" written all over it as you reduce the precision.

**To Reproduce**
Need to figure out exactly the cases / settings to reproduce this perfectly. A customer reported 19.90 as a membership level price that is bugged by this.

In Site Health, the PMPro section for Membership Levels shows values like this:

```json
{
[
...
“initial_payment”: 19.89999999999999857891452847979962825775146484375,
“billing_amount”: 19.89999999999999857891452847979962825775146484375,
...
]
}
```

**Expected behavior**
If a floating precision issue is detected, attempt to remove the extra precision.

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only Paid Memberships Pro plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Memberlite](https://wordpress.org/themes/memberlite/).
- [x] I can reproduce this bug consistently using the steps above.

sc0ttkclark:
Confirmed this issue happens on Local by Flywheel and on Nexcess hosting, but in different ways.

`json_encode()` is one of the most likely areas for this to happen on hosts, if not on both general float usage as well as `json_encode()`. Some PHP output handling will use the `serialize_precision` setting instead of `precision` so that complicates testing/confirming the issue as well.

PR #1929, opened 2022-01-28
When passing amounts to gateways, pass as a string with precision issues addressed

sc0ttkclark:
Fixes https://github.com/strangerstudios/paid-memberships-pro/issues/1922

### All Submissions:

Resolves problems with float precision and prevent passing along faulty floats to the gateway APIs.

1. Test with a membership level priced at $19.90
2. Test with one-time
3. Test with recurring
4. Test with trial at $19.90
5. Test checkout for every gateway

### Other information:

### Changelog entry

> BUG FIX: Resolved problems with PHP float precision and prevent passing along faulty floats to the gateway APIs.


sc0ttkclark:
I'm working on writing tests for the new functions introduced here, then I will set this PR as ready for review.

includes/functions.php
```diff
@@ -3119,6 +3119,97 @@ function pmpro_round_price( $price, $currency = '' ) {
...
+
+	$amount = (float) $amount;
+
+	$currency_info = pmpro_get_currency( $currency );
+
+	$price_info = [
+		// The amount represented as a float.
+		'amount'        => $amount,
+		// The flat amount represent (example: 1.99 would be 199).
+		'amount_flat'   => 0,
```
dparker1005, line 3167:
```suggestion
		'amount_flat'   => 0,
		// The number of decimals for this currency.
		'decimals'        => (int) $currency_info['decimals'],
```

includes/functions.php
```diff
@@ -3119,6 +3119,97 @@ function pmpro_round_price( $price, $currency = '' ) {
...
+			// The decimal part of the amount as a string (example: 1.99 would be 99, 1.00 would be 00).
+			'decimal_string' => '',
+		],
+		// The currency information.
+		'currency'      => $currency_info,
+	];
+
+	// Enforce integer.
+	$currency_info['decimals'] = (int) $currency_info['decimals'];
+
```
dparker1005, line 3184:
```suggestion
```

dparker1005 (approved):
Looks good, just a quick suggestion to immediately define the 'decimals' key in the $currency_info array

includes/functions.php
sc0ttkclark: Just checking here -- the decimals value is a string in many cases, enforcing as integer sets the expectation up for the below logic.

includes/functions.php
sc0ttkclark: What was the purpose of adding this here if we pass it in `$price_info['currency']` already?

ideadude:
I took a look as well and tested some different checkout scenarios. All looks good.

RE the test code. Is it possible that the tests like this wouldn't work on systems that aren't having the same precision power. The test overrides the settings, but I thought this issue might also be related to PHP versions and other server settings.

`
/**
 * @covers ::pmpro_round_price()
 */
public function test_precision_issue_with_float() {
	$this->assertEquals( '19.899999999999999', json_encode( 19.9 ) );
	$this->assertEquals( '19.890000000000001', json_encode( 19.89 ) );
}
`

So json_encode(19.9) might not break on some systems?

Should we leave the tests out, or are there other ways to confirm that the new functions return the results we expect in certain situations?

sc0ttkclark:
@ideadude The test itself is meant to set the settings to those that are the normal defaults for PHP regardless of configuration changes. That way, it can be sure it's testing and confirming based on the default PHP config that may be used across most sites. I haven't known many PHP configs to have those settings customized, however the note on PHP.net specifies why setting this manually here is most important:

From [PHP.net](https://www.php.net/manual/en/ini.core.php):

> Prior to PHP 7.1.0, the default value was 17.

So now [serialize_precision](https://www.php.net/manual/en/ini.core.php#ini.serialize-precision) is set to the value of `-1` by default in PHP 7.1+ so I'm just setting it to the most common denominator for testing purposes.

PR #1940, opened 2022-02-07
Better error handling when Stripe Price not found

dparker1005:
### All Submissions:

A couple of users have reported seeing the following error message while using Stripe at checkout: `Error creating subscription for customer. Cannot get price.`

This PR will modify that error message to also include the specific reason that the price could not be retrieved, which is generated by Stripe. This change will help us better assist users who are shown this error message in the future.

### Other information:

### Changelog entry
* ENHANCEMENT: Improved error handling for Stripe Prices.


classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -2424,8 +2426,8 @@ private function create_subscription_for_customer_from_order( $customer_id, $ord
 		}
 
 		$price = $this->get_price_for_product( $product_id, $amount, $order->BillingPeriod, $order->BillingFrequency );
-		if ( empty( $price ) ) {
-			$order->error = esc_html__( 'Cannot get price.', 'paid-memberships-pro' );
+		if ( is_string( $price ) ) {
```
sc0ttkclark, line 2429: This may still be returning as null in some cases

classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -2356,12 +2356,14 @@ private function get_price_for_product( $product_id, $amount, $cycle_period = nu
 		try {
 			$price = Stripe_Price::create( $price_args );
 			if ( ! empty( $price->id ) ) {
```
sc0ttkclark, line 2358: If it doesn't make it through this statement, the function will return as null. Is it possible that `Stripe_Price::create()` could not trigger an exception and return something? Sounds edge case-y but it may be a good idea to cover that case below in `create_subscription_for_customer_from_order()` instead of just checking for `is_string()`

sc0ttkclark (changes requested):
Minor issue here to deal with before we merge, just need to handle null cases

classes/gateways/class.pmprogateway_stripe.php
dparker1005: Hmm, good call. I don't think reaching the end of that function should ever be possible, but accounting for it anyways would be good. Instead of returning null and having to account for it everywhere, would you rather just instead return a string `'Price could not be created'` and just have the "error" return always be a string?

sc0ttkclark: Sounds good to me, let's return a string there

sc0ttkclark (approved):

PR #1951, opened 2022-02-16
Codebase housekeeping

mircobabini:
This is an idea of what can be done when running a static analyzer on functions.php.

During this analysis I tried to:
1. Fix all `Variable '$aaaaa' is probably undefined`;
2. Avoid `Unnecessary local variable` when can be replaced with a one-liner return statement;
3. Fix issues in docblocks (wrong var names, wrong return types...) and complete almost ready docblocks;
4. Fix some `'if' statement with common parts`;
5. Add global variables where missing;
6. Removed unused function params.

There are many more occurrence of the above and many notices which I ignored.
I think we should define a strategy and start doing some _prioritized_ maintenance:
- avoid errors (missing global statements, missing/wrong @throws @return in docblocks...)
- avoid warnings (fix all "local variable is probably undefined")
- enhance readability (avoid local variables when returning, avoid 'if' statement with common parts)
- add missing docblocks

About the code structure:
- start from functions.php
- move to MemberOrder.php and gateways
- etc...

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -43,7 +43,13 @@ function pmpro_setDBTables() {
 }
 pmpro_setDBTables();
 
-// thanks: http://wordpress.org/support/topic/is_plugin_active
+/**
+ * @param string $plugin
+ *
+ * @return bool
+ *
+ * thanks: http://wordpress.org/support/topic/is_plugin_active
+ */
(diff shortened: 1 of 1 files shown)


mircobabini:
Gonna close this one as soon as I have time to create another housekeeping PR _**without**_ refactoring.

PR #1952, opened 2022-02-16
Autoload ->Gateway when needed

mircobabini:
### All Submissions:

We had an issue discussed on slack; because of `setGateway` loaded to default gateway on `MemberOrder::__construct()`, we found a case in which the `->Gateway` instance does not reflect the `->gateway` type.

Code to reproduce:
```php
// I have Stripe as active Gateway, + PPE Addon
$order            = new MemberOrder();
$order->nogateway = true;
$order->getMemberOrderByCode( 'E9A0ED3C5F' ); // PPE order
// $order->setGateway( $order->gateway );

echo $order->gateway; // paypalexpress !!!!
echo get_class( $order->Gateway ); // PMProGateway_stripe !!!!
```

1. Add the code above somewhere using a PPE order (with stripe as default gateway)
2. Check the dumps
3. Apply the patchs and re-check the dumps

### Other information:

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -6,9 +6,6 @@
 		 */
 		function __construct($id = NULL)
 		{
-			//set up the gateway
-			$this->setGateway(pmpro_getOption("gateway"));
-
 			//get data if an id was passed
 			if($id)
 			{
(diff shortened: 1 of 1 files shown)


classes/class.memberorder.php
```diff
@@ -142,8 +158,11 @@ function getMemberOrderByID($id)
 				$this->checkout_id = $dbobj->checkout_id;
 
 				//reset the gateway
-				if(empty($this->nogateway))
+				if( empty( $this->nogateway ) ) {
 					$this->setGateway();
+				} else {
+					unset( $this->Gateway );
```
mircobabini, line 164: Must `unset` to be sure the `__get` is called. This should be safe to be done every time the ->nogateway is specified, but probably useless.

classes/class.memberorder.php
```diff
@@ -21,6 +18,25 @@ function __construct($id = NULL)
 				return $this->getEmptyMemberOrder();	//blank constructor
 		}
 
+		public function __get( $name )
+		{
+			if( 'Gateway' === $name ){
+				// we are here only if ->Gateway is NOT set
+				$this->setGateway( $this->gateway ? $this->gateway : '' );
```
mircobabini, line 25: Use empty string if $this->gateway is null, to be sure the `->Gateway` instance will be set (to the test gateway) and avoid fatals.

mircobabini:
As discussed with @dparker1005, the idea is to call `->setGateway()` when you want to use the `->Gateway` object. No magic tricks.

PR #1957, opened 2022-02-18
Feature/license updates

ideadude:
### All Submissions:

Now using the v2 endpoints to check licenses and get add on information.

Added functions and code to differentiate between standard and plus/builder plugins. This version will correctly warn standard users that they can't update plus plugins. But plus and builder licenses can update standard plugins.

### Other information:

### Changelog entry

### DIFF
--- a/includes/addons.php
+++ b/includes/addons.php
@@ -237,9 +237,12 @@ function pmpro_getPluginAPIObjectFromAddon( $addon ) {
 	if ( ! empty( $key ) && ! empty( $api->package ) ) {
 		$api->package = add_query_arg( 'key', $key, $api->package );
 	}
-	if ( empty( $api->upgrade_notice ) && ! pmpro_license_isValid( null, 'plus' ) ) {
-		$api->upgrade_notice = __( 'Important: This plugin requires a valid PMPro Plus license key to update.', 'paid-memberships-pro' );
-	}
+	
+	if ( empty( $api->upgrade_notice ) && pmpro_license_type_is_premium( $addon['License'] ) ) {
+		if ( ! pmpro_license_isValid( null, $addon['License'] ) ) {
+			$api->upgrade_notice = sprintf( __( 'Important: This plugin requires a valid PMPro %s license key to update.', 'paid-memberships-pro' ), ucwords( $addon['License'] ) );
+		}
+	}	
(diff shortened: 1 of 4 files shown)


includes/addons.php
```diff
@@ -271,28 +274,39 @@ function pmpro_admin_init_updating_plugins() {
...
-			echo '<div class="error"><p>' . $msg . ' <strong>' . implode( ', ', $plus_addons ) . '</strong></p></div>';
+		if ( ! empty( $premium_plugins ) ) {			
+			foreach( $premium_plugins as $license_type => $premium_plugin ) {				
+				// if they have a good license, skip the error				
+				if ( pmpro_can_download_addon_with_license( $license_type ) ) {
+					continue;
+				}
+				
+				// show error
+				$msg = sprintf( __( 'You must have a <a href="https://www.paidmembershipspro.com/pricing/?utm_source=wp-admin&utm_pluginlink=bulkupdate">valid PMPro %s License Key</a> to update PMPro %s add ons. The following plugins will not be updated:', 'paid-memberships-pro' ), ucwords( $license_type ), ucwords( $license_type ) );
```
sc0ttkclark, line 307: It's painful but we should reconsider usage of HTML in the translation text like this at some point, see:

https://core.trac.wordpress.org/changeset/45542

It's okay for now but we need to approach our strings with more consideration that we don't want HTML in the output of the strings.

includes/license.php
```diff
@@ -69,8 +69,12 @@ function pmpro_license_isValid($key = NULL, $type = NULL, $force = false) {
 	}
 	
 	// Check if a specific type.
-	if ( ! empty( $type ) && $type != $pmpro_license_check['license'] ) {
-		return false;
+	if ( ! empty( $type ) ) {
+		$type = (array)$type;
+		
+		if ( ! in_array( $pmpro_license_check['license'], $type ) ) {
```
sc0ttkclark, line 75: Best practices for WP PHPCS is to use third param as `true`

```suggestion
		if ( ! in_array( $pmpro_license_check['license'], $type, true ) ) {

```

includes/addons.php
ideadude: Will fix in a separate PR.

PR #1960, opened 2022-02-18
Subs-table-save

dparker1005:
Added some validation in save method of subscription class

### DIFF
--- a/classes/class-pmpro-subscription.php
+++ b/classes/class-pmpro-subscription.php
@@ -981,30 +981,16 @@ class PMPro_Subscription {
 			return false;
 		}
 
-		// TODO: Perform validation here. The setter should make sure that fields have valid values,
-		// but we should also check here that the values make sense together. For example, cancelled
-		// subscriptions should not have a next payment date and startdates should not be before end dates.
+		// Active subscriptions shouldn't have an enddate yet, and cancelled subscriptions shouldn't have a next payment date.
+		if ( 'active' === $this->status ) {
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
@dparker1005 Did we intend to remove hooks here?

classes/class-pmpro-subscription.php
```diff
@@ -981,30 +981,16 @@ public function save() {
 			return false;
 		}
 
-		// TODO: Perform validation here. The setter should make sure that fields have valid values,
-		// but we should also check here that the values make sense together. For example, cancelled
-		// subscriptions should not have a next payment date and startdates should not be before end dates.
-
-		$create = empty( $this->id );
-
-		if ( $create ) {
```
sc0ttkclark, line 990: @dparker1005 Did we intend to remove hooks here?

sc0ttkclark (commented):
Q

classes/class-pmpro-subscription.php
dparker1005: I took it out in case we wanted to rethink this logic. How this was originally coded is how our save order function works, but I'm not sure how frequently those hooks are ever used. What seems more useful to me is how WC uses hooks around subscription status changes where you can hook into a subscription changing to active, changing to canceled, etc. So I removed this just so that we're not committed to a single hook approach and we can discuss exactly what hooks we want/how they should work instead of just blindly adding them

sc0ttkclark (approved):

PR #1965, opened 2022-02-23
Subs-table-default-next-payment-date

dparker1005:
Allowing for updating PMPro Subscriptions from unsupported gateways

### DIFF
--- a/classes/class-pmpro-subscription.php
+++ b/classes/class-pmpro-subscription.php
@@ -641,10 +641,8 @@ class PMPro_Subscription {
 		// Create the subscription.
 		$new_subscription = new PMPro_Subscription( $subscription_data );
 
-		// Try to pull as much info as possible directly from the gateway.
-		$new_subscription->update_from_gateway();
-
-		$saved = $new_subscription->save();
+		// Try to pull as much info as possible directly from the gateway or from the database.
+		$saved = $new_subscription->update();
 		if ( ! $saved ) {
 			// We couldn't save the subscription.
 			return null;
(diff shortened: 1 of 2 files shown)


classes/class.memberorder.php
```diff
@@ -1103,11 +1098,26 @@ function get_test_order() {
...
 				return false;
 			}
 
+			// Make sure that this order has been completed.
+			if ( 'success' !== $this->status ) {
+				// The order is not complete, so we shouldn't create a subscription yet.
+				return false;
+			}
+
+			// Make sure that the user has the membership level for this order.
```
sc0ttkclark, line 1115: During checkout, does the order get processed and membership get set up before the subscription is supposed to be set up? This logic might not be right, but I'd like to look at it closer with you.

classes/class-pmpro-subscription.php
```diff
@@ -654,16 +652,47 @@ public static function create( $args ) {
...
+			if ( ! empty( $orders ) ) {
+				$order = current( $orders );
+				$this->startdate = date_i18n( 'Y-m-d H:i:s', $order->getTimestamp( true ) );
+			}
+
+			// Update the next payment date based on the most recent order.
+			if ( ! empty( $this->cycle_number ) ) {
+				// Only update the next payment date if we are not at checkout or there is no next payment date already set.
+				if ( ! isset( $pmpro_level ) || empty( $this->next_payment_date ) ) {
+					$orders = $this->get_orders( array( 'limit' => 1 ) );
```
sc0ttkclark, line 684: Should we reuse the $orders we received before at line 671?

classes/class.memberorder.php
dparker1005: In the checkout preheader, the subscription is set up on 446:
https://github.com/strangerstudios/paid-memberships-pro/blob/5ca74f946d62c7783bb248fd8d128289b39f6d68/preheaders/checkout.php#L446

And then the membership level is changed at 664 and, if successful, the order is saved on 680:
https://github.com/strangerstudios/paid-memberships-pro/blob/5ca74f946d62c7783bb248fd8d128289b39f6d68/preheaders/checkout.php#L664-L681

So by the time that the order is saved, the initial payment should be processed, a subscription set up, and a membership assigned.

classes/class-pmpro-subscription.php
dparker1005: I don't know if we can. On 671, we want the oldest order to get the subscription startdate. On 684, we want to get the most recent order to guess at the next payment date.

dparker1005: I guess we can grab them all and just pick out the first and last though

sc0ttkclark: If we rename `$orders` here to `$newest_orders` that would make it better for code readability and understanding intent too -- since we reuse `$orders` up above

sc0ttkclark (approved):

PR #1981, opened 2022-02-28
Enhancement/colon search

ideadude:
### All Submissions:

The default members list search runs the query against the user_login, user_email, display_name and THE ENTIRE USER META TABLE.

On sites with lots of users and/or user meta fields, these queries and searches can get slow.

This update checks for ":" in the search query and if found, it will split the query up and query against a specific user table column or user meta field. This speeds up the query A LOT.

This also allows you to search against specific user meta fields, which helps for search terms that might show up in multiple, but unrelated meta fields.

So for example `pmpro_bstate:pa` would return all members with the billing state of "pa". A search for `email:email@domain.com` would return just the user with that specific email.

This PR also swaps * for % in the search query.

User meta that might have * or : in them will break now. We don't allow for escaping these characters in the query. We could do that by swapping out ** and %% or \* and \% and then swapping them back in. I don't think that's needed. These characters aren't allowed in user names or email addresses and so shouldn't come up that often.

It would be nice in a future update to somehow share this hidden feature with users. We could e.g. detect a slow query or a query for an email address and share some text to perform the colon search.

1. Search the members list for things like `email:email@domain.com`, `first_name:Jason`, or `meta_key:meta_value`/etc.
2. Make sure the results match.
3. Export the results to CSV and make sure they also match.

### Other information:

### Changelog entry

> ENHANCEMENT: You can now search the members list on specific user table columns or user meta fields by using a colon in your search term. These queries are faster than the default queries. The format is `meta_key:meta_value` (no backticks). You can also use login, nicename, email, url, or display_name as the meta_key and the users table will be searched against the related column.

### DIFF
--- a/adminpages/memberslist-csv.php
+++ b/adminpages/memberslist-csv.php
@@ -37,20 +37,20 @@
 	//get users (search input field)
 	$search_key = false;
 	if( isset( $_REQUEST['s'] ) ) {
-		$s = sanitize_text_field( trim( $_REQUEST['s'] ) );		
+		$s = sanitize_text_field( trim( $_REQUEST['s'] ) );
 	} else {
 		$s = '';
 	}
-	
+
 	// If there's a colon in the search, let's split it out.
-	if( ! empty( $s ) && strpos( $s, ':' ) !== false ) {				
+	if( ! empty( $s ) && strpos( $s, ':' ) !== false ) {
 		$parts = explode( ':', $s );
 		$search_key = $parts[0];
 		$s = $parts[1];
 	}
-	
+
 	// Treat * as wild cards.
-	$s = str_replace( '*', '%', $s );	
+	$s = str_replace( '*', '%', $s );
 
 	// requested a level id
 	if(isset($_REQUEST['l']))
@@ -154,7 +154,7 @@
 		SELECT
 			DISTINCT u.ID
 		FROM $wpdb->users u ";
-		
+
 	if ( $s ) {
 		if ( ! empty( $search_key ) ) {
 			// If there's a colon in the search string, make the search smarter.
@@ -171,11 +171,11 @@
 		} else {
 			// Default search checks a few fields.
 			$sqlQuery .= "LEFT JOIN {$wpdb->usermeta} um ON u.ID = um.user_id ";
-			$search = " AND ( u.user_login LIKE '%" . esc_sql($s) . "%' OR u.user_email LIKE '%" . esc_sql($s) . "%' OR um.meta_value LIKE '%" . esc_sql($s) . "%' OR u.display_name LIKE '%" . esc_sql($s) . "%' ) ";			
+			$search = " AND ( u.user_login LIKE '%" . esc_sql($s) . "%' OR u.user_email LIKE '%" . esc_sql($s) . "%' OR um.meta_value LIKE '%" . esc_sql($s) . "%' OR u.display_name LIKE '%" . esc_sql($s) . "%' ) ";
 		}
 	} else {
 		$search = '';
-	}		
+	}
 
 	$sqlQuery .= "LEFT JOIN {$wpdb->pmpro_memberships_users} mu ON u.ID = mu.user_id ";
 	$sqlQuery .= "LEFT JOIN {$wpdb->pmpro_membership_levels} m ON mu.membership_id = m.id ";
@@ -193,7 +193,7 @@
 
 	// looking for a specific user
 	if ( ! empty( $s ) ) {
-		$sqlQuery .= $search;		
+		$sqlQuery .= $search;
 	}
 
 	// if ($former_members)
@@ -324,7 +324,7 @@
 
 		//increment starting position
 		$i_start += $max_users_per_loop;
-		
+
 		//escape the % for LIKE comparison with $wpdb
 		if(!empty($search))
 			$search = str_replace('%', '%%', $search);
--- a/adminpages/memberslist-csv.php
+++ b/adminpages/memberslist-csv.php
@@ -37,17 +37,20 @@
 	//get users (search input field)
 	$search_key = false;
 	if( isset( $_REQUEST['s'] ) ) {
-		$s = sanitize_text_field( trim( $_REQUEST['s'] ) );
-		
-		// If there's a colon in the search, let's split it out.
-		if( ! empty( $s ) && strpos( $s, ':' ) !== false ) {				
-			$parts = explode( ':', $s );
-			$search_key = $parts[0];
-			$s = $parts[1];
-		}
+		$s = sanitize_text_field( trim( $_REQUEST['s'] ) );		
 	} else {
 		$s = '';
-	}	
+	}
+	
+	// If there's a colon in the search, let's split it out.
+	if( ! empty( $s ) && strpos( $s, ':' ) !== false ) {				
+		$parts = explode( ':', $s );
+		$search_key = $parts[0];
+		$s = $parts[1];
+	}
+	
+	// Treat * as wild cards.
+	$s = str_replace( '*', '%', $s );	
 
 	// requested a level id
 	if(isset($_REQUEST['l']))
--- a/adminpages/memberslist-csv.php
+++ b/adminpages/memberslist-csv.php
@@ -154,9 +154,28 @@
 		SELECT
 			DISTINCT u.ID
 		FROM $wpdb->users u ";
-
-	if ($s)
-		$sqlQuery .= "LEFT JOIN {$wpdb->usermeta} um ON u.ID = um.user_id ";
+		
+	if ( $s ) {
+		if ( ! empty( $search_key ) ) {
+			// If there's a colon in the search string, make the search smarter.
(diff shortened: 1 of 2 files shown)


adminpages/memberslist-csv.php
```diff
@@ -35,10 +35,22 @@
...
-		$s = "";
+	$search_key = false;
+	if( isset( $_REQUEST['s'] ) ) {
+		$s = sanitize_text_field( trim( $_REQUEST['s'] ) );		
+	} else {
+		$s = '';
+	}
+	
+	// If there's a colon in the search, let's split it out.
+	if( ! empty( $s ) && strpos( $s, ':' ) !== false ) {
```
sc0ttkclark, line 46:
```suggestion
	if( ! empty( $s ) && strpos( $s, ':' ) !== false ) {
```

adminpages/memberslist-csv.php
```diff
@@ -35,10 +35,22 @@
 	global $wpdb;
 
 	//get users (search input field)
-	if(isset($_REQUEST['s']))
-		$s = sanitize_text_field($_REQUEST['s']);
-	else
-		$s = "";
+	$search_key = false;
+	if( isset( $_REQUEST['s'] ) ) {
+		$s = sanitize_text_field( trim( $_REQUEST['s'] ) );
```
sc0ttkclark, line 40:
```suggestion
		$s = sanitize_text_field( trim( $_REQUEST['s'] ) );
```

adminpages/memberslist-csv.php
```diff
@@ -35,10 +35,22 @@
...
+	} else {
+		$s = '';
+	}
+
+	// If there's a colon in the search, let's split it out.
+	if( ! empty( $s ) && strpos( $s, ':' ) !== false ) {
+		$parts = explode( ':', $s );
+		$search_key = $parts[0];
+		$s = $parts[1];
```
sc0ttkclark, line 49: Should we use `array_shift()` here so that subsequent `:` will get ignored but still used in the search string itself? Does it matter?

```suggestion
		$search_key = array_shift( $parts );
		$s = implode( ':', $parts );
```

adminpages/memberslist-csv.php
```diff
@@ -143,8 +155,27 @@
 			DISTINCT u.ID
 		FROM $wpdb->users u ";
 
-	if ($s)
-		$sqlQuery .= "LEFT JOIN {$wpdb->usermeta} um ON u.ID = um.user_id ";
+	if ( $s ) {
+		if ( ! empty( $search_key ) ) {
+			// If there's a colon in the search string, make the search smarter.
+			if( in_array( $search_key, array( 'login', 'nicename', 'email', 'url', 'display_name' ) ) ) {
```
sc0ttkclark, line 161:
```suggestion
			if( in_array( $search_key, array( 'login', 'nicename', 'email', 'url', 'display_name' ), true ) ) {
```

classes/class-pmpro-members-list-table.php
```diff
@@ -252,11 +252,24 @@ private function sql_table_data( $count = false ) {
...
+			$s = sanitize_text_field( trim( $_REQUEST['s'] ) );			
+		} else {
+			$s = '';
+		}
+		
+		// If there's a colon in the search, let's split it out.
+		if( ! empty( $s ) && strpos( $s, ':' ) !== false ) {				
+			$parts = explode( ':', $s );
+			$search_key = $parts[0];
+			$s = $parts[1];
```
sc0ttkclark, line 267: Should we use `array_shift()` here so that subsequent `:` will get ignored but still used in the search string itself? Does it matter?

```suggestion
		$search_key = array_shift( $parts );
		$s = implode( ':', $parts );
```

classes/class-pmpro-members-list-table.php
```diff
@@ -308,7 +321,23 @@ private function sql_table_data( $count = false ) {
 			";
 			
 		if ( !empty( $s ) ) {
-			$sqlQuery .= " LEFT JOIN $wpdb->usermeta um ON u.ID = um.user_id ";
+			if ( ! empty( $search_key ) ) {
+				// If there's a colon in the search string, make the search smarter.
+				if( in_array( $search_key, array( 'login', 'nicename', 'email', 'url', 'display_name' ) ) ) {
```
sc0ttkclark, line 326:
```suggestion
				if( in_array( $search_key, array( 'login', 'nicename', 'email', 'url', 'display_name' ), true ) ) {
```

adminpages/memberslist-csv.php
```diff
@@ -143,8 +155,27 @@
...
+				$key_column = 'u.user_' . esc_sql( $search_key );
+				$search = " AND $key_column LIKE '%" . esc_sql( $s ) . "%' ";
+			} elseif ( $search_key === 'discount' || $search_key === 'discount_code' || $search_key === 'dc' ) {
+				$user_ids = $wpdb->get_col( "SELECT dcu.user_id FROM $wpdb->pmpro_discount_codes_uses dcu LEFT JOIN $wpdb->pmpro_discount_codes dc ON dcu.code_id = dc.id WHERE dc.code = '" . esc_sql( $s ) . "'" );
+				$search = " AND u.ID IN(" . implode( ",", $user_ids ) . ") ";
```
sc0ttkclark, line 166: I'm not too concerned but even if we think the value will be an array of numbers, we should probably cast to int anyways like `$user_ids = array_map( 'absint', $user_ids );`

As it is right now, there is no escaping happening on the user IDs before it is inserted into the query and that could become problematic later.

classes/class-pmpro-members-list-table.php
```diff
@@ -297,28 +310,44 @@ private function sql_table_data( $count = false ) {
...
+					$key_column = 'u.user_' . esc_sql( $search_key );
+					$search_query = " AND $key_column LIKE '%" . esc_sql( $s ) . "%' ";
+				} elseif ( $search_key === 'discount' || $search_key === 'discount_code' || $search_key === 'dc' ) {
+					$user_ids = $wpdb->get_col( "SELECT dcu.user_id FROM $wpdb->pmpro_discount_codes_uses dcu LEFT JOIN $wpdb->pmpro_discount_codes dc ON dcu.code_id = dc.id WHERE dc.code = '" . esc_sql( $s ) . "'" );
+					$search_query = " AND u.ID IN(" . implode( ",", $user_ids ) . ") ";
```
sc0ttkclark, line 331: I'm not too concerned but even if we think the value will be an array of numbers, we should probably cast to int anyways like `$user_ids = array_map( 'absint', $user_ids );`

As it is right now, there is no escaping happening on the user IDs before it is inserted into the query and that could become problematic later.

classes/class-pmpro-members-list-table.php
```diff
@@ -297,28 +310,44 @@ private function sql_table_data( $count = false ) {
...
+				} else {
+					$user_ids = $wpdb->get_col( "SELECT user_id FROM $wpdb->usermeta WHERE meta_key = '" . esc_sql( $search_key ) . "' AND meta_value lIKE '%" . esc_sql( $s ) . "%'" );
+					$search_query = " AND u.ID IN(" . implode( ",", $user_ids ) . ") ";
```
sc0ttkclark, line 334: I'm not too concerned but even if we think the value will be an array of numbers, we should probably cast to int anyways like `$user_ids = array_map( 'absint', $user_ids );`

As it is right now, there is no escaping happening on the user IDs before it is inserted into the query and that could become problematic later.

adminpages/memberslist-csv.php
```diff
@@ -143,8 +155,27 @@
...
+			} else {
+				$user_ids = $wpdb->get_col( "SELECT user_id FROM $wpdb->usermeta WHERE meta_key = '" . esc_sql( $search_key ) . "' AND meta_value lIKE '%" . esc_sql( $s ) . "%'" );
+				$search = " AND u.ID IN(" . implode( ",", $user_ids ) . ") ";
```
sc0ttkclark, line 169: I'm not too concerned but even if we think the value will be an array of numbers, we should probably cast to int anyways like `$user_ids = array_map( 'absint', $user_ids );`

As it is right now, there is no escaping happening on the user IDs before it is inserted into the query and that could become problematic later.

sc0ttkclark (approved):
Just some questions here on whether we should address a couple of things.

I pushed two whitespace fixes to the branch to remove some extra space at the end of the some lines.

adminpages/memberslist-csv.php
ideadude: I thought about this. In general, I'd say "why not". Might as well be extra careful and escape the var just in case, even if you think it's going to be safe.

In this case, there is a performance hit to absint a potentially large array. There's also little room for the $user_ids var to get out of line. We use it right away on the next line. We should be able to assume the response from $wpdb->get_col() there is false or an array of IDs from the DB.

I'm going to push without it for now and keep this in mind.

ideadude: Another way to put this is I don't see a vector for attack here that doesn't already give you greater access. You'd have to filter the SQL query or alter the DB, and if you could do that, you don't need to mess with this query of ours.

PR #1997, opened 2022-03-11
is_renewal() fix for cache and better support recurring memberships

mircobabini:
### All Submissions:

**Recurring orders must have their logic.**
We should query the FIRST order for a certain subscription_transaction_id and then check if the order id is the same (not renewal) or not (renewal). Otherwise it will say "renewal" even for new orders. I know that the note says "renewal means from a user who has a previous paid order" but it's completely misleading.

**This can cause cache issues: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/class.memberorder.php#L224-L227**
We don't usually recycle MemberOrder instances but it's possible that some dev does.
So the value can be already set but for a different order; we should at least unset that value at the end of `getMemberOrderByID` . [It was introduced here](https://github.com/strangerstudios/paid-memberships-pro/commit/dc2d6c45c7ecc51c978a96752c6dee7cd6453bc9). (edited)

➕ The [current is_renewal implementation](https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/class.memberorder.php#L236-L245) does not consider the membership_id, which is something to reconsider imo (removed [here](https://github.com/strangerstudios/paid-memberships-pro/commit/b141126d1d663c27e64cc064d5f290d3d9fc44c9)).

Since you might have completely different memberships (one to show videos, one to view pictures, one to get support) and you can't say "renewal" if it's not for the same membership type. Or, at least, it's not so obvious. He is a returning customer for sure, but not always a real renewal.

### Other information:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?

### Changelog entry

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -141,6 +141,9 @@
 				$this->notes = $dbobj->notes;
 				$this->checkout_id = $dbobj->checkout_id;
 
+				//reset caches
+				$this->is_renewal = null;
+				
 				//reset the gateway
 				if(empty($this->nogateway))
 					$this->setGateway();
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -235,23 +235,33 @@
 				return $this->is_renewal;
 			}
 			
-			// Check the DB.
-			$sqlQuery = "SELECT `id`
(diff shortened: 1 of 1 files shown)


mircobabini:
Closed in favour of #2049

PR #2000, opened 2022-03-12
Separate email header and footer from "email body" filters

mircobabini:
Ask @mircobabini. Related to #1907.


dparker1005:
I like the direction that this is going. It makes sense for Add Ons to filter the actual email body (minus headers and footers), and then be able to filter those sections separately. One simplification that we could consider is always running `pmpro_kses()` on the header and footer. Are there any situations where we don't want that function to run?

PR #2024, opened 2022-04-12
Load JS only in admin Paid Membership Pro Screens and Code is Poetry

joseconti:
### All Submissions:

Amin PMSP JS is loading in all WP Admin pages and is breaking other plugins and Theme JS admin.
This fixes the problem making that PMSP JS load only in its own admin pages using global $current_screen

```
if (
	'toplevel_page_pmpro-dashboard' === $current_screen ||
	'memberships_page_pmpro-memberslist' === $current_screen ||
	'memberships_page_pmpro-orders' === $current_screen ||
	'memberships_page_pmpro-reports' === $current_screen ||
	'memberships_page_pmpro-membershiplevels' === $current_screen ||
	'memberships_page_pmpro-addons' === $current_screen ||
	'memberships_page_pmpro-license' === $current_screen
) {
...
}
```

### Other information:

### Changelog entry

> Added conditional with $current_screen for load JS
> Code is Poetry

### DIFF
--- a/includes/scripts.php
+++ b/includes/scripts.php
@@ -3,89 +3,103 @@
  * Enqueue frontend JavaScript and CSS
  */
 function pmpro_enqueue_scripts() {
-    global $pmpro_pages;
-    
-    // Figure out which frontend.css file to load.    
-    if( file_exists( get_stylesheet_directory() . "/paid-memberships-pro/css/frontend.css" ) ) {
-        $frontend_css = get_stylesheet_directory_uri() . "/paid-memberships-pro/css/frontend.css";        
-    } elseif( file_exists( get_template_directory() . "/paid-memberships-pro/frontend.css" ) ) {
-        $frontend_css = get_template_directory_uri() . "/paid-memberships-pro/frontend.css";        
-    } else {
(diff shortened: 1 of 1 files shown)


ideadude:
Jose, thanks for the contribution. Sorry for the slow reply here.

In theory, I understand the desire to only load JS and CSS on our pages. In practice, our code is run on other pages, while other plugins, our own add ons, and some custom code, also expect our JS and CSS to be loaded on other pages. Also, sometimes other plugins are loading THEIR stuff on OUR pages, and we'd like to be able to fix these things on our end instead of relying on them to unload their stuff.

So updating some of our code to only load on our own screens will break as much as it fixes.

When conflicts come up, I'd rather update our CSS and JS to be more hardened and less likely to cause conflicts. For example, here is a recent commit that made sure some of our CSS that wasn't prefixed is prefixed. https://github.com/strangerstudios/paid-memberships-pro/pull/1991

It's possible the original issue you had is fixed by this same update, which went out in PMPro 2.8 I believe. If you are still having a conflict, can you let us know more details about it so we can come up with a fix that we will be able to apply without potentially breaking other stuff from turning off our code on other pages.

Thanks so much.

PR #2033, opened 2022-04-20
Add ons admin redesign

kimcoleman:
### All Submissions:

This is the start of all the cycle work for Add Ons page redesign in the admin. More notes on this are in the Trello card for the project.

### Other information:

### Changelog entry
* ENHANCEMENT: Redesigned the Memberships > Add Ons screen to provide more detail, search options, and filtering.


includes/addons.php
```diff
@@ -88,6 +95,108 @@ function pmpro_getAddonBySlug( $slug ) {
...
+		return array();
+	}
+
+	// Get all Add Ons. Return if empty.
+	$addons = pmpro_getAddons();
+	if ( empty( $addons ) ) {
+		return array();
+	}
+
+	// Note: Do we need to make sure that $license_types is an array here?
```
andrewlimaza, line 119: I think it'll be good to ensure it's an array, if not cast it to an array to prevent any warnings.

andrewlimaza:
Here's an example of code change: https://github.com/strangerstudios/paid-memberships-pro/pull/2033/commits/c2fc8de6f7fb249f5413996e88de3715d0d9c4c3

Screen Shot 2022-04-28 at 13 56 31

adminpages/addons.php
```diff
@@ -22,281 +18,394 @@
...
+					}
+
+					if ( view_val != '' ) {
+						view_items.addClass('search-hide');
+						$(`[data-search-${view}*="${view_val.toLowerCase()}"]`).removeClass('search-hide').addClass('search-show');
+					} else {
+						view_items.removeClass('search-hide').addClass('search-show');
+					}
+
+/*					var n = $( '.search-show' ).length;
```
andrewlimaza, line 368: Just some stray commented-out data here :)

includes/addons.php
kimcoleman: I removed this function now actually because it isn't in use. I saved it in a private gist if we need to bring it back.

adminpages/addons.php
kimcoleman: Removed - thank you!

PR #2034, opened 2022-04-20
MemberOrder Class Refactor

JarrydLong:
### All Submissions:

1. All existing properties are now declared as public at the top of the MemberOrder class
2. Two new static methods have been introduced: get_order and get_orders
3. New hook has been added to support order status changes pmpro_order_status_{status}

1. Possible ways of testing the new static methods

```
$order = MemberOrder::get_order( 101 );

var_dump($order);

$order = MemberOrder::get_orders( array( 'user_id' => '1' ) );

var_dump($order);
```

2. Use of the order status action. This will only run if the status has changed.

```
$morder = new MemberOrder( 101 );

$morder->status = 'success';

$morder->SaveOrder();

var_dump($morder);

function oncancel( $order ) {

	error_log('it changed on cancel');

}
add_action( 'pmpro_order_status_cancel', 'oncancel', 10, 1 );
```

### Other information:

### Changelog entry

Enhancement:  All existing properties are now declared at the top of the MemberOrder class
Enhancement: Two new static methods have been introduced: get_order and get_orders
Enhancement: New hook has been added to support order status changes pmpro_order_status_{status}

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -2,6 +2,396 @@
 	class MemberOrder
 	{	
 		/**
+		 * The Member Order ID
+		 *
+		 * @since TBD
+		 *
+		 * @var int
+		 */
+		
+		public $id = 0;
+
+		/**
+		 * The Member Order Identifier, also used as invoie number
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $code = '';
+
+		/**
+		 * User ID
+		 *
+		 * @since TBD
+		 *
+		 * @var int
+		 */
+		public $user_id = 0;
+
+		/**
+		 * Level ID
+		 *
+		 * @since TBD
+		 *
+		 * @var int
+		 */
+		public $membership_id = 0;
+
+		/**
+		 * Session ID
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $session_id = '';
+
+		/**
+		 * PayPal Token 
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $paypal_token = '';
+
+		/**
+		 * Contain a billing address object
+		 *
+		 * @since TBD
+		 *
+		 * @var object
+		 */
+		public $billing = '';
+
+		/**
+		 * Subtotal value
+		 *
+		 * @since TBD
+		 *
+		 * @var float
+		 */
+		public $subtotal = 0.00;
+
+		/**
+		 * Tax Amount
+		 *
+		 * @since TBD
+		 *
+		 * @var float
+		 */
+		public $tax = 0.00;
+
+		/**
+		 * Discount Code Amount
+		 *
+		 * @since TBD
+		 *
+		 * @var float
+		 */
+		public $couponamount = 0.00;
+
+		/**
+		 * Certificate ID
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $certificate_id = '';
+
+		/**
+		 * Total order amount
+		 *
+		 * @since TBD
+		 *
+		 * @var float
+		 */
+		public $total = 0.00;
+
+		/**
+		 * Payment Type
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $payment_type = '';
+
+		/**
+		 * Card Type
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $cardtype = '';
+
+		/**
+		 * Account or Card Number
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $accountnumber = '';
+
+		/**
+		 * Card Expiration Month
+		 *
+		 * @since TBD
+		 *
+		 * @var sring
+		 */
+		public $expirationmonth = '';
+
+		/**
+		 * Expiration Year
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $expirationyear = '';
+
+		/**
+		 * Order Status
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $status = '';
+
+		/**
+		 * Gateway ID
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		public $gateway = '';
+
+		/**
+		 * Gateway Environment
+		 *
+		 * @since TBD
+		 *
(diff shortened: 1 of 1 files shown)


classes/class.memberorder.php
```diff
@@ -10,15 +411,192 @@ function __construct($id = NULL)
...
+		 * @since TBA
+		 * 
+		 * @param mixed $id Specify an order ID or code that you want to retrieve
+		 *
+		 */
+		public static function get_order( $id = NULL ) {
+
+			return new MemberOrder( $id );
+
+		}
```
dparker1005, line 440: This function should take an array of arguments just as the `get_orders()` method does, and should leverage that function to avoid duplicate logic.

For reference, see `get_subscription()` method from our subs table work and how it "forwards" the call to `get_subscriptions()`: https://github.com/strangerstudios/paid-memberships-pro/blob/951e7008b9a69e804d9471c171bdf85b39dc72a0/classes/class-pmpro-subscription.php#L262-L299

classes/class.memberorder.php
```diff
@@ -10,15 +411,192 @@ function __construct($id = NULL)
...
+			// Filter by billing amount(s).
+			if ( isset( $args['total'] ) && null !== $args['total'] ) {
+				if ( ! is_array( $args['total'] ) ) {
+					$where[]    = 'total = %f';
+					$prepared[] = $args['total'];
+				} else {
+					$where[]  = 'total IN ( ' . implode( ', ', array_fill( 0, count( $args['total'] ), '%f' ) ) . ' )';
+					$prepared = array_merge( $prepared, $args['total'] );
+				}
+			}
```
dparker1005, line 560: Let's add `code` and `payment_transaction_id` in here, I think that both may be useful in the future

classes/class.memberorder.php
```diff
@@ -1,6 +1,407 @@
...
 	class MemberOrder
 	{	
+		/**
+		 * The Member Order ID
+		 *
+		 * @since TBD
+		 *
+		 * @var int
+		 */
+		public $id = 0;
```
eighty20results, line 11: What is the reason for using public class member variables, and not using private with __get(), __set() magic methods, and adding public get(), set() methods to allow transitioning to a more expected class behavior?

dparker1005: Hi @eighty20results, thank you for your feedback. As I mentioned in my review of this PR, transitioning to magic getters and setters is something that we are going to look into after the other notes are addressed. Having public properties seemed like the "quick solution" to stop using dynamic properties ([being deprecated in PHP 8.2](https://php.watch/rfcs/deprecate_dynamic_properties)), but it does seem like there is a cleaner way that we can do this. We will update this PR as decisions are made on this.

JarrydLong: Thanks for sharing this @eighty20results

classes/class.memberorder.php
```diff
@@ -243,945 +809,996 @@ function is_renewal() {
...
-				return false;
+	}
+
+	/*
+	* Generates a test order on the fly for orders.
+	*/
+	function get_test_order() {
+		global $current_user;
+
+		// $test_order = $this->getEmptyMemberOrder();
```
andrewlimaza, line 1769: Can remove this comment.

JarrydLong: Removed and pushed this

classes/class.memberorder.php
```diff
@@ -1,24 +1,591 @@
...
+		 * The Subscription Transaction ID
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		private $subscription_transaction_id = '';
+
+		/**
+		 * The current time the order was created
```
dparker1005, line 216:
```suggestion
		 * The time the order was created (UTC YYYY-MM-DD HH:MM:SS)
```

classes/class.memberorder.php
```diff
@@ -1,24 +1,591 @@
...
+
+		}
+
+		/**
+		 * Set Magic Method
+		 *
+		 * @since TBA
+		 * 
+		 * @param string $property The property we want to reference
+		 * @param string $value The property we want to set
```
dparker1005, line 326:
```suggestion
		 * @param string $value The value we want to set for $property
```

classes/class.memberorder.php
```diff
@@ -768,7 +1335,7 @@ function saveOrder()
 				$highestval = $wpdb->get_var("SELECT MAX(checkout_id) FROM $wpdb->pmpro_membership_orders");
 				$this->checkout_id = intval($highestval)+1;
 				$pmpro_checkout_id = $this->checkout_id;
-			}
+			}
```
dparker1005, line 1338:
```suggestion
			}
```

classes/class.memberorder.php
```diff
@@ -1184,4 +1769,68 @@ function get_test_order() {
...
+		$this->gateway_environment = 'sandbox';
+		$this->timestamp           = time();
+		$this->notes               = __( 'This is a test order used with the PMPro Email Templates addon.', 'paid-memberships-pro' );
+
+		return apply_filters( 'pmpro_test_order_data', $this );
+	}
+} // End of Class
+=======
+	} // End of Class
+>>>>>>> parent of 9b238692 (WPCS Formatting)
```
dparker1005, line 1836: Probably from bad merge.
```suggestion
} // End of Class
```

dparker1005 (changes requested):

andrewlimaza (approved):

classes/class.memberorder.php
```diff
@@ -1,24 +1,590 @@
...
+			$sql_query = "SELECT `id` FROM `$wpdb->pmpro_membership_orders`";
+
+			$prepared = array();
+			$where    = array();
+
+			$orderby  = isset( $args['orderby'] ) ? $args['orderby'] : '`timestamp` DESC';
+			$limit    = isset( $args['limit'] ) ? (int) $args['limit'] : 100;
+
+			// Detect unsupported orderby usage (in the future we may support better syntax).
+			if ( $orderby !== preg_replace( '/[^a-zA-Z0-9\s,`]/', ' ', $orderby ) ) {
```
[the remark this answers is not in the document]
ideadude: There aren't filters that could be used to add a left join. Maybe this is okay. Maybe it should be added? Later? Or a way to handle order meta in particular?

ideadude (commented):
Just some notes on whether we should support user meta queries and/or hooks to extend the queries. Going to merge as is.

PR #2038, opened 2022-04-25
Duplicate Refund Processes

JarrydLong:
### All Submissions:

- We check if an order has been refunded already so that the webook doesn't process or try and mark it as refunded again
- Correct user object sent through in admin refunded email 
- Order notes show if the order note comes from the admin dashboard or a webhook
- 

1. Refund from the orders page
2. The same refund shouldn't be processed a second time from the webhook

### Other information:


ideadude:
Great stuff. I'm also going to add some language into the IPN/webhook logs when the order was already refunded. Would be nice to see that in the log.

PR #2043, opened 2022-04-26
Splitting cancel method of PMPro_Subscription

dparker1005:
Right now, we just have a cancel() method in the PMPro_Subscription class, but there’s really two cases when you would want to call that function:
1. When you would like to cancel the subscription at the gateway
2. When you would like to mark the subscription as cancelled in PMPro

This PR separates these two cases into separate methods in an effort to make the code more intuitive

### DIFF
--- a/classes/class-pmpro-subscription.php
+++ b/classes/class-pmpro-subscription.php
@@ -1118,10 +1118,13 @@ class PMPro_Subscription {
 	 * IPNs/Webhooks should then trigger the cancellation of the subscription in the
 	 * database.
 	 *
+	 * Legacy: Falls back on calling the gateway's cancel() method if the gateway does not
+	 * support cancelling PMPro_Subscription objects specifically.
+	 *
 	 * @since TBD
 	 */
 	public function cancel_at_gateway() {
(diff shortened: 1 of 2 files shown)


classes/class-pmpro-subscription.php
```diff
@@ -1113,43 +1113,47 @@ public function save() {
...
+	 * Cancels the subscription at the payment gateway.
 	 *
-	 * @since TBD
+	 * IPNs/Webhooks should then trigger the cancellation of the subscription in the
+	 * database.
+	 *
+	 * Legacy: Falls back on calling the gateway's cancel() method if the gateway does not
+	 * support cancelling PMPro_Subscription objects specifically.
 	 *
-	 * @return bool True if the subscription was canceled successfully in the payment gateway.
```
andrewlimaza, line 1120: Would be good to keep the @return docblock and tweak the description slightly.

andrewlimaza:
Overall this looks good to me. @ideadude needs your review as well.

classes/class-pmpro-subscription.php
dparker1005: Done

PR #2046, opened 2022-04-28
Autoload ->Gateway when needed (on MemberOrder refactored)

mircobabini:
Replaces #1952. Docs in that PR.

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -272,9 +272,6 @@ class MemberOrder {
 	 */
 	function __construct( $id = null ) {
 
-		// set up the gateway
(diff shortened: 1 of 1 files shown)


dparker1005:
The `$order->Gateway` should always be set to the correct gateway at the end of the `getMemberOrderByID()` function which is run whenever an order is loaded. Is there a specific use-case/custom code that this fixes? Ideally, I think that we typically want to avoid "special cases" in the magic getter

mircobabini:
Yes. It's all described in the issue #1952

mircobabini:
As discussed with @dparker1005, the idea is to call `->setGateway()` when you want to use the `->Gateway` object. No magic tricks.

PR #2047, opened 2022-04-29
Fix timestamp is_renewal()

andrewlimaza:
* BUG FIX: Minor fix for cases where the timestamp of the is_renewal method is empty. (i.e. calling is_renewal in the pmpro_added_order hook)

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -232,6 +232,9 @@
 				return $this->is_renewal;
 			}
 			
+			if ( empty( $this->timestamp ) ) {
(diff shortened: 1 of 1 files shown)


ideadude:
I fixed this when orders are added/updated to make sure a timestamp property is set there. Still kept a kind of check in the is_renewal function, but returning false if the timestamp is empty since we can't do the test. Maybe it's better to assume the current time, but I think better to lean toward returning false here.

I also fixed a spot where the timestamp property was set incorrectly in the adminorders page.

PR #2049, opened 2022-05-02
Deprecate is_renewal + new methods is_renewal_order, is_returning_customer

mircobabini:
### All Submissions:

This PR replaces #1997 which was a bit messy.

The old is_renewal method is confusing because of its name and it's used in a confusing way in our order list.

We are showing the "renewal" label on an order (and you would expect it to be a renewal order) but the method is declared this way:

> We currently define a renewal as any order from a user who has a previous paid (non-$0) order.

**That part is so much counterintuitive.**

This PR distinguish between renewal orders and returning customers.
Also, it deprecates the older misleading method and fixes some other minors.

If we want to show a label for the returning customers too, just have to choose in which column we want it to be shown and use the `is_returning_customer` method.

### Other information:


dparker1005:
One other thing that we need to consider here is how to handle "renewal" orders for one-time payment levels. These won't have a subscription transaction ID or be a part of a payment subscription at a gateway, but I imagine that many would also consider these "renewal" payments if there was not a lapse in membership.

I think it is necessary to have a larger discussion around exactly what a renewal is in PMPro while considering different site setups before finalizing these changes.

jczelayahn:
Screenshot 2023-01-03 at 2 25 09 PM

I created a support ticket out of the renewal issue. Im my opinion if a previous order with the same subscription ID exists than it SHOULD be labeled "Renewal" but if the last previous order has a different subscription ID then it should be labeled as a "reconnection" or not labeled with anything. For our business anyone that cancels their subscription and re-signsup than they are considered a "reconnection" or "re-signup"

jczelayahn:
Screenshot 2023-01-03 at 2 25 09 PM

In my opinion if there is a previous order with a matching subscription ID than it should say "renewal", but if the subscription ID is different than the last previous it should be a "reconnect" or a "re-signup".

mircobabini:
Snippet library proposal: new column added by filter, original column hidden via css.

https://github.com/strangerstudios/pmpro-snippets-library/pull/131/commits/4d39de871f733677ea736339ddf8cfe6faca804a

PR #2050, opened 2022-05-02
Add missing i18n string context

mircobabini:
### All Submissions:

That string doesn't have context, but must be translatable.

### DIFF
--- a/classes/class-pmpro-wisdom-integration.php
+++ b/classes/class-pmpro-wisdom-integration.php
@@ -215,7 +215,7 @@ class PMPro_Wisdom_Integration {
 	 * @return string
 	 */
 	public function override_notice() {
-		return __( 'Share your usage data to help us improve Paid Memberships Pro. We use this data to analyze how our plugin is meeting your needs and identify new opportunities to help you create a thriving membership business. You can always visit the Advanced Settings and change this preference. <a href="https://www.paidmembershipspro.com/privacy-policy/usage-tracking/">Read more about what data we collect</a>.' );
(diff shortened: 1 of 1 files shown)


classes/class-pmpro-wisdom-integration.php
```diff
@@ -215,7 +215,7 @@ public function bypass_local_tracking( $is_local = false ) {
 	 * @return string
 	 */
 	public function override_notice() {
-		return __( 'Share your usage data to help us improve Paid Memberships Pro. We use this data to analyze how our plugin is meeting your needs and identify new opportunities to help you create a thriving membership business. You can always visit the Advanced Settings and change this preference. <a href="https://www.paidmembershipspro.com/privacy-policy/usage-tracking/">Read more about what data we collect</a>.' );
+		return __( 'Share your usage data to help us improve Paid Memberships Pro. We use this data to analyze how our plugin is meeting your needs and identify new opportunities to help you create a thriving membership business. You can always visit the Advanced Settings and change this preference. <a href="https://www.paidmembershipspro.com/privacy-policy/usage-tracking/">Read more about what data we collect</a>.', 'paid-memberships-pro' );
```
andrewlimaza, line 218: For strings with links, we don't want the URL to be translatable. People can put in their own links and it's easy to miss.

We'd need only the text between the `<a href` translatable.

mircobabini: We can just use the sprintf and a placeholder then.

ideadude:
Thanks for the catch. I fixed with the link this way, including targeting the link and adding nofollow, noopener. https://github.com/strangerstudios/paid-memberships-pro/commit/9a46291e2b8b9fc2493c953824343ae75cb87433

PR #2051, opened 2022-05-02
Add level metas for site health info

mircobabini:
### All Submissions:

If you want to know exactly how the levels are configured, the levels metas are needed.
It's also useful to gather informations from the addons, but this should be delegated to them.

Example:
```php
function pmprosd_site_health_info_membership_level( $membership_level ) {
	$membership_level->subscription_delay = get_option( 'pmpro_subscription_delay_' . $membership_level->id, '' );

	return $membership_level;
}

add_filter( 'pmpro_site_health_info_membership_level', 'pmprosd_site_health_info_membership_level' );
```

**Btw, if we could use level metas in subscription delay addon, to avoid the need of the code above.**

### DIFF
--- a/classes/class-pmpro-site-health.php
+++ b/classes/class-pmpro-site-health.php
@@ -128,8 +128,10 @@ class PMPro_Site_Health {
 			return __( 'No Levels Found', 'paid-memberships-pro' );
 		}
 
-		foreach ( $membership_levels as $membership_level ) {
+		foreach ( $membership_levels as &$membership_level ) {
 			$membership_level->meta = get_pmpro_membership_level_meta( $membership_level->id );
+
+			$membership_level = apply_filters( 'pmpro_site_health_info_membership_level', $membership_level );
 		}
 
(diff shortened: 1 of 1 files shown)


ideadude:
Interesting idea. We definitely need the add on settings to get the full idea of how a level is set up.

We don't really know what might be stored in the level meta. There could be tons of data in there, which would flood the report. Or there could be sensitive data in there. (To think of it, there could be sensitive data in the confirmation message.)

Maybe we want to only export certain specific data.

mircobabini:
A note from wisdom code.

```php
$level_data->set_expiration_date = get_option( 'pmprosed_' . $level_id , '' );
$level_data->subscription_delay  = get_option( 'pmpro_subscription_delay_' . $level_id , '' );
```

andrewlimaza:
I like the idea of Add Ons filtering in their own level settings to Site Health as an alternative.

For example subscription delay could hold the code to adjust the Site Health info and we include this in every Add On that we feel would be beneficial to get this info.

PR #2053, opened 2022-05-04
Added subs table compatibility for Braintree

dparker1005:
Added subs table compatibility for Braintree


ideadude:
Do we need the is_loaded check in the update subscription method too?

PR #2060, opened 2022-05-05
Fix/email confirmation message wpautop

### LINKED ISSUE #2022: "Message Confirmation" visual editor strips paragraph or break tags in Email Confirmation sent to a member
**Describe the bug**
Text added to the message confirmation area of the edit level settings page shows in the email confirmation sent to a member without the paragraphs added in visual mode.

**To Reproduce**
1. Go to the Edit Level Settings page for any membership level
2. Add two sentences with two paragraphs to the editor in "Visual Mode"
3. Save, and send a test email confirmation template or checkout on this level
4. See the sentence in the email sent to the member outputted as one paragraph instead of two

**Screenshots**
Screen Shot 2022-04-11 at 11 45 29

Screen Shot 2022-04-11 at 11 37 24

**Expected behavior**
Paragraphs should not be stripped in the email. if you switch to a text editor, however (screenshot below), then save without switching back to visual, the first paragraph is not stripped in the email.
Screen Shot 2022-04-11 at 12 31 20

Screen Shot 2022-04-11 at 12 27 36

**WordPress Environment**
<details>
- WordPress Version | 5.9.3
- Memberlite Version | 4.5.2
- PHP version | 7.4.28
- Paid Memberships Pro | Version 2.7.5
</details>

mircobabini:
### All Submissions:

Resolves #2022.

Also clean up some useless global references (thanks PHPStorm static analyzer) and declare class properties.

### Other information:


classes/class.pmproemail.php
```diff
@@ -264,13 +297,12 @@ function sendCheckoutEmail($user = NULL, $invoice = NULL)
 			$this->data = array(
 								"subject" => $this->subject, 
 								"name" => $user->display_name, 
-								"display_name" => $user->display_name,
```
dparker1005, line 267: We should probably keep this email template variable in case live sites are already using it.

classes/class.pmproemail.php
```diff
@@ -1,10 +1,43 @@
...
+		 */
+		public $subject;
+
+		/**
+		 * @var string
+		 */
+		public $template;
+
+		/**
+		 * @var string
```
dparker1005, line 30:
```suggestion
		 * @var array
```

classes/class.pmproemail.php
```diff
@@ -1,10 +1,43 @@
...
+		public $data;
+
+		/**
+		 * @var string
+		 */
+		public $body;
+
+
+		function __construct() {
 		}
```
dparker1005, line 41:
```suggestion
```

classes/class.pmproemail.php
```diff
@@ -264,13 +297,12 @@ function sendCheckoutEmail($user = NULL, $invoice = NULL)
...
 								"subject" => $this->subject, 
 								"name" => $user->display_name, 
-								"display_name" => $user->display_name,
 								"user_login" => $user->user_login,
 								"sitename" => get_option("blogname"),
 								"siteemail" => pmpro_getOption("from_email"),
 								"membership_id" => $user->membership_level->id,
 								"membership_level_name" => $user->membership_level->name,
-								"membership_level_confirmation_message" => $confirmation_message,
+								"membership_level_confirmation_message" => wpautop( $confirmation_message ),
```
dparker1005, line 305: We may instead want to use the `pmpro_confirmation_message` filter which will `wpautop()` the message and let add ons filter it. Here is how we use that filter to display the confirmation message on the confirmation page:
https://github.com/strangerstudios/paid-memberships-pro/blob/1cbc4daf1a8977d2c9d9887ec84adac83c02a0c1/pages/confirmation.php#L48-L54

```suggestion
								"membership_level_confirmation_message" => apply_filters( 'pmpro_confirmation_message', $confirmation_message, $invoice ),
```

dparker1005 (commented):
Definitely like declaring the PMProEmail variables at the top of the class! Just a couple of decisions to think about and then should be good to go

andrewlimaza (changes requested):
I agree with @dparker1005 here, we will need these changes done to merge this in.

classes/class.pmproemail.php
mircobabini: Sure thing, but it's duplicated. The display_name key is already at the end of that array of data.

mircobabini: Merged from dev resolving this redundant key https://github.com/strangerstudios/paid-memberships-pro/pull/2060/commits/c54b9ebef6fe0934cacb59156f52c9c848c6bd21

mircobabini:
Changes applied, dev merged in.

dparker1005 (approved):

classes/class.pmproemail.php
```diff
@@ -486,7 +517,7 @@ function sendCheckoutEmail($user = NULL, $invoice = NULL)
 								'siteemail' => pmpro_getOption('from_email'),
 								'membership_id' => $membership_level->id,
 								'membership_level_name' => $membership_level->name,
-								'membership_level_confirmation_message' => $confirmation_message,
+								'membership_level_confirmation_message' => apply_filters( 'pmpro_confirmation_message', $confirmation_message, $invoice ),
```
dparker1005, line 520:
```suggestion
								'membership_level_confirmation_message' => wpautop( $confirmation_message ),
```

dparker1005 (approved):
Reverting back to explicitly using `wpautop()`. Using `pmpro_confirmation_message` could cause some Add Ons to add information to emails twice, which would be a breaking change. We can consider using that filter in 3.0.

PR #2062, opened 2022-05-05
Now sending all level data to Wisdom

dparker1005:
Now sending all level data to Wisdom. 

It is important to note that the `$stats['plugin_options_fields']['pmpro_level_setups']` data may be very large as it contains the settings for all membership levels. Not sure if this will cause issues for sites with many levels.

### DIFF
--- a/classes/class-pmpro-wisdom-integration.php
+++ b/classes/class-pmpro-wisdom-integration.php
@@ -267,6 +267,7 @@ class PMPro_Wisdom_Integration {
 
 		// Levels info.
 		$levels_info = $this->get_levels_info();
+		$stats['plugin_options_fields']['pmpro_level_count']    = $levels_info['pmpro_level_count'];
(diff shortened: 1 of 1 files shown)


classes/class-pmpro-wisdom-integration.php
```diff
@@ -385,68 +388,53 @@ public function get_gateway_info() {
...
-				if ( 0 === $group['range'][0] && 0 === $group['range'][1] && 0 === $billing_amount_price ) {
-					$range_groups[ $key ]['count'] ++;
+		global $wpdb;
 
-					break;
-				}
+		$stats = array(
+			'pmpro_level_setups'   => array(),
+			'pmpro_has_free_level' => 'no',
+			'pmpro_has_paid_level' => 'yes',
```
ideadude, line 396:
```suggestion
			'pmpro_has_paid_level' => 'no',
```

PR #2065, opened 2022-05-06
Template versions

JarrydLong:
### All Submissions:

Template Versioning implemented in this PR. 

- `pmpro_loadTemplate` has been deprecated. Now uses `pmpro_load_page_template`
- Support for template loading and checks in place for Add Ons. 
- New filters: 
- - pmpro_hide_template_version_notices
- - pmpro_base_template_versions

Add Ons should now load a template using the core function 

`pmpro_load_page_template( 'directory', $path, 'pmpro-member-directory' );`

and then add in their base templates so that versions can be compared

```
function pmpromd_base_templates( $templates ) {

    $path = dirname(__FILE__);

	$directory_templates = array(
		array( $path => array( 'type' => 'templates', 'page' => 'profile' ) ),
		array( $path => array( 'type' => 'templates', 'page' => 'directory' ) )
    );

    $templates = array_merge( $templates, $directory_templates );
    
    return $templates;

}
add_filter( 'pmpro_base_template_versions', 'pmpromd_base_templates', 10, 1 );
```

Templates that are outdated are displayed in a notice

image

Steps above on changes made 

### Other information:

### Changelog entry

Introduced template versioning

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -340,26 +340,35 @@ function pmpro_isLevelExpiringSoon( &$level ) {
  *
  * @since 1.8.9
  */
-function pmpro_loadTemplate( $page_name = null, $where = 'local', $type = 'pages', $ext = 'php' ) {
+function pmpro_loadTemplate( $page_name = null, $where = 'local', $type = 'pages', $ext = 'php', $plugin_dir = PMPRO_DIR, $plugin_slug = 'paid-memberships-pro', $load = true ) {
 	// called from page handler shortcode
 	if ( is_null( $page_name ) ) {
 		global $pmpro_page_name;
 		$page_name = $pmpro_page_name;
 	}
 
+	if( $plugin_slug == 'paid-memberships-pro' ) {
+		//Core template being loaded
+		$pmpro_file_structure = 'paid-memberships-pro';
+	} else {
+		//An Add On is loading a template now
+		$pmpro_file_structure = 'paid-memberships-pro/$plugin_slug';
+	}
+	
+
 	if ( $where == 'local' ) {
 		// template paths in order of priority (array gets reversed)
 		$default_templates = array(
-			PMPRO_DIR . "/{$type}/{$page_name}.{$ext}", // default plugin path
-			get_template_directory() . "/paid-memberships-pro/{$type}/{$page_name}.{$ext}", // parent theme
-			get_stylesheet_directory() . "/paid-memberships-pro/{$type}/{$page_name}.{$ext}", // child / active theme
+			$plugin_dir . "/{$type}/{$page_name}.{$ext}", // default plugin path
+			get_template_directory() . "/{$pmpro_file_structure}/{$type}/{$page_name}.{$ext}", // parent theme
+			get_stylesheet_directory() . "/{$pmpro_file_structure}/{$type}/{$page_name}.{$ext}", // child / active theme
 		);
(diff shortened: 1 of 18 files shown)


includes/functions.php
```diff
@@ -341,6 +341,10 @@ function pmpro_isLevelExpiringSoon( &$level ) {
  * @since 1.8.9
  */
 function pmpro_loadTemplate( $page_name = null, $where = 'local', $type = 'pages', $ext = 'php' ) {
+
+	///TO DO: Version number?
+	_deprecated_function( __FUNCTION__, '3.0', 'pmpro_load_page_template()' );
```
dparker1005, line 346:
```suggestion
	_deprecated_function( __FUNCTION__, 'TBD', 'pmpro_load_page_template()' );
```

includes/functions.php
```diff
@@ -391,6 +395,247 @@ function pmpro_loadTemplate( $page_name = null, $where = 'local', $type = 'pages
...
+		if ( file_exists( $template_path ) ) {
+			//Checks and records the current template version
+			$template_version = pmpro_get_template_version( $template_path, $type );
+
+			$mmpu_present = false;
+
+			if( strpos( $template_path, 'pmpro-multiple-memberships-per-user', 0 ) !== FALSE ) {
+				/**
+				 * MMPU is loading a template.
+				 * Don't record the version because we maintain these files
```
dparker1005, line 465:
```suggestion
				 * Don't record the version because we maintain these files
				 * TODO: Remove this check when MMPU is merged into core.
```

includes/functions.php
```diff
@@ -391,6 +395,247 @@ function pmpro_loadTemplate( $page_name = null, $where = 'local', $type = 'pages
...
+			if( !empty( $used_version ) ){
+				//Only compare if we have a used version number
+				$compared = version_compare( floatval( $used_version ), floatval( $actual_version ), "<");
+
+				if( $compared ) {
+					$affected_templates[$template_slug] = array( 'our_version' => $actual_version, 'your_version' => $used_version );
+				} else {
+					$affected_templates[$template_slug] = array( 'our_version' => $actual_version, 'your_version' => __( 'Unknown', 'paid-memberships-pro' ) );
+				}
+			}
```
dparker1005, line 631: I was running into some issues testing around here. Even without using any custom page templates, I was still getting errors shown in the dashboard that I need to update. This is because we never check if `the actual version === the used version`.

I think the following logic would make sense here:
1. If there is not an $actual_version, show a warning that the custom template may be outdated and show "unknown" as the used version number
2. If $actual_version > $used_version, show a warning that they are using an outdated template and show version numbers
3. Otherwise,  template is ok

dparker1005 (changes requested):

includes/functions.php
ideadude: Could we update the options in the upgradecheck.php file? With this release, and then any time we increment a template version, we'd have to do that again.

ideadude: I'm worried about showing a warning that could confuse the site owner.

I think I would rather detect the case where there is no actual_version and just wait until the template gets loaded. I think sometimes actual_version would be the old version until the template is loaded, and in that case we wouldn't show the warning either until the template got loaded.

I would prefer to just wait until the template gets loaded to detect the issue and "queue up" the notice than show a warning.

PR #2067, opened 2022-05-06
Added function pmpro_calculate_profile_start_date()

### LINKED ISSUE #1007: BUG: ProfileStartDate and pmpro_next_payment() inconsistencies for subscriptions with multiple month billing periods.
There are some inconsistencies with the "next payment date" and subscription profile start date when a subscription cycle = multiple months. This is because the subscription profile start date is calculated assuming each month = 30 days, while pmpro_next_payment() uses strtotime() instead:

https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/gateways/class.pmprogateway_stripe.php#L1467

https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/functions.php#L613

e.g.
1 Year = 365 $trial_period_days
12 Months = 360 $trial_period_days

We could use strtotime() to calculate the $trial_period_days instead to sync dates better. This might also solve issues when users try to set expiration dates for levels with recurring subscriptions.

**Support Ticket (Moderators Only)**
https://www.paidmembershipspro.com/forums/topic/renewing-early/

dparker1005:
Added function `pmpro_calculate_profile_start_date()` to begin standardizing the process of calculating profile start dates.

Currently fully integrated for Stripe and PayPal Express. Other gateways do weird things while calculating profile start dates and probably need overhauls from the gateway side.

Importantily, this also fixes an issue where profile start dates with Stripe always consider a month to be 30 days, meaning that users may be charged days too early.

Resolves #1007


includes/gateway-helpers.php
```diff
@@ -0,0 +1,39 @@
...
+ * @param bool        $filter      Whether to filter the profile start date.
+ *
+ * @return string The profile start date in UTC time and the desired $date_format.
+ */
+function pmpro_calculate_profile_start_date( $order, $date_format, $filter = true ) {
+	// Calculate the profile start date.
+	$profile_start_date = date_i18n( 'Y-m-d H:i:s', strtotime( '+ ' . $order->BillingFrequency . ' ' . $order->BillingPeriod ) );
+
+	// Filter the profile start date if needed.
+	if ( $filter ) {
```
andrewlimaza, line 19: Not sure if we'd totally need this parameter here. Do we have plans to ever not filter the profile start date when calling `pmpro_calculate_profile_start_date`?

dparker1005: I'm currently hoping to use that param to determine if we can bundle initial/recurring payments when sending users to Stripe to check out:
https://github.com/strangerstudios/paid-memberships-pro/blob/415acf9cb02015d5310f9fa68d4ef2d4990b070d/classes/gateways/class.pmprogateway_stripe.php#L1657-L1669

The idea is that if the filtered profile start date is the same as the unfiltered profile start date, then no add on is changing the date of the first recurring payment so the subscription can be started immediately with the initial payment as a part of the subscription itself.

If there is another way to figure this out, though, then I would be ok removing that parameter.

andrewlimaza: Ah got it, overlooked the $unfiltered_trial_period_days. 👍

andrewlimaza:
Thanks, looks good. @ideadude need your once over.

ideadude:
Great work.

We have been discussing in person turning the pmpro_calculate_profile_start_date() function into a method on the order class. We are only calling this function ourselves in contexts where we have an order object.

In some ways this is putting a checkout/gateway related method into the order class, which is an ongoing problem. We will eventually want to refactor things so there is a PMProCheckout class or equivalents, and may have to move the method at that time. But we can cross that bridge when we get there.

dparker1005:
@ideadude Renamed gateway-helpers.php to checkout.php as per our discussion yesterday

PR #2069, opened 2022-05-09
Health: catch addons installed in wrong folder

mircobabini:
### All Submissions:

Sometimes you download an addon from github and you don't remove the _-branch_ suffix.

This is from a conversation with Kim White; sometimes you are investigating a weird situation on a customer's site and you don't see that there's an addon installed with that wrong folder name, which prevents to be shown into PMPro > Add Ons > Installed list. Also prevents it to be updated.

Thanks to this PR, you will immediately see it into Site Health. Can be a starting point to show a warning into PMPro > Add Ons.
Screenshot 2022-05-09 at 16 49 09

1. Download a copy of an addon from github and install
2. Check Site Health > Info > PMPro section

### Other information:


dparker1005 (approved):

includes/addons.php
```diff
@@ -71,6 +71,45 @@ function pmpro_getAddons() {
 	return $addons;
 }
 
+/**
+ * Search installed addons with wrong folder name.
+ *
+ * @since TBD
+ */
+function pmpro_searchWrongFolderNameAddons() {
```
ideadude, line 79: Is this the best name for this function.

We've been inconsistent on this in the past, but basically any new function we had should use snake case, e.g. pmpro_search_wrong_folder_name_addons().

But then I also think pmpro_find_wrong_folder_name_addons() might be more descriptive? Is there something even better? What other contexts are we going to use this? We only use this once, so we're abstracting a little earlier, but I think it's okay here. Let's just imagine how it might be used going forward and have the function name and doc block make the most sense.

Thanks!

PR #2076, opened 2022-05-11
We were incorrectly calculating the TrialBillingFrequency in pmpro_pmpro_subscribe_order_startdate_limit filter.

ideadude:
### All Submissions:

We have a filter pmpro_pmpro_subscribe_order_startdate_limit() in filters.php that is meant to detect cases where we are going to set the subscription start date to > 1 year in the future. PayPal can't handle this, so we try to "squeeze" the extra time into a free trial if possible.

However, that was meant to figure out the number of days for the trial would always be set to 365 instead of having that as an upper limit.

This update correctly calculates the days past 1 year and only if it's > 1, sets the trial up.

This issue might have come up randomly for 1 year recurring levels or more consistently if you had custom code or level settings that would result in subscriptions with renewals > 1 year.

1. Create a recurring level for $5 every 370 days.
2. Checkout with PayPal Express.
3. Note that the subscription in PayPal is set up with a 5 (or 4 if leap year) day free trial.

### Other information:

### Changelog entry

> BUG FIX: Fixed issue where in rare instances PayPal Express subscriptions had an extra 1 year $0 trial added to them.

### DIFF
--- a/includes/filters.php
+++ b/includes/filters.php
@@ -194,13 +194,14 @@ function pmpro_pmpro_subscribe_order_startdate_limit( $order, $gateway ) {
 		$one_year_out = strtotime( '+1 Year', current_time( 'timestamp' ) );
 		$two_years_out = strtotime( '+2 Year', current_time( 'timestamp' ) );
 		$one_year_out_date = date_i18n( 'Y-m-d', $one_year_out ) . 'T0:0:0';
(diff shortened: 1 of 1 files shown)


dparker1005 (approved):
Seems like a good fix for this issue, though it feels like this "hack" should be a part of the PayPal/PayPal Express classes as it is specific to those gateways. But since this code has already been out there in its current form, seems reasonable just to fix it.

PR #2085, opened 2022-05-17
Showing warning when user leaves user fields settings without saving

dparker1005:
Could not completely test as there were some issues with the current User Fields branch. Going to look into those and then test afterwards.


ideadude:
I lied. I was testing wrong. Merging in now.

PR #2094, opened 2022-05-18
Now showing description for field groups in admin profile

dparker1005:


ideadude:
Merging, but should probably escape. There are other places we need to escape field output too.

PR #2113, opened 2022-05-24
Now creating new Stripe product if saved ID doesn't exist

dparker1005:
There are a couple cases in support where users switch Stripe accounts and are then seeing the error `No such product: prod_......` at checkout. This fixes that issue by checking that a product exists before attempting to create a charge for it and, if the product does not exist, creating a new one.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -2691,7 +2691,7 @@ class PMProGateway_stripe extends PMProGateway {
 		}
 
 		// Return the product ID.
-		return empty( $stripe_product_id ) ? null : $stripe_product_id;
+		return ! empty( $stripe_product_id ) ? $stripe_product_id : null;
 	}
 
 	/**


classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -2648,30 +2648,50 @@ private function get_subscription( $subscription_id ) {
...
+		// If a valid product does not exist for this level, create one.
 		if ( empty( $stripe_product_id ) ) {
 			$stripe_product_id = $this->create_product_for_level( $level, $gateway_environment );
 		}
 
-		if ( ! empty( $stripe_product_id ) ) {
-			return $stripe_product_id;
-		}
+		// Return the product ID.
+		return empty( $stripe_product_id ) ? null : $stripe_product_id;
```
andrewlimaza, line 2694: Looking at this it may read better for `return ! empty( $stripe_product_id ) ? $stripe_product_id : null` following the example of the WPCS -> https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#ternary-operator

andrewlimaza, line 2694: Looking at this it may read better for `return ! empty( $stripe_product_id ) ? $stripe_product_id : null` following the example of the WPCS -> https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#ternary-operator

PR #2123, opened 2022-06-02
Showing update notices when update is available for PMPro

dparker1005:
Screen Shot 2022-06-02 at 11 27 11 AM

Useful for future PMPro updates where we want to warn users about breaking changes.

I have not been able to test this completely since this would require having an update for PMPro that includes an update notice; however, the code does work when setting PMPro to an old version and adding the following line of code at the top of the `pmpro_maybe_show_upgrade_notices()` function:
`$update_data->upgrade_notice = 'Your update notice...';`

Based off of this article: https://andidittrich.com/2015/05/howto-upgrade-notice-for-wordpress-plugins.html

### DIFF
--- a/css/admin.css
+++ b/css/admin.css
@@ -1039,3 +1039,15 @@ button.pmpro_report_th_closed:before {
 		width: 0;
 	}
 }
+
+/* plugin update notice */
+.pmpro_plugin_update_notice {
+	background-color: #d54e21 !important;
+	padding: 10px !important;
+	color: #f9f9f9 !important;
+	margin-top: 10px !important;
+}
+
+.pmpro_plugin_update_notice + p:before {
+	display: none;
+}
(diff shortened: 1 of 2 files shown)


andrewlimaza:
Thanks for this. I think we need the following (so we can start leveraging the upgrade notice in the readme beyond the V2 -> V3 update).

* If updating from V2 -> V3.x show a static message regardless of the current upgrade notice. People don't upgrade immediately to major versions, some may wait until v3.2 (If we add an upgrade notice to V3.2 update it won't warn people about the V3 update and backwards compatibility concerns we have).

* Keeping this logic in allows us to show security notices more prominently with future releases, i.e. V3.2 has an upgrade notice of "This includes an important security fix".

ideadude:
added an extra esc_html around the strip_tags to be safe.

PR #2128, opened 2022-06-08
User-fields-duplicate

dparker1005:


andrewlimaza:
Works great! A nice to have would be to automatically add a number to the end of meta_key when duplicating (just in case someone doesn't rename this and have a duplicate field name).

PR #2137, opened 2022-06-15
Reduce calls to get_option() by nearly 40%

JJJ:

This commit simplifies the logic inside of `pmpro_getOption()` by only calling `get_option()` a single time and passing `''` as the default value.

The benefit of this change is that `get_option()` contains a bunch of preflight checks and a few filters, and there is no longer any benefit to running all of them two times per call to `pmpro_getOption()` now that it doesn't do anything other than prefix the option key.

### Other information:

### Changelog entry

Simplify `pmpro_getOption()` as a wrapper for `get_option()` using `''` as the default value


JJJ:
Relatedly, there are 79 places where `get_option()` is called where `pmpro_getOption()` would potentially fit, _except_ that `pmpro_getOption()` does not have a second parameter for a custom default value (and `$force` isn't used).

The other 380'ish calls to `pmpro_getOption()` appear to always expect a string, so they seem fine to return `''`.

`pmpro_getOption()` is now just a wrapper for `get_option()` so y'all may want to deprecate it or improve it to normalize all of the calls.

Tangentially, PMP doesn't use `register_setting()` so there are other areas of improvement, too! 🤣

ideadude:
Hey, thanks for this work. Overhauling/fixing how we store options is on target for us later this year, maybe in a 3.0 release (so we can break some things) or earlier. I appreciate this PR. We will review it when we can.

PR #2139, opened 2022-06-22
Testing PHP Sessions in Site Health

### LINKED ISSUE #2117: Check if PHP Sessions are enabled in Site Health.
Issues arise from PHP Sessions not being enabled correctly on hosts and creates support.

The solution here would be to test PHP Session if it's enabled or not and display it in the Site Health debug information we collect. We should also check if the WP Native PHP Sessions plugin is installed - https://wordpress.org/plugins/wp-native-php-sessions/

JarrydLong:
### All Submissions:

We now test to see if Sessions are available and return this data in Site Health

Resolves #2117 

1. Navigate to Tools -> Site Health -> Paid Memberships Pro
2. Session data is present 

image

### Other information:

### Changelog entry

We now test PHP sessions in our Site Health data.

### DIFF
--- a/classes/class-pmpro-site-health.php
+++ b/classes/class-pmpro-site-health.php
@@ -81,6 +81,10 @@ class PMPro_Site_Health {
 					'label' => __( 'Discount Codes', 'paid-memberships-pro' ),
 					'value' => self::get_discount_codes(),
 				],
+				'pmpro-sessions'       => [
+					'label' => __( 'PHP Sessions', 'paid-memberships-pro' ),
+					'value' => self::test_sessions(),
+				],
 				'pmpro-membership-levels'    => [
 					'label' => __( 'Membership Levels', 'paid-memberships-pro' ),
 					'value' => self::get_levels(),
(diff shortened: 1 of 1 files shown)


andrewlimaza:
Reworked the logic slightly to make sure that PHP Sessions aren't entirely disabled. It may help catch the obvious cases, I've seen before that sites may have "PHP_SESSION_NONE" status in some cases which can cause conflicts.

This should be treated lightly and if we suspect issues that are often caused by SESSION conflicts to still test this. Just because SiteHealth says it's okay might be a false positive if they're doing "weird" things on the frontend - but still helps identify the obvious issues here.

ideadude:
I'm going to merge this in.

If there are further session issues not caught by this check, we can do a "real" check. Something like:

1. When the site health page is loaded, store a value into `$_SESSION['pmpro_test_session'] = 'itsworking';`
2. Then on the site health page, load JS to make an AJAX call to a callback that just looks for that `$_SESSION['pmpro_test_session']` var and outputs it. The JS then checks if it is 'itsworking'.
3. If the session is not available during the callback, set an option or transient.
4. Check the option or transient and raise an issue on the site health page RE sessions not working. Let's have a page to link to.
5. Maybe via JS reload the site health page so the new issue is shown (or can we squeeze it in via JS)

PR #2163, opened 2022-07-13
Now checking if actually on a checkout form before preventing default

dparker1005:
If pmpro-stripe.js is loaded on a page with a non-checkout PMPro form, it would prevent submit buttons from working. This could often be the case when using this code recipe which sets up a checkout form on every page:
https://www.paidmembershipspro.com/process-membership-checkout-in-a-modal-or-popover-window/

This change tries to minimize this issue by only preventing default if a checkout form was submitted. The criteria for checking this is currently if there is a "level" input in the form, but we can revise this later if necessary if there are additional cases where issues come up.


ideadude:
Merged. We should remember somehow that we need to change this when/if we change the level parameter name in PMPro 3.0.

PR #2197, opened 2022-08-18
Now creating user before payment at checkout

dparker1005:
### All Submissions:

PMPro currently waits until after a payment is successful to create a WP user, but this can result in checkout issues that are often plugin conflicts that are difficult to debug (see [troubleshooting article here](https://www.paidmembershipspro.com/troubleshooting-guide-users-not-created-at-checkout/)). Furthermore, not having a WP user set up causes orders to be created in `token` status without an associated user when an offsite gateway, which can be confusing to customers.

Instead, this PR moves the user creation step to before the payment collection step to avoid these kinds of issues entirely.

Main downside is that some websites don't want to have users created unless they have actually paid, so we need to decide if it is more important to keep things as-is for those users or to avoid checkout issues for others.

### Other information:

### Changelog entry

### DIFF
--- a/preheaders/checkout.php
+++ b/preheaders/checkout.php
@@ -435,6 +435,78 @@ if ( $submit && $pmpro_msgt != "pmpro_error" ) {
 				}
 			}
 
+			// Do we need to create a user account?
+			if ( ! $current_user->ID ) {
+				// Yes, create user.
+
+				//first name
+				if ( ! empty( $_REQUEST['first_name'] ) ) {
+					$first_name = $_REQUEST['first_name'];
+				} else {
+					$first_name = $bfirstname;
+				}
+				//last name
+				if ( ! empty( $_REQUEST['last_name'] ) ) {
+					$last_name = $_REQUEST['last_name'];
+				} else {
+					$last_name = $blastname;
+				}
+
+				//insert user
+				$new_user_array = apply_filters( 'pmpro_checkout_new_user_array', array(
+						"user_login" => $username,
+						"user_pass"  => $password,
+						"user_email" => $bemail,
+						"first_name" => $first_name,
+						"last_name"  => $last_name
+					)
+				);
+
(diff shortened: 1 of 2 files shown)


preheaders/checkout.php
```diff
@@ -435,6 +435,78 @@
 				}
 			}
 
+			// Do we need to create a user account?
+			if ( ! $current_user->ID ) {
+				// Yes, create user.
+
+				//first name
+				if ( ! empty( $_REQUEST['first_name'] ) ) {
+					$first_name = $_REQUEST['first_name'];
```
ideadude, line 444:
```suggestion
					$first_name = sanitize_text_field( $_REQUEST['first_name'] );

```

preheaders/checkout.php
```diff
@@ -435,6 +435,78 @@
...
+
+				//first name
+				if ( ! empty( $_REQUEST['first_name'] ) ) {
+					$first_name = $_REQUEST['first_name'];
+				} else {
+					$first_name = $bfirstname;
+				}
+				//last name
+				if ( ! empty( $_REQUEST['last_name'] ) ) {
+					$last_name = $_REQUEST['last_name'];
```
ideadude, line 450:
```suggestion
					$last_name = sanitize_text_field( $_REQUEST['last_name'] );

```

ideadude (commented):
Still need to test with all gateways, and I guess decide we really want to make this change with no setting/etc to put it back the old way.

But wanted to get a couple of sanitization changes in there. Rest looks good.

ideadude:
Nevermind the comment RE having a setting to disable. The future of the Internet and PMPro is creating users before checkout. (Multistep checkout will require this. And there are tons of reasons to want to have the user account first, and it helps tremendously to KNOW that that is absolutely the case. I.e. getting rid of the "checkout out but not yet a user" case simplifies things.)

PR #2199, opened 2022-08-18
Removing redundant settings of $current_user->membership_level

dparker1005:
### All Submissions:

In July 2014, this code was added to every preheader to allow PMPro pages to retrieve a user's membership level using the global `$current_user->membership_level`:
https://github.com/strangerstudios/paid-memberships-pro/commit/11f0fa711d07ff88cce1499ea7e36a0cef9d6326

In August 2014, this process was abstracted to instead set `$current_user->membership_level` on every page load:
https://github.com/strangerstudios/paid-memberships-pro/commit/b33c0fdaa303d6ae28de82c3a21385a04a03b1e3

Since that property is now set on every page load, it should no longer be necessary to set it in preheaders.

### Other information:

### Changelog entry

### DIFF
--- a/preheaders/account.php
+++ b/preheaders/account.php
@@ -10,11 +10,6 @@ if ( ! is_user_logged_in() ) {
 	}
 }
 
-// Make sure the membership level is set for the user.
-if( $current_user->ID ) {
-    $current_user->membership_level = pmpro_getMembershipLevelForUser( $current_user->ID );
-}
-
 // Process the msg param.
 if ( isset($_REQUEST['msg'] ) ) {
     if ( $_REQUEST['msg'] == 1 ) {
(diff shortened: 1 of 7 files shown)


ideadude:
Nice catch. I think we need to test this because for some reason there were cases where the $current_user global wasn't going through that set_current_user hook, or maybe it was running too early to really set up the level correctly. I just vaguely remember bugs where the ->membership_level wasn't set when I would expect it to be. I'll try to remember when those cases were.

ideadude:
I think we move this to 3.0 to since it could cause minor backwards compat issues. We can set the membership_level on the current_user object, but in all of our code, let's just use $membership_levels = pmpro_getMembershipLevelsForUser(); (or the single version of it) whenever we want to work with the membership level.

Adding params to WP core objects should be avoided, even if those objects aren't "FINAL" in the code.

PR #2210, opened 2022-08-25
PR1 - Adds checkboxes to the delete user page

JarrydLong:
### All Submissions:

Adds options to the delete user page, allowing you to choose if you want to remove active subscriptions, and delete member history for that user. 

1. Make sure a user has an active subscription
2. Navigate to Users and click delete on that user
3. You'll be given the option to cancel their active subscription and membership history

### Other information:

### Changelog entry

You can now choose if you want to remove an active subscription and delete membership history when deleting a user.


includes/cleanup.php
```diff
@@ -31,19 +43,35 @@ function pmpro_delete_user_form_notice( $current_user, $userids ) {
...
+	<div class='pmpro_delete_user_actions'>
+		<p><?php esc_html_e( 'What should be done with the PMPro Membership data for these users?', 'paid-memberships-pro' ); ?></p>
+	<?php
 	// Show a notice if users for deletion have an an active membership level.
 	if ( ! empty( $userids_have_levels ) ) { ?>
 		<div class="notice notice-error inline">
 			<?php
 			if ( count( $userids ) > 1 ) {
-				_e( '<p><strong>Warning:</strong> One or more users for deletion have an active membership level. Deleting a user will also cancel their membership and recurring subscription.</p>', 'paid-memberships-pro' );
+				_e( '<p><strong>Warning:</strong> One or more users for deletion have an active membership level.</p>', 'paid-memberships-pro' );
```
dparker1005, line 55:
```suggestion
				echo  '<p><strong>' . esc_html__( 'Warning', 'paid-memberships-pro' ) . ':</strong>' . esc_html__( 'One or more users for deletion have an active membership level.', 'paid-memberships-pro' ) . '</p>';
```

includes/cleanup.php
```diff
@@ -31,19 +43,35 @@ function pmpro_delete_user_form_notice( $current_user, $userids ) {
...
 			} else {
-				_e( '<p><strong>Warning:</strong> This user has an active membership level. Deleting a user will also cancel their membership and recurring subscription.</p>', 'paid-memberships-pro' );
+				_e( '<p><strong>Warning:</strong> This user has an active membership level.</p>', 'paid-memberships-pro' );
```
dparker1005, line 57:
```suggestion
				echo  '<p><strong>' . esc_html__( 'Warning', 'paid-memberships-pro' ) . ':</strong>' . esc_html__( 'This user has an active membership level.', 'paid-memberships-pro' ) . '</p>';
```

dparker1005 (changes requested):

includes/cleanup.php
```diff
@@ -31,19 +43,43 @@ function pmpro_delete_user_form_notice( $current_user, $userids ) {
 		}
 	}
 
+	$member_history = $wpdb->get_var( "SELECT COUNT(*) as members FROM $wpdb->pmpro_memberships_users WHERE user_id IN (" . implode( "," , $userids ) . ")" );
```
ideadude, line 46: Let's use $wpdb->prepare here.

ideadude: Create a var like userids_string and pass that as the param to prepare. Because user IDs are ints (in MySQL) we shouldn't need quote around the string. Just the commas.

includes/cleanup.php
```diff
@@ -31,19 +43,43 @@ function pmpro_delete_user_form_notice( $current_user, $userids ) {
...
+	<div class='pmpro_delete_user_actions'>
+		<p><?php esc_html_e( 'What should be done with the PMPro Membership data for these users?', 'paid-memberships-pro' ); ?></p>
+	<?php
 	// Show a notice if users for deletion have an an active membership level.
 	if ( ! empty( $userids_have_levels ) ) { ?>
 		<div class="notice notice-error inline">
 			<?php
 			if ( count( $userids ) > 1 ) {
-				_e( '<p><strong>Warning:</strong> One or more users for deletion have an active membership level. Deleting a user will also cancel their membership and recurring subscription.</p>', 'paid-memberships-pro' );
+				echo  '<p><strong>' . esc_html__( 'Warning', 'paid-memberships-pro' ) . ': </strong>' . esc_html__( 'One or more users for deletion have an active membership level.', 'paid-memberships-pro' ) . '</p>';
```
ideadude, line 63: Need to put the : inside the translated string so RTL translations move it correctly.

includes/cleanup.php
```diff
@@ -31,19 +43,43 @@ function pmpro_delete_user_form_notice( $current_user, $userids ) {
...
 			} else {
-				_e( '<p><strong>Warning:</strong> This user has an active membership level. Deleting a user will also cancel their membership and recurring subscription.</p>', 'paid-memberships-pro' );
+				echo  '<p><strong>' . esc_html__( 'Warning', 'paid-memberships-pro' ) . ': </strong>' . esc_html__( 'This user has an active membership level.', 'paid-memberships-pro' ) . '</p>';
```
ideadude, line 65: Need to put the : inside the translated string so RTL translations move it correctly.

includes/cleanup.php
ideadude: More complicated then because it's 2 translated strings and the second needs to be moved with the translation too.

So to keep the strong tags in there use wp_kses with strong as an allowed html. Then something like:

`'<p>' . wp_kses( __( '<strong>Warning:</string> One or more users for deletion have an active membership level.', 'paid-memberships-pro' ), $allowed_html ) . '</p>' );`

includes/cleanup.php
```diff
@@ -68,3 +104,19 @@ function pmpro_delete_post( $post_id = null ) {
...
 add_action( 'delete_post', 'pmpro_delete_post' );
+
+/**
+ * Delete all membership data for a specific user from the membership users table.
+ *
+ * @param int $user_id The WordPress user ID.
+ * @since TBD
+ */
+function pmpro_delete_membership_history( $user_id ) {
+	global $wpdb;
```
ideadude, line 115: We should bail here (return, maybe return false) if $user_id is empty. Otherwise, I think our query here deletes all user history.

ideadude (commented):
A couple escaping updates and some things to make sure our functions don't accidentally delete more info that we want them to.

includes/cleanup.php
```diff
@@ -5,13 +5,22 @@
 // deleting a user? remove their account info.
 function pmpro_delete_user( $user_id = null ) {
```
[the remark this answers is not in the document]
andrewlimaza: Added return false here, as well as the function below to keep uniformity.

includes/cleanup.php
andrewlimaza: Added return false here as well to match the above check as well. That way if the function returns false we know the user ID is empty.

andrewlimaza (approved):

includes/cleanup.php
```diff
@@ -31,19 +47,47 @@ function pmpro_delete_user_form_notice( $current_user, $userids ) {
 		}
 	}
 
+	$prepare_member_history = $wpdb->prepare( "SELECT COUNT(*) as members FROM $wpdb->pmpro_memberships_users WHERE user_id IN (%s)", implode( "," , $userids ) );
+
+	$member_history = $wpdb->get_var( $prepare_member_history );
```
ideadude, line 52:
```suggestion
	$sqlQuery = $wpdb->prepare( "SELECT COUNT(*) as members FROM $wpdb->pmpro_memberships_users WHERE user_id IN (%s)", implode( "," , $userids ) );

	$member_history = $wpdb->get_var( $sqlQuery );
```

PR #2212, opened 2022-08-25
Removing unnecessary code. Level in $invoice

dparker1005:
The checkout and billing failure emails have been updated to pull the membership_id from the passed $invoice, so we no longer need to set $user->membership_id (which was not MMPU compatible)


ideadude:
I'm a little worried add ons or other custom code might rely on $current_user->membership_level set up here, so just pushing this off to 2.10 so it's included in a minor release.

ideadude:
Hmmm. Related to https://github.com/strangerstudios/paid-memberships-pro/pull/2199 which I think should be a 3.0 update now. So maybe push this off until then too?

andrewlimaza:
This looks good and glad we're being extra cautious. I have updated the base branch to v3.0 👍

PR #2223, opened 2022-08-30
PR3 - Pause Mode Functionality

JarrydLong:
### All Submissions:

Moving your site to a staging site can result in unexpected changes such as crons running and live actions occurring. This will check if the site URL has changed and disable the PMPro crons. 

1. Install PMPro - check if the pmpro_site_url option has been created in the db
2. Change the current site URL or copy the site to another domain
3. A notice will appear notifying you of the change, which then puts the site in pause mode.

### Other information:

### Changelog entry

Core features can now temporarily be turned off or paused when setting up a staging site.

### DIFF
--- a/includes/admin.php
+++ b/includes/admin.php
@@ -97,9 +97,17 @@ function pmpro_site_url_check() {
 	}
 
 	//The WP_ENVIRONMENT_TYPE has been changed, we should pause everything
-	if( ! pmpro_is_production_site() ) {
+	//But only if we're not forcing pause mode to be turned off
+	//local forces the WP_ENVIRONMENT_TYPE to be set to local
+	if( ! pmpro_is_production_site() && ! pmpro_getOption( 'pause_mode_override' ) ) {
 		//Site URL's don't match - enable pause mode
 		pmpro_setOption( 'pause_mode', true );
+	}
+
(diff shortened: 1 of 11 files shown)


includes/admin.php
```diff
@@ -82,4 +82,90 @@ function pmpro_init_site_health_integration() {
...
+			//Site URL's do match - disable pause mode
+			pmpro_setOption( 'pause_mode', false );				
+		}
+	} else {
+		//We are paused, show a notice.
+		add_action( 'admin_notices', 'pmpro_pause_mode_notice' );
+	}
+
+}
+add_action( 'init', 'pmpro_site_url_check' );
```
ideadude, line 120:
```suggestion
```
Let's just do admin_init

dparker1005:
How would this PR function if a staging site was updated to 2.10 (or whatever version) so that the staging site was set as the `last_known_url`, and then that staging site was pushed to a live site? Could this create situations where live sites are  being treated as a staging site instead of the other way around?

It seems like it could be a good idea to, in the admin notice, say something like "It seems like this site may be a staging site. As a precaution, some PMPro features have been disabled. If this is incorrect, please click here to confirm that this site is your live site." And instead of just setting the override option, we should update the `last_known_url` option in that case.

dparker1005 (approved):

PR #2252, opened 2022-09-28
Fixing col name for membership_id in get_orders()

dparker1005:
### All Submissions:

In the database, orders have a `membership_id` column, but the `get_orders()` method tries to reference a non-existent `membership_level_id` column.

### Other information:

### Changelog entry


JarrydLong (approved):

dparker1005:
> Nice catch. Should we change the arg you would pass to "membership_id" to match that?

This change would make sense, but there may already be code out there using `membership_level_id`. We would probably want to account for both if we decided to make that change

ideadude:
I'll merge and in another commit update the param with support for the old version too.

PR #2258, opened 2022-10-20
Increases Stripe Application Fees

JarrydLong:
### All Submissions:

- New installation will make use of the 2% Stripe fee
- Existing installations that update to the latest version will continue using the 1% Stripe fee

1. New Installation: Install and activate the plugin. Run a test purchase through Stripe Connect and check the fees. 2% should be charged.
2. Existing Installation: Updae the plugin so that the DB version is updated. Run a test purchase through Stripe Connect and check the fees. 1% should be charged.

### Other information:

### Changelog entry

Stripe application fee changes


classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -3956,7 +3956,18 @@ public static function get_application_fee_percentage() {
 		if ( self::using_legacy_keys() ) {
 			return 0;
 		}
-		$application_fee_percentage = pmpro_license_isValid( null, pmpro_license_get_premium_types() ) ? 0 : 1;
+
+		//These are percentage values 0%, 1% and 2%
+		
+		$stripe_app_fee = pmpro_getOption( 'stripe_app_fee' );
+		if ( $stripe_app_fee === FALSE ) {
```
dparker1005, line 3963:
```suggestion
		if ( ! empty( $stripe_app_fee ) ) {
```
We should use the saved application fee if there is one saved in options.

dparker1005 (changes requested):

dparker1005:
Just a discussion, right now the code sets the user's fee to 1% if the site existed before the 2.10 update. This means that sites will get the reduced rate even if:
- They weren't previously using Stripe
- They were previously using legacy keys and therefore not paying the old Connect fee
- They previously had a PMPro license key to avoid the fee, but got rid of it after updating to 2.10

Should we try to limit the sites that get 1% to sites that were already collecting payments at that rate?

dparker1005 (approved):

mircobabini:
I've just read the code and it seemed a bit complex imho. A bit _spaghetti_.

I'd say we should:
1. Use the `function pmpro_upgrade_1()` to set the application fee to 2% (default behaviour for fresh installs)
1. Use the 2.10 update condition to lower it to 1% under certain conditions (`!$fresh_install`, ecc...)
1. Reduce [this code](https://github.com/strangerstudios/paid-memberships-pro/blob/724c1e0e10a76d860cabc69748c14d75fe21ccbc/classes/gateways/class.pmprogateway_stripe.php#L3962-L3967) to a simpler intval(). That method should be atomic, knowing nothing about the business logics.

ideadude:
> Should we try to limit the sites that get 1% to sites that were already collecting payments at that rate?

Good point. I think we can safely treat users with no Stripe orders and no Stripe settings as "new" here and use the 2% fee.

I'm also going to process Mirco's suggestions a bit. There were a few ways we could have gone about this. We also have to support folks who are/were filtering the app fee already and also the countries that don't allow fees.

ideadude:
Check for an existing stripe connect api key, and if found, keep it at 1% for them.

@dparker1005 if Mirco's suggestion makes things better or simpler, I trust your judgement.

dparker1005:
Closing in favor of #2342 which checks if a user has live connect credentials instead of if this is a new site.

PR #2281, opened 2022-11-14
Modularizing async checkout handling

dparker1005:
### All Submissions:

Added functions for saving checkout data into order meta before sending a user offsite to pay and for completing checkouts asynchronously during webhook/IPN handlers.

This approach was tested during the Stripe Checkout beta period. We should begin to use it for all other offsite gateways so that they all work the same way and so that we don't need to worry about server session settings.

### Other information:

### Changelog entry


dparker1005:
Closing in favor of #2616

PR #2291, opened 2022-11-30
Fix warning on user edit.

andrewlimaza:
* Remove `enddate = ''` check in SQL when expiration date is set to NO to resolve PHP warning on profile save.


andrewlimaza:
This would need further testing, running the SQL directly in the DB throws an error for '0000-00-00 00:00:00' but doesn't throw an error in PHP's side.

We can also set this to be: `enddate < '1970-01-01 00:00:00'` if it's still an issue. I don't think this would cause any issues on inserting into a database on the edit user profile after digging deeper into this.

PR #2314, opened 2022-12-28
Convert arrays to comma separated strings.

ideadude:
### All Submissions:

1. Create a user field of type multiselect or select2 with a few options.
2. Edit your own user profile to choose a couple of these values.
3. Insert the shortcode [pmpro_member field="myfieldkey"]
4. Note that before this fix, the output will be "Array" with a notice in the error log. After, you get a comma separated string of the values in the array.

### Other information:

Note here that in my testing, a user field with values like "1:One;2:Two;3:Three" seemed to store the user meta value as One/etc instead of 1/etc. Let's make sure this is saving as intended... and also if we change it, it supports some kind of conversion. (sigh) And then if user fields are stored with keys instead of values, let's see if we can look up the values in the user meta field and output those instead of the keys.

Consider handling this issue before merging and releasing this. But this PR as is would at least be a move in the right direction.

### Changelog entry

> BUG FIX/ENHANCEMENT: Now handling multiselect, select2, and other user field types that are stored in user meta as an array.

### DIFF
--- a/shortcodes/pmpro_member.php
+++ b/shortcodes/pmpro_member.php
@@ -130,6 +130,11 @@ function pmpro_member_shortcode($atts, $content=null, $code='')
 		else
 			$r = pmpro_escape_price( pmpro_formatPrice($r) );
 	}
+	
+	//Check for arrays to reformat them.
+	if ( is_array( $r ) ) {
+		$r = implode( ', ', $r );
+	}
 
 	/** 
 	 * Filter


ideadude:
> a user field with values like "1:One;2:Two;3:Three" seemed to store the user meta value as One/etc instead of 1/etc.
Okay, this is expected because of how we test if the array of values is "associative" (i.e. all numerical keys in order) here: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/class-pmpro-field.php#L1340-L1343

So I was trying to test a case where values were stored instead of labels, but set up a bad test case. We'd need something like 101:One;202:Two;303:Three to test that case.

If we set it up that way, the output here would be like "101, 202" instead of "One, Two". So we should update this PR to look up the field and values and swap keys for labels when outputting.

And we shouldn't have to worry about any kind of script/code to convert how old user meta was stored in arrays. It is working as expected (despite the fact that the is_assoc check there isn't really expected sometimes... but let's look into that separately, since it's been working this way for a while).

ideadude:
Going to merge this now. A follow up PR will add the code to look up the labels from the values when needed.

PR #2350, opened 2023-01-31
Fix double-prefix when using pmpro_setOption in wisdom integration

mircobabini:
### All Submissions:

When enabling wisdom integration from the Allow button on a back-end notice, the Advanced Settings option was still set to no. This is because we were saving the user choice into a double prefixed option.

### Other information:

### DIFF
--- a/classes/class-pmpro-wisdom-integration.php
+++ b/classes/class-pmpro-wisdom-integration.php
@@ -34,7 +34,7 @@ class PMPro_Wisdom_Integration {
 	 *
 	 * @var string
 	 */
-	public $plugin_option = 'pmpro_wisdom_opt_out';
+	public $plugin_option = 'wisdom_opt_out';
 
 	/**
 	 * The plugin settings pages to include Wisdom notices on.


dparker1005 (approved):

ideadude:
This var is used in a few other places in that class.

https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/class-pmpro-wisdom-integration.php#L95-L117

If it's not pmpro_ prefixed, it could clash with another plugin using this library. And the hooks to sync on option update won't run.

So the better fix I think is to change the pmpro_setOption call to just update_option. Or perhaps unprefix the var when making the setOption call, but I think that's confusing and not needed.

In general, it's nice to use pmpro_setOption in case we bundle up options in the future and need to change how things are called/etc, but here we can call update_option directly.

ideadude:
Going to close in favor of this: https://github.com/strangerstudios/paid-memberships-pro/commit/0823ac6b5ac7597da9d907e3a64de9dc61f52c2b

Also going to look into overwriting pmpro_wisdom_opt_out with pmpro_pmpro_wisdom_opt_out on upgrade so we start getting data from folks who opted in. I am pretty sure that anyone who clicked "yes" meant to allow tracking. I suppose they may have noticed the advanced option and left it off. Need to think about it.

mircobabini:
@ideadude we can bug them once if pmpro_pmpro_wisdom_opt_out is set to yes. "Do you confirm you still want to permit us tracking bla bla bla?" etc. Same notice as before, changing just the title, like we are asking them if they're still in. Then, based on the answer, delete the extra option and store the new answer.

PR #2370, opened 2023-02-08
Discount Code List Table

JarrydLong:
### All Submissions:

Replaces the current Discount Code table with a WP List Table 

1. Navigate to Memberships > Settings > Discount Codes
2. The new table displays along with the ability to sort

### Other information:

* [] Have you added an explanation of what your changes do and why you'd like us to include them?
* [] Have you successfully run tests with your changes locally?

### Changelog entry

Replaces the current discount code table with a WP List Table


adminpages/discountcodes.php
```diff
@@ -732,7 +739,7 @@
 			</form>
 		</div>
 
-	<?php } else { ?>
+	<?php } else { //here ?>
```
dparker1005, line 742:
```suggestion
	<?php } else { ?>
```

classes/class-pmpro-discount-code-list-table.php
```diff
@@ -0,0 +1,516 @@
...
+	 * @return mixed
+	 */
+	public function column_uses( $item ) {
+		
+		global $wpdb;
+		
+		
+
+		if( $item['uses'] > 0 ) {
+			echo "<strong>" . (int)$item['used'] . "</strong>/" . $item->uses;
```
dparker1005, line 447:
```suggestion
			echo "<strong>" . (int)$item['used'] . "</strong>/" . (int)$item['uses'];
```
Fixing PHP warning for uses > 0

classes/class-pmpro-discount-code-list-table.php
```diff
@@ -0,0 +1,516 @@
...
+		}
+
+		if ( ! empty( $s ) ) {
+			$sqlQuery .= "WHERE code LIKE '%" . esc_sql( $s ) . "%' ";
+		}
+
+		if ( ! $count ) {
+
+			if( isset( $_REQUEST['orderby'] ) ) {
+				$orderby = sanitize_text_field( $_REQUEST['orderby'] );
```
dparker1005, line 206: Just taking a note to start a discussion. When sorting by `uses`, the sort goes by the number of total uses, not the number of times that the code has already been used.

The weird thing here is that when a code has 0 uses, it is considered unlimited. This means that when sorting from lowest uses to highest, `unlimited` will be first.

dparker1005 (changes requested):

classes/class-pmpro-discount-code-list-table.php
JarrydLong: This wasn't working properly, it should have been sorting by used instead. Pushed a commit further down that fixes this

dparker1005: Just tested this PR again on my dev site and it still appears to be sorting based on `uses` instead of `used`. Can you check if you are seeing the same behavior on your site?

dparker1005 (approved):
Looks good!

PR #2391, opened 2023-02-21
Changes how we load reports

JarrydLong:
### All Submissions:

Changes how we include report files and selectively loads them instead of just loading everything in the reports folder

1. Navigate to Memberships > Reports
2. No errors should be thrown

### Other information:

### Changelog entry

Changes how we load reports

### DIFF
--- a/includes/reports.php
+++ b/includes/reports.php
@@ -1,17 +1,5 @@
 <?php
 /*
-	Load All Reports
-*/
-$pmpro_reports_dir = dirname(__FILE__) . "/../adminpages/reports/";
-$cwd = getcwd();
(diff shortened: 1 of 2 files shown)


ideadude:
Just noting here that technically adding these includes AFTER we load includes/reports.php, which loads extra reports from inside the active theme, will change the load order so that the built in reports run AFTER the ones in the theme. If the reports code in the theme is coded well, this shouldn't change anything other than possibly the default order things show up on the reports dashboard. I think reports in themes is a small enough use case that it's okay to do this. It feels right to load includes/reports.php first which could in the future have other code supporting reports.

PR #2410, opened 2023-03-16
Trims spacing on dropdown values to ensure consistent results

JarrydLong:
### All Submissions:

Trims spacing around a dropdown's values so that checks validate when the value is selected

1. Create a dropdown field in User fields
2. set the values to have a space after each of them (normal whitespace)
3. Edit the profile using the fronend profile edit page. Choose a new value from the dropdown and save. It saves correctly as before - but it should show the correct/selected field from the dropdown.

### Other information:

### Changelog entry

Trims whitespace from dropdown values

### DIFF
--- a/classes/class-pmpro-field.php
+++ b/classes/class-pmpro-field.php
@@ -727,20 +727,13 @@ class PMPro_Field {
 			$r .= ">\n";
 			foreach($this->options as $ovalue => $option)
 			{
-
-				if( is_array( $value ) ) {
-					$selected = ( trim( $ovalue ) == array_walk( $value, 'trim' ) );
-				} else {
-					$selected = ( trim( $ovalue ) == trim( $value ) );
-				}	
-
 				$r .= '<option value="' . 
 					
 					
 					trim( $ovalue ) . '" ';
 				if(!empty($this->multiple) && in_array($ovalue, $value))
 					$r .= 'selected="selected" ';
-				elseif( $selected )
+				elseif ( ! empty( $ovalue ) && is_string( $value ) && trim( $ovalue ) == trim( $value ) )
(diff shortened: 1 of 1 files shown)


ideadude:
Good idea. Is there any chance this breaks things for users who chose values on old fields with the extra space? Can we test that?

1. Use the non-patched version of PMPro.
2. Create a select field with 2 options, values "a " and "b " (note spaces).
3. Have the user select option "b " and save from the edit user or profile page.
4. Apply the patch. Visit the edit user or profile page.
5. Is "b " still selected for them?

We could potentially trim things going into the `selected()` calls and/or update some other code to make sure old values get checked correctly.

I think it's possible that " " at the front or back is important. Choose your separator: , or / or | or " " LOL But I think that should be a rare enough edge case to ignore. However if a site was set up and recorded everyone's city as "New York City ", after this update, I'm worried that data appears to be lost to admins and users.

JarrydLong:
Just making a note here that trimming only really happens on the selected function to show if selected, so if you have 1000 users with the space it 'should' fix itself the next time you view/edit your profile and then save

ideadude:
What Jarryd says makes sense. We still need to test. After testing, let's merge.

dparker1005 (changes requested):
Just tested and the change works nicely for `select` field. In the current dev branch, if a user chose an option with whitespace, it would save in the database without the white space and fields would not default to their selected option when displayed again. After this PR, those selections are recovered nicely and show what the user selected, even if the original selection was before this PR was implemented.

I am requesting changes because other fields still run into the same problem for options with whitespace. Specifically, I tested with select2. I assume that the fix is similar, but it should be tested with all field types that use options.

PR #2411, opened 2023-03-16
Fixes a fatal error on payment settings page with Spanish lang

JarrydLong:
### All Submissions:

The original code made use of esc_html__ inside a sprintf. This works fine in English but fails with other languages, resulting in a fatal error. 

Putting the sprintf inside the esc_html__ instead (swapping them around) fixes the issue.

1. Switch the default language to Spanish
2. Ensure no license key is set or activate on a clean install
3. Go to Memberships > Settings > Payment Gateway & SSL
4. The page should load without issues. 

### Other information:

### Changelog entry

Fixes a fatal error on the Payment Gateway & SSL settings page.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -2447,7 +2447,7 @@ class PMProGateway_stripe extends PMProGateway {
 						} else {
 							$application_fee_percentage = self::get_application_fee_percentage();
 							if ( ! empty( $application_fee_percentage ) ) {
-								echo sprintf( esc_html__( 'Note: You are using the free Stripe payment gateway integration. This includes an additional %s fee for payment processing. This fee is removed by activating a premium PMPro license.', 'paid-memberships-pro' ), intval( $application_fee_percentage ) . '%' );
+								echo esc_html__( sprintf( 'Note: You are using the free Stripe payment gateway integration. This includes an additional %s fee for payment processing. This fee is removed by activating a premium PMPro license.', intval( $application_fee_percentage ) . '%', 'paid-memberships-pro' ) );
 							} else {
 								esc_html_e( 'Note: You are using the free Stripe payment gateway integration. There is no additional fee for payment processing above what Stripe charges.', 'paid-memberships-pro' );
 							}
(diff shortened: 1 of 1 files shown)


dparker1005 (approved):

ideadude:
I couldn't reproduce, but maybe I needed to use a specific Espanol locale... or maybe the translation has since been fixed or this was due to a custom translation.

It seems the root cause here might be sprintf throwing a fatal error when there is no placeholder in the translated string? Can we test that assumption?

Is there a fix for that in particular. I don't think we want to reverse the order of these function calls. The string with the %s should get localized and then the fee swapped into that place. Otherwise, the localization needs a translation for all variations of this string and they wouldn't be prompted for that.

Is there some way to avoid sprintf throwing an error here without writing a bunch of code to test for the placeholder (that would be overkill IMO). Maybe using the %1$d type placeholders is handled better? (I haven't tested.)

We want to fix this in the translations and also come up with a solution to the core problem, which may show up elsewhere.

JarrydLong:
@ideadude I've changed the order back to what it was and just added a positioning value to the %s so it becomes %1s which fixes it. 

I suspect that the 2*%* part is causing an issue somewhere along the line - just weird that it only happens with some languages and not all.

ideadude:
Closing this.

The latest commit is wrong. It should use `%1$s` (with a dollar sign) instead of just `%1s`. 

It also doesn't fix the issue. The reason it seems to is because changing this literal string means the broken translation isn't used.

The real fix was to fix the translations. [Here is one.](https://translate.wordpress.org/projects/wp-plugins/paid-memberships-pro/stable/es/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=12330517&filters%5Btranslation_id%5D=103393367)

I don't think there is a reasonable way to fix this. We have to trust our translations to have the correct number, type, and positioning of placeholders.

Some folks will suggest always using the numbers arguments in the placeholders. I am not sure the reasoning, but I think we shouldn't needlessly break our translations to introduce the number argument if it doesn't actually do anything.

Some reading RE sprintf, etc.

https://wordpress.stackexchange.com/questions/294327/what-does-the-token-1s-in-wordpress-represent

https://www.php.net/manual/en/function.sprintf.php

PR #2415, opened 2023-03-17
Marking Register Helper as deprecated.

dparker1005:
### All Submissions:

Marking Register Helper Add On as deprecated. Fixes issues when both RH and PMPro 2.9+ are active.

### Other information:

### Changelog entry

### DIFF
--- a/includes/deprecated.php
+++ b/includes/deprecated.php
@@ -161,7 +161,7 @@ function pmpro_check_for_deprecated_add_ons() {
 			'file' => 'pmpro-register-helper.php',
 			'label' => 'Register Helper',
 			// TODO: Update link to Gist to a real blog post.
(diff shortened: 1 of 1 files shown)


ideadude:
Before merging and releasing this, we want a gist or plugin replacement for the limit by email/username options. I'd be okay with a one file gist that includes a plugin header vs a true repository... unless we think we could potentially extend this functionality in the future to a full on add on.

Do we need to notify people RE the change somehow other than changelogs? We can detect that there is data in the option used (find the meta keys) and show a dismissible notice linking to the new plugin.

dparker1005:
Created the gist here: https://gist.github.com/dparker1005/7c2e90ed4d29f6923734bad6e53a2856

We can add a notice to the `deprecations` file in PMPro checking for the metadata and show a notice if they need to add to their site?

dparker1005:
Most recent commit now shows a deprecation message when RH is on the site, but not enabled. The deprecation message currently reads:
**Register Helper**: The "allow specific emails and usernames" feature was not merged into the core PMPro plugin. If this is a feature that was being used on this website, it can be added again using this code recipe: https://gist.github.com/dparker1005/7c2e90ed4d29f6923734bad6e53a2856

The message is currently showing whether or not the feature was previously being used. This seems ok since the site should be deleting RH immediately anyways.

We should write a blog post for that gist and update the link in this code.

PR #2417, opened 2023-03-21
Memberlist Export Joindate produces same result as table

JarrydLong:
### All Submissions:

In the export we reference the joindate `array_push($csvoutput, pmpro_enclose(date_i18n($dateformat, $theuser->user_registered)));` While in the table we reference `user_registered` 

I've tested an export with referencing $theuser->user_registered and we get the same results that the user sees in the table.

This issue came up on a customer's site where I suspect they had some users imported. 

The start date showed fine in the table but gave a 1970 date when exporting. 

1. Export a member list as normal from the Members tab. 
2. As mentioned above though, this issue was experienced on a customer site with the possibility of users who were imported - making this change just produced the same results in the export as one sees in the member list table.

### Other information:

### Changelog entry

Member list export join date now references the user_registration date in an export

### DIFF
--- a/adminpages/memberslist-csv.php
+++ b/adminpages/memberslist-csv.php
@@ -98,7 +98,7 @@
 		$headers[] = 'Content-Disposition: attachment; filename="members_list.csv"';
 
 	//set default CSV file headers, using comma as delimiter
-	$csv_file_header = "id,username,firstname,lastname,email,billing firstname,billing lastname,address1,address2,city,state,zipcode,country,phone,membership,initial payment,fee,term,discount_code_id,discount_code,joined";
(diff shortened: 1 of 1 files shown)


ideadude:
Maybe the heading is wrong instead?

dparker1005:
After digging into this more, the members list has both "registered" and "start date", whereas the CSV export only has "joined".

We should update the CSV to have both "registered" and "start_date" to mirror the members list.

dparker1005 (changes requested):
After digging into this more, the members list has both "registered" and "start date", whereas the CSV export only has "joined".

We should update the CSV to have both "registered" and "start_date" to mirror the members list.

JarrydLong:
We haven't been able to replicate the original issue that spurred this PR on - closing this for now - if it becomes an issue in the future we can relook at it.

PR #2419, opened 2023-03-24
Sanitize Getfile URL using sanitize_url instead of sanitize_text_field

JarrydLong:
### All Submissions:

Sanitizes the getfile URL by using sanitize_url instead of sanitize_text_field as this function strips too much out of the URL, resulting in the URL breaking

1. Set up getfile to protect files on the site
2. Reference a file that contains dashes, underscores and other characers (even spaces)
3. The file should load/redirect without breaking. 

### Other information:

### Changelog entry

Changed how we sanitize the getfile URL

### DIFF
--- a/services/getfile.php
+++ b/services/getfile.php
@@ -25,7 +25,7 @@
 	global $wpdb;
 
 	// Get the file path.
-	$uri = sanitize_text_field( $_SERVER['REQUEST_URI'] );
+	$uri = sanitize_url( $_SERVER['REQUEST_URI'] );
 
 	// Remove the query string from the path.
 	$uri_parts = explode( '?', $uri );


kimcoleman:
I looped Jason in on this one to understand why we used sanitize_text_field instead of sanitize_url or esc_url.

This is all some feedback from him on the decision to keep sanitize_text_field:

--  - - - -- - - - -- 

When we did our [major escaping work earlier this year](https://github.com/strangerstudios/paid-memberships-pro/commit/515743db8c3c22d425ca7e4e647fee776d9fedac), the `sanitize_url` function is relatively new to WP, so we didn’t want to use that. It was removed back in 2.8 and added back in version 5.9.

sanitize_url is a wrapper for esc_url. I think the WPCS scans that .org was doing were upset RE this being an “escaping” function... which is probably related to why the sanitize_url function was added back.

In reality here, the sanitize_text_field function does the same kinds of protections, might even protect things better than esc_url here.

Note that we also sanitize part of the URI here: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/services/getfile.php#L127-L129

--  - - - -- - - - -- 

I think for now we should consider to have a discussion about this all and figure out how to update. We could probably wrap [this whole thing in esc_url before redirecting](https://github.com/strangerstudios/paid-memberships-pro/blob/dev/services/getfile.php#L127-L129). One protection of esc_url we don’t have here but want is that it double checks the protocol. The wp_safe_redirect use here makes sure that it still redirects to the same host/domain.

PR #2425, opened 2023-03-31
Prevent multiple clicks on cancel page

dparker1005:
### All Submissions:

Now disabling links on cancel page once one is clicked.

### Other information:

### Changelog entry

### DIFF
--- a/js/pmpro-cancel.js
+++ b/js/pmpro-cancel.js
@@ -1,0 +1,7 @@
+jQuery( document ).ready( function() { 
+    // Prevent links from being clicked multiple times on PMPro cancel page.
+    jQuery( '#pmpro_cancel .pmpro_btn-submit' ).click( function() {
+        jQuery( '#pmpro_cancel a' ).unbind( 'click' );
+        jQuery( '#pmpro_cancel a' ).attr( 'disabled', 'disabled' );
+    });
+});
--- a/preheaders/cancel.php
+++ b/preheaders/cancel.php
@@ -89,3 +89,11 @@
 			$pmpro_msgt = "pmpro_error";
 		}
 	}
+
+	wp_register_script(
+		'pmpro_cancel',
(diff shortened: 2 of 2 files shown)


MaximilianoRicoTabo (commented):
I think the real problem here is that we're sending a GET request over the browser to cancel a membership instead of using a DELETE or PUT request.  That means a couple of  different problems. 

- A conceptual and semantic issue behind misuse of links. Links aren't meant to be used as buttons. Links should bring the user to new content. A section, a  new page, and external site. 
-  An accessibility issue with links
- A bad practice behind performing a delete or soft delete or update through a HTTP  GET request by hitting a page with JS.

We could cancel the membership either with  AJAX (i'd do this due simplicity)  and show the same message on the success function callback. Or even submit an appropriate request, catch it in the server and redirect to the success page.

Said this above, the PR does what it says, it works at local, prevent clicking the link twice and i guess the issues related to that scenario. Yet it seems to me that is treating the symptom instead of atack the real issue.

ideadude:
We will merge now, and take these notes to potentially update our cancel process in 3.0 when the CONPD add on is merged in. Specifically let's consider (1) rewriting to vanilla JS, (2) using a form here so we can do a POST request instead of a GET (I think PUT and DELETE are still not supported unless we are managing the whole interaction via JS which doesn't feel right for a WP site yet) and (2) potentially adding in some other params to the request so the CONPD code knows if the user intended to cancel the sub or the full membership so it doesn't do both back to back by accident.

mircobabini:
Jumping in just bc I worked on this some time ago (had to git blame!) here: https://github.com/strangerstudios/paid-memberships-pro/commit/9d9597ac50aebf2e5ed6fd8aa262fcc8f53c37c5

Maybe we also want to move/remove the inline `onlick` javascript.

PR #2431, opened 2023-04-04
Escaping string values in pmpro_membershiplevels_page_action_links filter

dparker1005:
### All Submissions:

Escaping string values in pmpro_membershiplevels_page_action_links filter and updating inline docs.

We could potentially use a function other than `esc_html()` to have more control over how the string is escaped, but as this case should never come up if the filter is being used correctly, escaping aggressively seems like a reasonable solution. In cases where this is over-escaping (ex escaping `<a href=...`), the code adding the link should be updated to use the filter correctly with the expected array input.

### Other information:

### Changelog entry

### DIFF
--- a/adminpages/membershiplevels.php
+++ b/adminpages/membershiplevels.php
@@ -1069,6 +1069,8 @@
 			/**
 			 * Filter the Membership Levels page title action links.
 			 *
+			 * @since 2.9
+			 *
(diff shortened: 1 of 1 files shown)


ideadude:
I don't know how widespread it was to use this filter without passing arrays. But if so, it was most likely to add links, so we should either (1) at least allow links in the output or (2) just remove the feature all together... skip the value if it's not an array.

PR #2435, opened 2023-04-05
Deleting spam users who do not complete checkout

dparker1005:
### All Submissions:

Deleting users who are registered with PMPro but do not complete the checkout process. This is helpful to delete spam accounts that fail the payment step and then bail.

As currently coded, this feature is powered by an hourly cron that deletes spam accounts 24 hours after creation. There is currently a filter to change that window of time and another filter to disable user deletion altogether.

Before releasing, we may want to make UI settings for this feature and we need to figure out how it should be set by default.

### Other information:

### Changelog entry

### DIFF
--- a/includes/crons.php
+++ b/includes/crons.php
@@ -35,6 +35,9 @@ function pmpro_get_crons() {
 		'pmpro_license_check_key'                  => [
 			'interval' => 'monthly',
 		],
+		'pmpro_cron_spam_pending'                  => [
+			'interval' => 'hourly',
+		],
 	];
 
 	/**
--- a/includes/spam.php
+++ b/includes/spam.php
@@ -190,3 +190,23 @@ function pmpro_disable_checkout_for_spammers( $required_fields ) {
 	return $required_fields;
 }
 add_filter( 'pmpro_required_billing_fields', 'pmpro_disable_checkout_for_spammers' );
+
+/**
+ * Set a user meta value when a user is created during a PMPro checkout.
+ */
+function pmpro_set_spam_pending_user_meta( $user_id ) {
+	// Set a user meta value with the current timestamp.
+	update_user_meta( $user_id, 'pmpro_spam_pending', current_time( 'timestamp' ) );
+}
+add_action( 'pmpro_checkout_before_user_auth', 'pmpro_set_spam_pending_user_meta', 10 );
+
+/**
+ * After a checkout is complete, remove the pmpro_pending user meta value.
+ *
+ * @param int $user_id The user ID.
+ */
+function pmproabandoned_remove_spam_pending_user_meta( $user_id ) {
+	// Remove the user meta value.
+	delete_user_meta( $user_id, 'pmpro_spam_pending' );
+}
+add_action( 'pmpro_after_checkout', 'pmproabandoned_remove_spam_pending_user_meta', 10 );
--- a/scheduled/crons.php
+++ b/scheduled/crons.php
@@ -311,3 +311,32 @@ function pmpro_cron_admin_activity_email() {
 		$pmproemail->sendAdminActivity();
 	}
 }
+
+add_action( 'pmpro_cron_spam_pending', 'pmpro_cron_spam_pending' );
+function pmpro_cron_spam_pending() {
+	// Determine how many seconds need to pass for a checkout to be considered abandoned.
+	$seconds_until_abandoned = apply_filters( 'pmpro_spam_pending_seconds_until_deletion', 24 * HOUR_IN_SECONDS );
+
+	// Get all user IDs that have been pending for more than $hours hours.
+	$user_ids = get_users(
+		array(
+			'meta_key'     => 'pmpro_spam_pending',
+			'meta_value'   => current_time( 'timestamp' ) - $seconds_until_abandoned,
+			'meta_compare' => '<',
+			'fields'       => 'ID',
+		)
+	);
+
+	// Make sure that we can delete users.
+	if ( ! function_exists( 'wp_delete_user' ) ) {
+		require_once( ABSPATH . 'wp-admin/includes/user.php' );
+	}
+
+	// Loop through the users and delete them.
+	foreach ( $user_ids as $user_id ) {
+		// Add a filter to give custom code a chance to bail.
+		if ( apply_filters( 'pmpro_spam_pending_delete_user', true, $user_id ) ) {
+			wp_delete_user( $user_id );
+		}
+	}
+}


dparker1005:
There is the `pmpro_spam_pending_delete_user`, but I think that you're right that either a filter to turn off the cron or maybe an advanced setting would be nice. Should consider this before merging. We also need to choose the default time to wait before deleting spam users. Should probably be 2 weeks or so to account for Stripe delayed payment methods

ideadude:
Good start. I like the general idea of tagging users at checkout and then removing that tag when checkout is complete. Some notes from our discussion to use when refactoring this:

* Let's use a user taxonomy instead of user meta for performance reasons
* We can maybe use some of the user fields code for the above
* Let's expand on the situations where we will remove the user taxonomy, e.g. creating a post, updating their profile, WC purchase, EDD purchase, others?? so we don't delete users who interacted with non-PMPro stuff after getting logged in by PMPro
* Add an "incomplete membership checkouts" submenu to the users page (that's a mouthful).
* clicking on it adds a param to the URL to filter the WP list table user query by our user taxonomy
* Now folks can manually check users and bulk delete them
* We will develop a paid add on to automate some of this (there are other types of old inactive users we'd want to include in  the users the add on addresses)

dparker1005:
Just finished most of those bullets. The remaining "to do's" are:

- Expanding on situations where we will remove the taxonomy
- Finalizing name for "incomplete membership checkouts" menu item

One other thing on my mind is that when admins filter the users list by these "incomplete" users, there is no visual indication that this is being done (just shows the normal "Users" page UI) and there isn't a way to get a sense of how long ago the user was created. This is problematic because if a user is mid-checkout or if they used a payment method that takes multiple days to complete, the admin does not have a way of knowing this.

ideadude:
btw for this I mean a link at the top of the users table... not a link in the sidebar menu.
> Add an "incomplete membership checkouts" submenu to the users page (that's a mouthful).

dparker1005:
Just moved the "Incomplete Membership Checkokuts" option next to the list of roles and added a sortable "User Registered" column when that the "Incomplete Membership Checkouts" option is selected. Should be ready for final review now.

includes/abandoned-signups.php
```diff
@@ -0,0 +1,196 @@
+<?php
+
+/**
+ * Set up a user taxonomy to track users who were created during the PMPro
+ * checkout process but did not complete the checkout.
+ *
+ * @since TBD
+ */
+function pmpro_abandoned_signups_taxonomy() {
```
ideadude, line 9: I'm thinking it's good to get picky about the naming here. Our first priority is to flag users that never complete checkout so we can delete them after a few days, but in reality this taxonomy can be used later for abandoned cart functionality, which would be really cool.

ideadude: So I'm thinking something more like "pmpro_checkout_user_status" and having terms like "pending", "abandoned", and "completed".

Pending = user created at checkout
Abandoned = maybe updated via cron X days later
Completed = updated when checkout is completed

includes/abandoned-signups.php
```diff
@@ -0,0 +1,196 @@
...
+ * @param int $user_id The user ID.
+ */
+function pmpro_set_abandoned_signup_taxonomy( $user_id ) {
+	// Bail if the user ID is empty.
+	if ( empty( $user_id ) ) {
+		return;
+	}
+
+	// Add the abandoned signup taxonomy to the user.
+	var_dump( wp_set_object_terms( $user_id, 'abandoned-signup', 'pmpro_abandoned_signup' ) );
```
ideadude, line 39: Note the var_dump here.

includes/abandoned-signups.php
```diff
@@ -0,0 +1,196 @@
...
+function pmpro_remove_abandoned_signup_taxonomy( $user_id ) {
+	// Bail if the user ID is empty.
+	if ( empty( $user_id ) ) {
+		return;
+	}
+
+	// Remove the abandoned signup taxonomy from the user.
+	wp_remove_object_terms( $user_id, 'abandoned-signup', 'pmpro_abandoned_signup' );
+}
+add_action( 'pmpro_after_checkout', 'pmpro_remove_abandoned_signup_taxonomy' );
```
ideadude, line 57: I think we need to think of some other actions that would complete here. So e.g. something on login (but make sure to skip the login we do at checkout) would be good, because if a user logged in user that user/pass a day or two later that's an indication that they are expecting to continue using that user account.

Similarly, we need to check for posts added with them as the author (could be literal posts or maybe show that they interacted with some plugin via CPT)... or maybe specifically check for WooCommerce checkouts, EDD checkouts, etc. What other actions would indicate "this user expects to keep using this user account" besides completing a PMPro checkout?

includes/abandoned-signups.php
```diff
@@ -0,0 +1,196 @@
...
+add_action( 'deleted_user', 'pmpro_remove_abandoned_signup_taxonomy' );
+
+/**
+ * Filter the views on the Users page to include a view for abandoned signups.
+ *
+ * @since TBD
+ *
+ * @param array $views The views on the Users page.
+ */
+function pmpro_add_users_table_view_abandoned_signups( $views ) {
```
ideadude, line 67: If we have 3 statuses now, we probably want to at least allow filtering by pending as well. I don't think filtering by completed is as useful, but maybe people would expect it.

ideadude (commented):
Updated with notes RE changing the taxonomy name to "pmpro_checkout_user_status" and using 3 terms (pending, abandoned, completed).

Also an errant var_dump in there.

includes/abandoned-signups.php
ideadude: Let's merge pending and abandoned into an "incomplete" status. This avoids the need for a cron, but doesn't have the loaded "abandoned" language that might be problematic if it "gets out" in future updates.

ideadude: Let's merge pending and abandoned into an "incomplete" status. This avoids the need for a cron, but doesn't have the loaded "abandoned" language that might be problematic if it "gets out" in future updates.

includes/abandoned-signups.php
ideadude: Let's focus on the original target issue here of spam checkouts and just folks who never come back after failing checkout. What we can do is hook into the "wp" hook and if not pmpro_is_checkout() then consider this a user who was logged in and did *something* other than try to checkout with PMPro and remove the inactive/abandoned/spam checkout taxonomy term (or whatever we end up calling it).

So we'll hook into the pmpro_after_checkout and user_deleted hooks, but add a new callback hooked on wp to check if the user has the term and remove it.

This will then run on wp for every logged in user, but should be a quick enough check that it doesn't slow things down.

includes/abandoned-signups.php
ideadude: No longer relevant if we go back to 1 or 2 statuses.

dparker1005:
Made all the requested changes. The one thing that we still need to think about before 2.11 releases is the name of the taxonomy in the code to potentially remove the word "abandoned".

PR #2448, opened 2023-04-24
Accessibility and usability improvements based on the UX testing we contracted

kimcoleman:
### All Submissions:

Lots of updates to wording, accessibility, and more throughout the plugin (especially the Wizard) based on the feedback from Piccia's usability review.

More to come but this is a big update already.

### Other information:


kimcoleman:
Added lots more work to this one - the latest improvements were updates following the Accessibility Audit by Alex Stine and Amber Hinds: https://equalizedigital.com/paid-memberships-pro-accessibility-audit/

There is an internal Google Doc with more items that require discussion. But this is alot of progress so far. Some discussions remains surrounding:

- role="alert" for messages in the admin settings area. Since we kind of have a hacky message output system that is either contextual to the action OR general setup progress notices that we keep active, this fix isn't as straightforward as it could. Let's future plan to redo the notifications for save separate of the "system status" and "set up status" and tackle that later."

- my Macbook screen reader is pronouncing the word "signups" strangely (like sig-nups"). If we hyphenate the term (like sign-ups), it fixes. That's not a decision I feel like I can make on my own everywhere. 

- There are fixes needed around the level template modal "Add New Level". We need to make the "page action link" a button. The action / button needs to shift context into the modal. Some inspiration perhaps: https://uxdesign.cc/how-to-trap-focus-inside-modal-to-make-it-ada-compliant-6a50f9a70700 maybe
https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element:represents-2

- Overall, we need the h1 on pages with menus to be before the menus. But this is all a part of some admin area revamp work we can tackle at that time.

- We need a separate PR for html5 "required" fields for our frontend forms. We will not add it to required fields in user fields when on the profile page. Only on checkout. This can probably be accomplished just with JS.

- placeholder variables for email templates and on the edit level > confirmation text area. Screen reader isn't include the !! or underscores when reading. I haven't found a suitable fix to force this to read fully as displayed.

PR #2453, opened 2023-04-26
Issue 1819 Email Header variable returning incorrect user name for ad…

### LINKED ISSUE #1819: Email Header variable returning incorrect user name for admins.
**Describe the bug**
The !!name!! variable in the email header is returning incorrect results to admins. It returns the user's name for admin emails.

**To Reproduce**
1. Go to edit a user's profile page on WordPress.
2. Change the user's level.
3. Look at email sent to admin, should have Dear <<user>>, instead of Dear <<admin>>

**Screenshots**
Admin email received, showing the user's name instead of the admin name.

Screen Shot 2021-10-13 at 14 35 43

**Expected behavior**
This should change the !!name!! variable to be the admin's name, i.e. Dear Admin, (or user the email is addressed to).

MaximilianoRicoTabo:
…mins.

 * Add a function to get admin name by email
 * Add that name to the sendAdminChangeEmail function array data.
 * Add a new 'related_user_name' field to have both. admin name to be notified an user being modified
 * Change template accordingly
 
 
image

### All Submissions:

* [ x] Have you followed the [Contributing guideline](https://github.com/strangerstudios/paid-memberships-pro/blob/master/.github/CONTRIBUTING.md)?
* [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/strangerstudios/paid-memberships-pro/pulls) for the same update/change?

 * Add a function to get admin name by email
 * Add that name to the sendAdminChangeEmail function array data.
 * Add a new 'related_user_name' field to have both. admin name to be notified an user being modified
 * Change template accordingly

Resolves #1819 .

1. Go to any  non admin user profile.
2. Change user's membership current level 
3. Check `Send the user an email about this change.` checkbox
4. Check inbox and see if header email matches with admin user or modified user.

### Other information:

* [ x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x ] Have you successfully run tests with your changes locally?

### Changelog entry

### DIFF
--- a/classes/class.pmproemail.php
+++ b/classes/class.pmproemail.php
@@ -1005,7 +1005,7 @@
 
 			$this->data = array(
 								'subject' => $this->subject, 
-								'name' => $user->display_name, 
+								'name' => $user->display_name,
 								'user_login' => $user->user_login,
 								'sitename' => get_option( 'blogname' ),
 								'siteemail' => pmpro_getOption( 'from_email' ),
@@ -1275,14 +1275,15 @@
 
 			$this->data = array(
 				'subject' => $this->subject, 
-				'name' => $user->display_name, 
-				'display_name' => $user->display_name, 
+				'name' => $this->get_admin_name($this->email),
+				'display_name' => $this->get_admin_name($this->email),
+				'related_user_name' => $user->display_name,
 				'user_login' => $user->user_login, 
 				'user_email' => $user->user_email, 
 				'sitename' => get_option('blogname'), 
 				'membership_id' => $membership_level_id, 
 				'membership_level_name' => $membership_level_name,
-				'siteemail' => get_bloginfo('admin_email'), 
+				'siteemail' => $this->email,
 				'login_link' => pmpro_login_url(), 
 				'login_url' => pmpro_login_url(),
 				'levels_url' => pmpro_url( 'levels' )
@@ -1458,4 +1459,16 @@
 						
 			return $this->sendEmail();
 		}
+
+		/**
+		 * Gets the admin user name.
+		 *
+		 * @param string $email The admin email address.
+		 * @return string The admin user display name.
+		 */
+		private function get_admin_name($email) {
+			$admin = get_user_by('email', $email );
+			return $admin ? $admin->display_name : 'admin';
+		}
+
 	}
--- a/includes/email-templates.php
+++ b/includes/email-templates.php
@@ -44,7 +44,7 @@ $pmpro_email_templates_defaults = array(
 	'admin_change_admin'       => array(
 		'subject'     => __( "Membership for !!user_login!! at !!sitename!! has been changed", 'paid-memberships-pro' ),
 		'description' => __('Admin Change (admin)', 'paid-memberships-pro'),
-		'body' => __( '<p>An administrator at !!sitename!! has changed a membership level for !!name!!.</p>
+		'body' => __( '<p>An administrator at !!sitename!! has changed a membership level for !!related_user_name!!.</p>
 
 <p>!!membership_change!!</p>
 


classes/class.pmproemail.php
```diff
@@ -1458,4 +1459,16 @@ function sendPaymentActionRequiredAdminEmail($user = NULL, $order = NULL, $invoi
 						
 			return $this->sendEmail();
 		}
+
+		/**
+		 * Gets the admin user name.
+		 *
+		 * @param string $email The admin email address.
+		 * @return string The admin user display name.
+		 */
```
MaximilianoRicoTabo, line 1468: should i add a `@since ` changelog  tag here ?

dparker1005 (changes requested):
I really like how you abstracted the logic to get the display name for an admin into its own method here. This will definitely help when we update all admin emails (not just the admin change email) with this fix. Even better is that you chose to make that new method `private` so that we have the flexibility to change how that function works in the future without worrying about breaking any code outside of this class. Awesome! This could definitely help us down the line if we decide to generalize that method to get the display name for any email address, not just for admins.

My main concern with this approach is that we are changing how the existing `!!name!!` and `!!display_name!!` email template variables work. While it makes sense for those variables to be the name/display_name of the user that the email is being sent to, existing sites are already built to expect those variables to contain the name/display_name of the user that was updated and we don't want to break sites that are already relying on that functionality in custom email templates.

In order to preserve functionality on existing sites, we may need to get a bit clever with our approach. Since we don't want to change how existing email template variables work, one solution could be to:
1. Update the body of the 'header' template in `includes/email-templates.php` to use a new email template variable `!!header_name!!` instead of the `!!name!!` variable that it is currently using
2. In the `sendEmail()` method of the `PMProEmail` class after we filter `$this->data`,  check if `$this->data['header_name']` is empty and, if so, update with `$this->data['name']`. This step will ensure backwards compatibility for all existing PMPro emails
3. For all admin emails, we should now be able to set a new email template variable `!!header_name!!` which we can set to the display name for the admin using the `get_admin_name()` method that you have already created

Let me know if this all makes sense or if you have any concerns. Happy to talk it over

MaximilianoRicoTabo:
Done @dparker1005 . I changed every function sending an email  in the PMProEmail class but just tested the case of the described bug which is a bit dangerous i think. I'd like to test these changes further but i need some help to reproduce different cases. Despite i can figure some of them by reading source code, perhaps different emails are documented somewhere in p2 ? Otherwise, is a good opportunity to document them ? 
Finally, i saw a couple of functions that have parameters than aren't used. Also i see one function that passes de email as a parameter instead of take it from site with `get_bloginfo("admin_email");` Should we refactor this or is it a breaking change thinking on backwards compatibility ?

dparker1005:
Nice job with this! I did some testing today and it has been going well, will do some more tomorrow before approving.

> I'd like to test these changes further but i need some help to reproduce different cases. Despite i can figure some of them by reading source code, perhaps different emails are documented somewhere in p2 ? Otherwise, is a good opportunity to document them ?
The closest thing we have is this blog post which lists most of the various email templates: https://www.paidmembershipspro.com/documentation/member-communications/list-of-pmpro-email-templates/

If there are any emails that aren't there or that you need more information on, ping me and I can give you steps to trigger them.

> Finally, i saw a couple of functions that have parameters than aren't used.
Removing unused function parameters is worth thinking about. The one that I'm seeing is `sendMembershipExpiredEmail()`, are there any others? That one may be good to keep since it's already there. I can see us potentially using that information in the future.

>Also i see one function that passes de email as a parameter instead of take it from site with get_bloginfo("admin_email"); Should we refactor this or is it a breaking change thinking on backwards compatibility ?
It's the `sendBillingFailureAdminEmail()` method that you're looking at, right? I agree that it would be nice to make that method consistent with the others, but as you mentioned, this would be a breaking change. We probably wouldn't address this unless it starts causing issues down the line

MaximilianoRicoTabo:
thank you so much @dparker1005

PR #2468, opened 2023-05-11
Making sure account number is not saved in database when sending to Stripe Checkout

dparker1005:
### All Submissions:

When PMPro sends a user to Stripe Checkout to pay, it saves all $_REQUEST variables so that it can process the checkout when the payment is complete. Normally, credit card fields are not shown when Stripe Checkout enabled, but there are cases when they could be. For example, if multiple gateways are being used and some of them have credit card fields.

This PR ensures that the AccountNumber value in $_REQUEST is never saved to the database.

Resolves XXX.
### Other information:

### Changelog entry

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1650,10 +1650,22 @@ class PMProGateway_stripe extends PMProGateway {
 		// Save some checkout information in the order so that we can access it when the payment is complete.
 		// Save the request variables.
 		$request_vars = $_REQUEST;
-		unset( $request_vars['password'] );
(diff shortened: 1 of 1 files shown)


ideadude:
Kind of related: we should delete the checkout_request_vars order meta when a user is data erased via GDPR request. See here: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/privacy.php#L74

We should still retain the orders, but all of the checkout request var data can be erased. (We may want to erase it entirely after some time in general, but for now let's focus on removing the sensitive info.)

dparker1005:
> 

I imagine that there could be other order meta that could have user data. Would we want to just wipe all order meta for the user's orders to be safe?

PR #2474, opened 2023-05-17
Reworks the password redirect function

JarrydLong:
### All Submissions:

Reworks part of the password reset redirect function to make use of pmpro_url when getting the login page URL

### Other information:

### DIFF
--- a/includes/login.php
+++ b/includes/login.php
@@ -598,25 +598,27 @@ function pmpro_lost_password_form() { ?>
  * @since 2.3
  */
 function pmpro_lost_password_redirect() {
-	if ( 'POST' == $_SERVER['REQUEST_METHOD'] ) {
-		$login_page = pmpro_getOption( 'login_page_id' );
 
-		if ( empty( $login_page ) ) {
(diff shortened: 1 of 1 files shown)


ideadude:
I need to test some more, but noting that we should test the case where folks set the login page to the homepage. In that case, I'm not sure if the URL returned from pmpro_url is "/" or "". If it's the latter, we need to account for that.

Setting the homepage as the login page is something that has caused issues in the past. I would avoid doing it on my own sites, but people do do that and we should try to support it vs just say "this doesn't work". Also if that case breaks, folks are soft locked out of their site (until they realize they need to visit /wp-login.php directly).

dparker1005 (approved):

dparker1005:
I did a quick test setting the homepage as the login page and it seemed to work correctly.

Dumping `add_query_arg( array( 'random-key' => 'random-value' ), pmpro_url( 'login' ) )` on my test site with the login page set to the dashboard gives me `http://generaltesting.local/?random-key=random-value`.

PR #2506, opened 2023-06-07
Prefix level and discount code parameters at checkout

dparker1005:
### All Submissions:

Prefixing the `level`, `discount_code`, and `other_discount_code` parameters at checkout/confirmation. There is still code to support the non-prefixed parameters as a fallback.

Other changes include:
- Adding caching to the `pmpro_getLevelAtCheckout()` function
- Adding a new filter `pmpro_default_discount_code` for cases where you want to dynamically add a discount code at checkout (ex Sitewide Sales)
- Including the discount code used in the level object returned from `pmpro_getLevelAtCheckout()` if a code was used

With these changes, we should no longer have to reply on the `$pmpro_level` or `$discount_code` globals either. We should instead always call the `pmpro_getLevelAtCheckout()` function whenever that information is needed, which should be fairly quick now that it is cached.

### Other information:

### Changelog entry


includes/functions.php
```diff
@@ -2323,56 +2323,72 @@ function pmpro_are_any_visible_levels() {
  */
 function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
 	global $pmpro_level, $wpdb, $post;
+	static $pmpro_getLevelAtCheckout_cache = array();
```
ideadude, line 2326: Have to fix this to avoid warnings when the isset check is made.

Either a better check and setting up the array of arrays on demand. Or concat the level_id . ',' . $discount_code and  hash that and use the hash for the cache key. And/or use the WP get/set cache functions.

includes/functions.php
```diff
@@ -2383,22 +2399,26 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
...
-	// what level are they purchasing? (no discount code)
+	// If we don't have a level object yet, pull it from the database.
 	if ( empty( $pmpro_level ) && ! empty( $level_id ) ) {
 		$pmpro_level = $wpdb->get_row( "SELECT * FROM $wpdb->pmpro_membership_levels WHERE id = '" . esc_sql( $level_id ) . "' AND allow_signups = 1 LIMIT 1" );
 	}
 
 	// filter the level (for upgrades, etc)
 	$pmpro_level = apply_filters( 'pmpro_checkout_level', $pmpro_level );
 
+	// Cache it.
```
ideadude, line 2419: Fix cache code here too.

ideadude (commented):
Fix the caching then merge. We will need to test a bunch, but this is all good otherwise.

PR #2526, opened 2023-06-30
Add "No Access Message" option to DIVI

andrewlimaza:
* ENHANCEMENT: Added a "Show no access message" option to DIVI Paid Memberships Pro settings.
* REFACTOR: Removed duplicate code as they were identical settings functions for rows and sections. We don't have any plan to handle rows/sections differently at this point in time.

### All Submissions:


dparker1005 (approved):
Looks good! Only note is that it may be good to keep `section_settings()` as a deprecated method and have it just throw a deprecation notice and call `row_settings()`; however, this would only matter if there were custom code out there that is using `section_settings()` which seems unlikely.

PR #2550, opened 2023-07-25
Add support to Member HomePages Add On and Elementor

andrewlimaza:
* ENHANCEMENT: Added support for Member HomePages Add On and Elementor builder.

Pushed to core as we already have a compatibility file here and might be easier to maintain moving forward.


andrewlimaza:
Moving this to Member HomePages Add On instead.

PR #2553, opened 2023-07-26
Bug Fix: Pagination fixed when searching orders

JarrydLong:
### All Submissions:

Pagination on the orders page broke when searching

1. Navigate to Memberships > Orders and search for something
2. Results will be returned but the count will show as 1 and no page numbers
3. Once the fix has been applied the pagination shows as expected

### Other information:

### DIFF
--- a/classes/class-pmpro-orders-list-table.php
+++ b/classes/class-pmpro-orders-list-table.php
@@ -349,7 +349,11 @@ class PMPro_Orders_List_Table extends WP_List_Table {
 
 			//Not escaping here because we escape the values in the condition statement
 			$sqlQuery .= 'AND ' . $condition . ' ';
-			$sqlQuery .= 'GROUP BY o.id ORDER BY o.id DESC, o.timestamp DESC ';
+
+			if( ! $count ) {
+				$sqlQuery .= 'GROUP BY o.id ORDER BY o.id DESC, o.timestamp DESC ';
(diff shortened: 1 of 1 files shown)


dparker1005 (approved):

ideadude:
I see what was throwing me off here. I was worried the `GROUP BY o.id` bit was only being used when the $count param was passed in. If that was really needed, it would be needed whether we were getting the count or the results themselves.

In reality, the GROUP BY there is not needed. Often when using LEFT JOINs, you need to GROUP BY in case the LEFT JOIN returns more than one row with the o.id/etc. The GROUP BY will bunch those up so there's only one row per order id, which is what we want here.

Random note here RE the members list table and PMPro 3.0... in that case we are planning to potentially have more than one row per member (one for each membership level) and so don't want to GROUP BY.

So I think a possible fix here is to remove the GROUP BY entirely, it's not needed.

That said, another option is to move the GROUP BY out of the `$count` check. The GROUP BY adds very little overhead here. And it might be nice to have it in case future edits to the SQL or filters change the SQL to have additional LEFT JOINs (e.g. on order meta) that might mean we get more than one row per order id... the GROUP BY would be needed to fix that.

Removing the GROUP BY is the cleaner option. Moving it out of the $count check is the more robust one. I lean toward the latter just in case.

PR #2566, opened 2023-08-09
Setting collate when creating tables

### LINKED ISSUE #162: Database encoding for text
Hi,
There is a problem in some column (text) in the database, the current encoding (collection) is "latin1_swedish_ci" which made a problem when saving non-latin character (Arabic) it saved as question marks!,
some off this column is "name, description, and confirmation" in "wp_pmpro_membership_levels"  table.
I suggest to change all tables encoding to utf-8.

Thanks.

dparker1005:
### All Submissions:

Set the collation setting on PMPro tables to match the setting defined in WordPress. Based off of similar code in WooCommerce:
https://github.com/woocommerce/woocommerce/blob/d5d282dafa2b7db9cbb7a285893a61a40fce05c4/plugins/woocommerce/includes/class-wc-install.php#L1152-L1159

When testing, this change updated the collation settings for new tables that were created to the collation setting defined in `wp-config.php`, but did not update the setting for existing tables. @ideadude Does this sound correct to you?

Resolves #162

### Other information:

### Changelog entry

### DIFF
--- a/includes/upgradecheck.php
+++ b/includes/upgradecheck.php
@@ -370,6 +370,11 @@ function pmpro_db_delta()
 	$wpdb->pmpro_groups = $wpdb->prefix . 'pmpro_groups';
 	$wpdb->pmpro_membership_levels_groups = $wpdb->prefix . 'pmpro_membership_levels_groups';
 
+	$collate = '';
+	if ( $wpdb->has_cap( 'collation' ) ) {
+		$collate = $wpdb->get_charset_collate();
+	}
+
 	//wp_pmpro_membership_levels
 	$sqlQuery = "
 		CREATE TABLE `" . $wpdb->pmpro_membership_levels . "` (
@@ -391,7 +396,7 @@ function pmpro_db_delta()
 		  KEY `allow_signups` (`allow_signups`),
 		  KEY `initial_payment` (`initial_payment`),
 		  KEY `name` (`name`)
-		);
+		) $collate;
 	";
 	dbDelta($sqlQuery);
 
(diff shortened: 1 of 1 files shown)


ideadude:
> but did not update the setting for existing tables. @ideadude Does this sound correct to you?

This sounds correct. I think we don't want to update the collation for existing tables. We can share a SQL script for people to use carefully to update the collation on their existing tables.

However, does the dbDetla function ignore the collation on later calls? For a couple (old) upgrades, we call pmpro_db_delta() and might do so in the future (I think we actually do in the 3.0 upgrade script too). And won't that trigger the DB tables to update their collation?

If so, we maybe need a way to make sure we only set the collation on the first pass for each table. I believe if we leave it out, it will leave the collation alone (that's how it works now).

If having an existing table with latin1 and then upgrading to PMPro 3.0 where the collation would be set DOESN'T update the collation of the changed table, then we can use this code as is and merge it in.

dparker1005:
@ideadude I have just tested whether setting collate in `dbDelta()` calls affects existing tables and concluded that it does not. My test was:
1. Check the existing collate for all PMPro tables. They were set to `utf8_general_ci`.
2. Dump `$wpdb->get_charset_collate()`. It is `DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_520_ci`.
3. Delete a PMPro table (used subscription meta).
4. Ran `pmpro_db_delta()`

After this test, all tables remained `utf8_general_ci` except for the re-generated subscription meta table, which was `utf8mb4_unicode_520_ci`

Code I used to get the collate setting for an existing table:

> $table_name = $wpdb->pmpro_subscriptionmeta;
	$query = "SHOW TABLE STATUS LIKE '$table_name'";
	$table_status = $wpdb->get_results($query);
	if (!empty($table_status)) {
		$collation = $table_status[0]->Collation;
		echo "Current collation for table $table_name: $collation\n";
	} else {
		echo "Table $table_name not found.\n";
	}

Please feel free to duplicate my test if you would like to verify on your own install

ideadude:
Thank you for testing! I'm going to merge now.

FWIW, this is an interesting little tidbit (dbDelta DOESN'T update the collation, even when it's set in the SQL) that could be shared socially/etc.

PR #2567, opened 2023-08-10
Add pmpro_checkout_message filter at bottom

ipokkel:
### All Submissions:

When hooking into the `pmpro_checkout_message` filter to adjust the returned string it only displayed at the top message on the checkout page and not on the bottom message by the submit button. Adding this filter at the bottom as well should allow displaying the filtered message in both locations.

1. Add custom code that hooks into the `pmpro_checkout_message` filter and set a custom checkout error.
2. Check out for a paid membership level using Stripe as the payment gateway and the testing credit card number `4000000000000002` that should be declined.
3. Observer that the custom error message is displayed in both locations on the checkout page.

### Other information:

### Changelog entry

> ENHANCEMENT: Added a pmpro_checkout_message filter at the bottom of the checkout page that can be used to filter error messages shown at checkout.

### DIFF
--- a/pages/checkout.php
+++ b/pages/checkout.php
@@ -551,7 +551,7 @@ if ( empty( $default_gateway ) ) {
 	<div class="<?php echo esc_attr( pmpro_get_element_class( 'pmpro_submit' ) ); ?>">
 		<hr />
 		<?php if ( $pmpro_msg ) { ?>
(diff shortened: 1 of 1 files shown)


ideadude:
Nice catch. Normally editing things so a filter is used more than once should trigger us to think about refactoring things so the filter is only used once.

In this case we could create a new pmpro_get_checkout_message() function or something similar and then call that in 2 places. However, that would be a little complicated right now. The function would have to be able to account for the different id's used in the top and bottom. The top message also has `role="alert"` to help with screen readers, and we only want this in one of the messages.

We want to eventually refactor the notice/message system for the front and back end in general, and creating that function now would make the refactor more complicated.

So in this case, I approve of just using the filter twice to get the expected behavior. Just taking notes here for future reference.

Thanksa gain.

PR #2580, opened 2023-08-16
don't filter custom taxonomies

andrewlimaza:
* BUG FIX: Fixes an issue where custom taxonomies may be filtered when they shouldn't have. (WooCommerce product categories).

### All Submissions:

Resolves: https://github.com/strangerstudios/pmpro-woocommerce/issues/195.

1. Create a post category that requires a level to view it's content.
2. Create a WooCommerce product category with the same name in step 1.
3. Set your "Filter Archives and Search" to "Yes", so only members can see content which they have access.
4. While viewing your site without a membership level, navigate to site/product-category/category-slug

Test step 4 before and after this PR.

### Other information:

### Changelog entry

* BUG FIX: Fixes an issue where custom taxonomies may be filtered when they shouldn't have. (WooCommerce product categories).
* ENHANCEMENT: New filtered added in to allow developers to filter custom taxonomy archive pages from the Filter Archives and Search option. Namely: `pmpro_dont_filter_custom_taxonomy_archives` which takes a boolean parameter to let the category filtering alter the output. (NOTE: This might not be totally needed but thought it'd be a nice to have)

### DIFF
--- a/includes/content.php
+++ b/includes/content.php
@@ -255,6 +255,11 @@ function pmpro_search_filter($query)
         //make this work
         if( $hidden_cat_ids ) {
 
+			// Do not filter on custom taxonomies by default, bail.
+			if ( apply_filters( 'pmpro_dont_filter_taxonomy_archives', is_tax() ) ) {
+				return;
+			}
+
 			// Get all registered category ID's so we may remove the ones we need to hide. This is to support posts that may belong to multiple categories and only one of them is hidden.
 			$all_cat_ids = get_terms( array(
 				'taxonomy' => 'category',


MaryOJob:
Moderators Only: PR worked to resolve the issue for this ticket #526161

andrewlimaza:
This works in some cases but not ideal. The actual issue for Woo is that their product category pages uses the default "category" taxonomy and that's why we're filtering it. I don't think this should be merged in after spending more time on it.

It might be okay for a patch but looking into improving upon this from another angle. I've left it as a draft for now to keep it easily available.

Issue #1667, opened 2021-06-02
Password reset functionality broken in some cases


vdwijngaert:
Hi there!

When the default reset password page (`wp-login.php?action=rp&key=*KEY*&login=*LOGIN*`)  is loaded, the PMPro plugin hijacks the default form and redirects to the PMPro login page on the frontend. It uses the `login_form_resetpass` hook for that: https://github.com/strangerstudios/paid-memberships-pro/blob/bf1d2376e77e7156aac52e5cf126f4268a04bec6/includes/login.php#L631.

In that function, [the code however assumes](https://github.com/strangerstudios/paid-memberships-pro/blob/bf1d2376e77e7156aac52e5cf126f4268a04bec6/includes/login.php#L613) the key and login request variables are named `rp_key` and `rp_login`, which are undefined, resulting in the `invalidkey` error being triggered. 

A temporary workaround is using a plugin to dynamically populate those variables, if necessary:

```php
if(isset($_GET['key'], $_GET['login'])) {
    $_GET['rp_key'] = $_GET['key'];
    $_GET['rp_login'] = $_GET['login'];
}
```

The custom logic PMPro uses probably takes care of this, but when for some reason, the default WP handler was used for generating the reset link, this is an issue. I've seen it happen in at least two ways:

1. When in a multisite-environment, reset logic has been moved to the currently active site, instead of the network site.
2. When a reset link from before activating the PMPro plugin is being used while PMPro is active.  

Thanks for taking a look at this.

PR #2595, opened 2023-08-17
Refactor logic on the password reset/lost password functionality.

andrewlimaza:
### Changelog entry

* BUG FIX/ENHANCEMENT: Improved the Lost Password/Reset Password functionality provided by the frontend login. This now honors password requests coming from elsewhere and will try not override relevant information when we detect this (i.e. requests from wp-login.php will continue the flow throughout wp-login.php and not be redirected to the /login frontend page of PMPro and vice versa).

### All Submissions:

The focus of this PR is to support Paid Memberships Pro frontend password reset functionality (Lost password and resetting the password) as well as honoring the wp-login.php flow (as users of a site can navigate to both URL's to do the same functionality).

Improvements to the redirect to password reset page has been simplified which was causing majority of our conflicts.

PMPro makes use of some custom query parameters to process these requests and we try to replace URL's and parameters with our own when the request comes from our page/forms. This PR should solve these cases, and potentially fix issues where we hijack URLs/emails mistakenly.  (See steps below to test this PR).

Resolves https://github.com/strangerstudios/paid-memberships-pro/issues/1667 (and other issues that may not have been marked here).

Before testing this pull request, please test password reset from Paid Memberships Pro frontend login page, and then password reset from the wp-login.php (default) password reset functionality. Issues have been occurring between these two areas and other plugins like TML (and most likely others) hijack the process and there's no easy way to prevent this right now.

1. Set your PMPro Log In page, in the page settings, to a page that uses `[pmpro_login]` or the block.
2. Navigate to your /login frontend page from step 1, reset your password and follow the steps. (All steps should remain in the /login path/frontend page without redirecting).
3. Navigate to wp-login.php and reset your password through the default WordPress page. Your entire journey during this step should remain on the wp-login.php path (No hijacking or replacing of URLs should occur).

This could also be tested with other plugins like TML or a frontend login plugin but for the purposes of this PR was to focus on the two points: PMPro Log In and wp-login.php. I noticed other plugins hijack our requests and nuke all other requests which make it difficult to support but it could fix some compatibility issues we've seen in the past.

### Other information:

### DIFF
--- a/includes/login.php
+++ b/includes/login.php
@@ -789,6 +789,11 @@ function pmpro_do_password_reset() {
 		return;
 	}
 
+	// We're on the default WP page, let's not try to interfere.
+	if ( ! empty( $_REQUEST['wp-submit'] ) ) {
+		return;
+	}
+
 	$rp_key = sanitize_text_field( $_REQUEST['rp_key'] );
 	$rp_login = sanitize_text_field( $_REQUEST['rp_login'] );
 
@@ -864,6 +869,11 @@ function pmpro_password_reset_email_filter( $message, $key, $user_login ) {
 
 	$login_url = pmpro_url( 'login' );
 	if ( ! $login_url ) {
+		return $message;
+	}
+
+	// Don't replace the password reset link if it came from the default WP form. (Our form uses 'submit')
+	if ( ! empty( $_REQUEST['wp-submit'] ) ) {
 		return $message;
 	}
 
--- a/includes/login.php
+++ b/includes/login.php
@@ -647,12 +647,6 @@ function pmpro_reset_password_redirect() {
 		return;
 	}
 
-	// While doing the redirect, if the pages initial load isn't our login page. Don't redirect. (i.e. Another plugin called this password reset)
-	$current_page_slug = get_permalink();
-	if ( $current_page_slug !== $login_url ) {
-		return;
-	}
-
 	// Get current REQUEST PARAMS and just add it to ours and then redirect to our login_url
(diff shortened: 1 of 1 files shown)


dparker1005:
Maybe I’m oversimplifying here, but what if we entirely deprecated the pmpro_reset_password_redirect() function? With this PR, we’re preventing the redirect in that function from running on the core WP login page, so the only place that function would redirect is on the password reset pages of 3rd party plugins. If that’s the case, it seems like all we’re doing is hijacking other plugins’ PW reset processes, which is frustrating when other plugins do this to us.

Are we still getting any benefits from that pmpro_reset_password_redirect() function that I am missing? Are there cases that it helps with?

kimcoleman:
Turns out we do indeed need the pmpro_reset_password_redirect function, since our own frontend “lost password” form submits to the WP form and then sends the user back to the PMPro login page with the message of whether the submit to reset sent the email or not.

Without this, our own password reset form, as its currently coded, wouldn’t show on the frontend if people use the PMPro frontend login. And we want errors to show on the frontend login form if there are errors.

Some of the complexity of this code, in the way it is different from WP core is due to us (1) trying to support TML which was the majority of people moving to PMPro’s FE login and (2) WP core updating how they do logins recently and way back in the past before TML.

We could probably refactor this down the road not to submit to the wp-login.php form for our own reset and rather trigger this all on our own. This would mean we don’t need that redirect anymore.

I’m going to merge this in as is for now. It fixes things without breaking anything else.

Tested with other password / frontend login stuff like Woocommerce and all working well.

PR #2613, opened 2023-09-06
Improvements to "Extra Fields"

andrewlimaza:
* BUG FIX: Added better compatibility for User Fields and Groups created via code and only show fields in the checkout email that are for that particular checkout.

### All Submissions:

Resolves XXX.

1. Create User Fields and Groups via code and the settings page.
2. Test checkout and see that "Extra Fields" are correctly displaying for the current checkout level and both coded and setting fields show.

### Other information:

### DIFF
--- a/includes/fields.php
+++ b/includes/fields.php
@@ -84,6 +84,7 @@ function pmpro_add_field_group( $name, $label = NULL, $description = '', $order 
 	$temp->label = $label;
 	$temp->description = $description;
 	$temp->order = $order;
+	$temp->checkout = 'yes';
 
 	//defaults
 	if( empty( $temp->label ) ) {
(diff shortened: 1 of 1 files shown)


includes/fields.php
```diff
@@ -1065,19 +1068,24 @@ function pmpro_add_user_fields_to_email( $email ) {
...
-					}
+			}
+			
+
+			// Remove any field groups that are not for the membership level that was purchased.
+			foreach ( $fields_groups as $key => $group ) {
+				if ( ! empty( $group->levels ) && ! in_array( $level_id, $group->levels ) ) {
+					unset( $fields_groups[ $key ] );
 				}
 			}
```
dparker1005, line 1082: Previously, we had the `foreach` inside the `if` in case `$email->data['membership_id'];`, and there for `$level_id`, is not set. Should we default `$level_id` to a value in this case to avoid PHP errors? We also check `$level_id` lower down in the `foreach` loop of fields, so we want to make sure to avoid errors in both of those cases

dparker1005 (commented):
Haven't been able to replicate the original issue yet, but just some notes from looking over the code. Will review again once I can replicate

dparker1005:
Pushed a commit to simplify the "add user fields to admin checkout emails" functionality overall. The key change is not worrying about whether the group is shown at checkout, which is what made the previous code complicated. The group settings are already trickling down to each individual field, so as long as we check each field correctly, the correct fields will make it into the email content.

I also found an unrelated bug in `pmpro_get_label_for_user_field_value()` where field values could be replaced with labels from an entirely different field. That function now checks that the field that we are pulling a label from is the same field that we have a value for.

andrewlimaza:
Thanks @dparker1005 all works well and moving to simpler code makes sense.

Issue #2619, opened 2023-09-15
Inconsistent autogenerated values for user fields with options.


ipokkel:
**Describe the bug**
User fields, when having multiple required select fields and setting the values for a select field with only the first value having a _value:label_ the registration check returns all other fields need filling in if only one of the select fields were not filled in.

``` 
:Select 
Yes
No
```

It appears that if a select field has a _value:label_ option set the rendered values are numerical and when the field does not have a _value:label_ set the rendered values are as entered.

**To Reproduce**
1. Go to **Memberships > Settings > User Fields** and create a `select` type field with the following options:
``` 
:Select 
Yes
No
```
2. Create another `select` field with the following options:
``` 
Yes
No
```
3. Save all field settings.
4. Navigate to the checkout page and review the values set for these `select` type fields in the browser's tools.

**Screenshots**
![2023-09-14_13-22-45](https://github.com/strangerstudios/paid-memberships-pro/assets/9744519/3c115e68-b542-41a6-93fd-41f0be6bd872)

**Expected behavior**
The automated value and label generated for options to be consistent.

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only Paid Memberships Pro plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Memberlite](https://wordpress.org/themes/memberlite/).
- [x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
<details>
Paid Memberships Pro 2.12.2
</details>

dparker1005:
To me, it does not feel like the flexibility that we are trying to give admins by making `:` optional is worth the amount of time we spend trying to eliminate every edge case. If we made a bold decision to require that all field options be `key:label`, we could enforce that requirement on user field save and "fix" user input that doesn't fit that pattern. Users can then see the correct way for the data to be formatted and understand what their user fields will look at on the frontend.

ex. User tries to save option `Yes`,  after save if they edit the field again, they will see `Yes:Yes`.

Or, maybe we even enforce `key:label` in the UI by having two separate text fields for each option: one for key and a separate text field for label. In this case, we could automatically "guess" at option keys based on the label that the admin creates, similar to how we automatically try to guess field names from field labels that are entered.

PR #2624, opened 2023-09-27
Inconsistent autogenerated values for user fields with options. #2619

MaximilianoRicoTabo:
* If there's a key, we assume the intention was for the other options to use the values as keys.

### All Submissions:

Resolves [2619](https://github.com/strangerstudios/paid-memberships-pro/issues/2619).
Check the options to see if we have a key on any of its option then let's assume the intention was for the other options to use the values as keys. Check for a ':' and if it exists we name all the key after its values. 

1. Create different user fields with selects
2. Add different combinations with only values, with single key/value option, with empty key and a value, etc.

### Other information:

### Changelog entry

> Fixes Inconsistent autogenerated values for user fields with options

### DIFF
--- a/includes/fields.php
+++ b/includes/fields.php
@@ -1582,13 +1582,21 @@ function pmpro_load_user_fields_from_settings() {
             if ( in_array( $settings_field->type, $option_types ) ) {
                 $options = array();
                 $settings_options = explode( "\n", $settings_field->options );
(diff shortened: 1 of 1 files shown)


dparker1005:
This doesn't feel like the exact solution that we want for this issue. I'm going to close this PR and keep the issue open and we can continue discussing there to find the path that we want moving forwards.

Issue #2615, opened 2023-09-07
Duplicate user fields when duplicate field groups exists.


ipokkel:
**Describe the bug**
Two copies of a user field group and associated fields are displayed on the checkout page.

This may occur if the same field group is created with code and through creating a field group in the GUI Settings for User Fields.

**To Reproduce**
1. Create a user field group and a user field with code and add the code snippet to a customization plugin. (Example: https://gist.github.com/ipokkel/a97eb6c5308b0dda3fc07d84df47771a)
2. In the WordPress dashboard navigate to Members > Settings > Users Fields and create a user field group with the same group name as the group created with code, e.g. "More Information", add a field to the group, and save settings.
3. Navigate to the front-end membership checkout page.

**Screenshots**
![2023-09-06_16-38-48](https://github.com/strangerstudios/paid-memberships-pro/assets/9744519/f926ef51-1d4a-44d6-a5b8-1495b55d7958)

**Expected behavior**
A field group should be displayed once only on the checkout page.

A workaround for this is to remove the line that creates the field group using the `pmpro_add_field_group` function from the customization code when a group with the same name was created in the GUI Settings for User Fields.

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only Paid Memberships Pro plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Memberlite](https://wordpress.org/themes/memberlite/).
- [x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
WordPress 6.3.1
PMPro 2.12.2

MaximilianoRicoTabo:
<img width="609" alt="image" src="https://github.com/strangerstudios/paid-memberships-pro/assets/1678457/89e7c85f-8f1e-4c22-a67b-f61812171684">
It also adds dup ID which is not ok

MaximilianoRicoTabo:
https://github.com/strangerstudios/paid-memberships-pro/pull/2625

andrewlimaza:
The fix has been merged in.

PR #2625, opened 2023-09-28
Duplicate user fields when duplicate field groups exists. #2615

MaximilianoRicoTabo:
* Bail if the group already exists in pmpro_add_field_group function

### All Submissions:

 Check if group name exist before add it. From ou UI this won't happen because it handles in a way that adds a number within parenthesis to the name if it's repeated before submit the form in a jQuery event handler.  So, despite it's an edge case probably causes some trouble to support team and it's good to handle.

Resolves  https://github.com/strangerstudios/paid-memberships-pro/issues/2615

1. Follow steps from issue above
2.  Observe if you see duplicates or not

### Other information:

### Changelog entry

Added validation for duplicated names in pmpro_add_field_group function


includes/fields.php
```diff
@@ -78,6 +78,12 @@ function pmpro_add_user_field( $where, $field ) {
  */
 function pmpro_add_field_group( $name, $label = NULL, $description = '', $order = NULL ) {
 	global $pmpro_field_groups;
+	//Bail if the group already exists.
+	foreach($pmpro_field_groups as $group) {
+		if($group->name == $name)
+		//already exists	
+		return false;
+	}
```
ideadude, line 86:
```suggestion
	//Bail if the group already exists.
	foreach( $pmpro_field_groups as $group ) {
		if( $group->name === $name ) {
			//already exists
			return false;
		}
	}
```
WPCS and using === here. Both values should always be strings, but just in case.

ideadude (commented):
Minor WPCS fix.

I'm also wondering if we should return false here or maybe overwrite the group info with the new data on the second call?

Do we know what the user would intend?

When is this coming up most often? I assume maybe there is a group in the user field settings in the admin, then we run some custom code afterward that is trying to add the groups. In this case, should the code (or whatever call comes later) overwrite the group label and description then?

includes/fields.php
MaximilianoRicoTabo: Thanks. Added the suggested change. Regarding  the questions 

> Do we know what the user would intend?
> When is this coming up most often?

From the issue itself it seems it comes from custom code added to functions.php file in the theme or snippet plugin. Actually, we deal with dups in our UI by adding a number between parenthesis. We may try that approach for dups coming from custom code as well, but also we may enforce users to adopt user fields from UI, i guess is something that comes from RH deprecation and should be an edge case and should tend to stop happening.

includes/fields.php
```diff
@@ -78,6 +78,12 @@ function pmpro_add_user_field( $where, $field ) {
  */
 function pmpro_add_field_group( $name, $label = NULL, $description = '', $order = NULL ) {
 	global $pmpro_field_groups;
+	//Bail if the group already exists.
+	foreach($pmpro_field_groups as $group) {
+		if($group->name === $name)
+		//already exists	
+		return false;
+	}
```
dparker1005, line 86: WPCS
```suggestion
	// Bail if the group already exists.
	foreach ( $pmpro_field_groups as $group ) {
		if ( $group->name === $name ) {
			// Group already exists.
			return false;
		}
	}
```

PR #2641, opened 2023-10-30
Fixed issue with membership cancellation

rsm0128:
SQL query to get the latest order for a user was incorrect and it caused an issue with refund.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1061,7 +1061,7 @@ function pmpro_changeMembershipLevel( $level, $user_id = null, $old_level_status
 		$pmpro_cancel_previous_subscriptions = apply_filters( 'pmpro_cancel_previous_subscriptions', $pmpro_cancel_previous_subscriptions );
 
 		$other_order_ids = $wpdb->get_col(
-			"SELECT id, IF(subscription_transaction_id = '', CONCAT('UNIQUE_SUB_ID_', id), subscription_transaction_id) as unique_sub_id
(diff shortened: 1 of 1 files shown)


rsm0128:
```
		$other_order_ids = $wpdb->get_col(
			"SELECT id, IF(subscription_transaction_id = '', CONCAT('UNIQUE_SUB_ID_', id), subscription_transaction_id) as unique_sub_id
											FROM $wpdb->pmpro_membership_orders
											WHERE user_id = '" . esc_sql( $user_id ) . "'
												AND status = 'success'
											GROUP BY unique_sub_id
											ORDER BY id DESC"
		);
```
We need to use group function Max instead of `ORDER BY id DESC`

andrewlimaza:
@rsm0128 What was the issue with the refund? Did it refund the wrong order/subscription?

Please let me know how I can replicate this on PMPro's side so we can thoroughly test it.

rsm0128:
@andrewlimaza `GROUP BY unique_sub_id` groups orders and we select `id`. But it returns random order id in the group not the latest order id.
`ORDER BY id DESC` is not necessary here. We should use `MAX(id)`

rsm0128:
@andrewlimaza Any update on this?

dparker1005:
Hi @rsm0128,

In PMPro v3.0, the logic for terminating subscriptions associated with cancelled memberships was overhauled to use the new `PMPro_Subscriptions`.  This eliminated the query that this PR modified. 

Can you please confirm that this issue is resolved in PMPro v3.0+?

dparker1005:
I am closing this PR as it should not still be an issue in PMPro v3.0. If this is still an issue that you are seeing on your website, please open a new PR/Issue and we can look into this behavior further there

PR #2663, opened 2023-11-14
Updating confirmation page logic for 3.0

dparker1005:
### All Submissions:

The main functional change in this PR is that the `$pmpro_invoice` global is no longer being unset during the confirmation preheader if the checkout was free. Instead, free checkouts are accounted for in the logic of the confirmation page itself.

The other changes in the confirmation page template code basically make the template rely solely on the `MemberOrder` passed in via `$pmpro_invoice` instead of pulling information from other sources. This eliminates the need to use the `$current_user` global and grab the membership level for the user separately.

### Other information:

### Changelog entry

### DIFF
--- a/pages/confirmation.php
+++ b/pages/confirmation.php
@@ -36,7 +36,7 @@ if ( empty( $pmpro_invoice ) ) {
 		$pmpro_invoice->user->membership_level = $pmpro_invoice->membership_level; // Backwards compatibility.
 
 		// Start building the confirmation message.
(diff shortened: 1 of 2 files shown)


kimcoleman (approved):
Noting that the template version should be updated now or pre-3.0 release. I would suggest now but up to you.

kimcoleman:
There is an issue in the preheader check about $confirmation_level and backwards compatibility (it is also resetting to null). I sent a message in Slack that we can tackle this in this PR before merge or another PR.

PR #2664, opened 2023-11-14
Improving MMPU compatibility for cancel page

dparker1005:
### All Submissions:

This PR eliminates the use of some unnecessary globals and avoids using `$current_user->membership_level` which is not good practice with MMPU.

### Other information:

### Changelog entry

### DIFF
--- a/pages/cancel.php
+++ b/pages/cancel.php
@@ -10,7 +10,7 @@
  *
  * @author Paid Memberships Pro
  */
-global $pmpro_msg, $pmpro_msgt, $pmpro_confirm, $current_user, $wpdb;
+global $pmpro_msg, $pmpro_msgt, $current_user, $wpdb;
 
 if(isset($_REQUEST['levelstocancel']) && $_REQUEST['levelstocancel'] !== 'all') {
 	// Odd input format here (1+2+3). These values are sanitized.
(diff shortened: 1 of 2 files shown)


kimcoleman:
Testing this and see some more opportunities to improve. We could do these in a separate PR but here are some things:

1. When cancelling "all", we could have better wording on the page in this screenshot:
Screenshot 2023-11-24 at 10 07 03 AM
 
`
Are you sure you want to cancel all of your memberships?
[ Yes, cancel all memberships ] No, keep all memberships
`

2. Should we consider prefixing the levelstocancel URL attribute? Changing it? Now's the time. I highly doubt in 2,000 years this same URL parameter would cause conflicts (for the reasons we adjusted the ?level checkout URL parameter) but noting that here for consideration.

3. After you cancel a single membership, if you still have "some" other levels, we should adjust the "Click here to go to the home page." to rather be  "View Your Membership Account" as we do on the single invoice.

Screenshot 2023-11-24 at 10 09 47 AM

4. Since we are now MMPU environment, should all the language across single level cancel process include the level name? We should the level name on the single cancel page but we don't repeat it on the confirm page.

When we are cancelling all, we could list them all that are going to be cancelled.

5. CONPD merge - this isn't working for me on a same day checkout that I then cancel right away (at least with the built in testing or pay by check gateway setting). I can test same day cancel with Stripe / others also.

Older existing subscriptions that I cancel with recurring are triggering properly.

Should we consider updating how users see their "Billing" to use strikethrough OR show a "cancelled" label on the membership account page?
Screenshot 2023-11-24 at 10 20 02 AM

kimcoleman (approved):
I made some adjustments to give more clear messaging when cancelling a single level (and still having some), cancelling two levels and still having some, and also cancelling all messages. Also did the a11y update for the tbody here for the first item to use <th>. We did this on levels table also so good to catch everywhere we can.

Ready for merge.

PR #2666, opened 2023-11-14
Deprecating global $pmpro_levels

dparker1005:
### All Submissions:

Removing uses of the `$pmpro_levels` global besides for those needed for backwards compatibility. The `pmpro_getAllLevels()` function should be used instead.

Deprecating that global also allows simplifying the arguments for the `pmpro_getAllLevels()` function as the second `$use_cache` argument was unclear.

Looking at our Add Ons, it does not any use the `$pmpro_levels` global without setting it using `pmpro_getAllLevels()`, so backwards compatibility with our Add Ons should not be an issue. The following regex can be used to find all instances of the global being defined: `^(?=.*\bglobal\b)(?=.*\$pmpro_levels).*`

### Other information:

### Changelog entry

### DIFF
--- a/preheaders/checkout.php
+++ b/preheaders/checkout.php
@@ -815,6 +815,9 @@ if ( empty( $submit ) ) {
 	}
 }
 
+// Preventing conflicts with old checkout templates that depend on the $pmpro_level global being set.
+pmpro_getAllLevels();
+
 /**
  * Hook to run actions after the checkout preheader is loaded.
  * @since 2.1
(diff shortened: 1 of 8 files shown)


preheaders/checkout.php
```diff
@@ -819,6 +815,9 @@
 	}
 }
 
+// Preventing conflicts with old checkout templates that depend on the $pmpro_level global being set.
+pmpro_getAllLevels();
```
kimcoleman, line 819: Curious how to test just this one piece. Why we need this and how we'd know it is or isn't needed.

shortcodes/pmpro_account.php
```diff
@@ -4,7 +4,7 @@
 */
 function pmpro_shortcode_account($atts, $content=null, $code="")
 {
-	global $wpdb, $pmpro_msg, $pmpro_msgt, $pmpro_levels, $current_user, $levels;
+	global $wpdb, $pmpro_msg, $pmpro_msgt, $current_user, $levels;
```
kimcoleman, line 7: I don't see $levels used anywhere in this. Perhaps that global isn't needed anymore either? Just asking - could be a separate PR.

kimcoleman (changes requested):
I have a couple comments in here to chat through but mostly otherwise approved.

preheaders/checkout.php
kimcoleman: Confirming this is needed and resolving.

shortcodes/pmpro_account.php
kimcoleman: Going to remove $levels but also I don't think we need pmpro_msg or pmpro_msgt. Removing both - we will see if we find that these are needed. This code wasn't touched since 8+ years ago. I think it isn't used.

Leinad4Mind:
Just one doubt, is global $pmpro_level deprecated too? Or only the plural one?

dparker1005:
The `$pmpro_level` global is still available at checkout, but we recommend switching to using `$pmpro_level = pmpro_getLevelAtCheckout();` to make sure that the level object has the correct unaltered information.

PR #2679, opened 2023-11-17
Deprecate edit user UI

dparker1005:
### All Submissions:

- Deprecate "Edit User" UI in favor of the Single Member dashboard.
- Add the `pmpro_after_membership_level_profile_fields` hook to the "other" tab of the Single Member dashboard

Note: The PMPro Unlock Add On looks to be the only add on that uses the `pmpro_after_membership_level_profile_fields` and has save functionality for the information shown on that hook. This Add On should be updated to have its own Single Member Dashboard panel before 3.0 is released. I opened an Issue for this conflict here:
https://github.com/strangerstudios/pmpro-unlock/issues/13

### Other information:

### Changelog entry


kimcoleman:
This all looks good - but we need some way to connect the Edit User page to the Edit Member page. Not only bwcause that's where people are historically looking but also so that people can logically see how to get to adding member/editing from an existing edit user screen. I'm thinking about that....

One note on this: user fields will still show. I think this is ok and good. Plus no new code needed as it already works this way.

We need to make a list of Add Ons that instead are using the edit_user_profile and show_user_profile hooks and should instead use pmpro_after_membership_level_profile_fields.

We should also consider whether a better hook name should be used instead of the one we were using so that it makes more sense moving forward that these are on the "other" tab. Really every panel of the Edit Member screen needs its own hook 
(maybe we have this) and then Add Ons can decide where their hooked in feature makes the most sense. Or a new panel...

kimcoleman:
I almost wonder whether the hook pmpro_after_membership_level_profile_fields should go on the User panel, not the "other" panel.... let's have a conversation about this.

kimcoleman:
Chatted to Jason - he felt that a basic view only table of all their memberships would be good - a quick overview. Here's an example of that for a user with levels and one without.

Screenshot 2023-11-24 at 3 15 23 PM

Screenshot 2023-11-24 at 3 15 16 PM

We could style this more .. or not.

Also need to decide if this is replacing and using the same function name vs. deprecate old function and replace with this. I vote just update the function. If you are good with that I'll commit my update to this PR.

PR #2700, opened 2023-12-04
Enable backwards-incompatible 3.0 queries

dparker1005:
### All Submissions:

Enable backwards-incompatible 3.0 queries.

### Changelog entry

### DIFF
--- a/includes/updates/upgrade_3_0.php
+++ b/includes/updates/upgrade_3_0.php
@@ -46,14 +46,14 @@ function pmpro_upgrade_3_0() {
 		SET status = 'success'
 		WHERE status = 'cancelled'
 		";
-	//$wpdb->query( $sqlQuery ); // Disabled for now to not interfere with development sites.
+	$wpdb->query( $sqlQuery ); // Disabled for now to not interfere with development sites.
 
 	// PMPro Stripe Billing Limits Add On has been merged into core and no longer needs 'pmpro_stripe_billing_limit' user meta.
 	$sqlQuery = "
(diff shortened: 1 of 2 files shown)


kimcoleman:
I think this is a tricky to test PR - but I'll clone my local and set the version to 3.0 for testing.

What concerns do we have about this "timing out" on LARGE sites? Do we need it to run as a background process?

We should also update the comment in includes/upgradecheck.php to include that this is a destructive/not reversible DB update and what it does. Something like:
`/**
 * Version 3.0
 * Running pmpro_db_delta to add subscription and subscription meta tables.
 * Populate subscription and subscription meta tables based on this site's order data.
 * Irreversible changes: All orders in `cancelled` status will be set to `success` so that we can remove `cancelled` status.
 */
`

Some additional thoughts - for Beta, we are going to (according to talks with Jason) set version to something like: 2.99.99. Then if we have other betas OR when we get to RC land, will be 2.99.991, 2.99.992, etc.

He told me it is ok to trigger this with the first beta release though and set DB version in the upgrade script to 3.0 even if the plugin isn't on 3.0 officially in the plugin header / PMPRO_VERSION constant. Hopefully we don't find other db scripts we need. But he also said - db version doesn't have to match upgrade script version / plugin version.

So we can adjust the check in includes/upgradecheck.php for beta to run on 2.99.99, do it again if we need to run things again, then do it once more right before real 3.0 release and set to 3.0 for version compare.

kimcoleman (approved):

PR #2717, opened 2023-12-08
Ignore taxonomy queries and categories (for now)

andrewlimaza:
Ignore taxonomy queries when categories are involved. This is set to default to Woo product categories,can be removed.


andrewlimaza:
Here is the issue - https://github.com/strangerstudios/pmpro-woocommerce/issues/195. (It's related to PMPro core, not our Woo integration).

PR #2746, opened 2023-12-22
Now creating subscription when constructing orders

dparker1005:
### All Submissions:

Previously, subscriptions were only created when an order is saved. This works fine in most cases, but not in a scenario like this:
1. Site upgrades to 3.0
2. Site downgrades to 2.x for whatever reason
3. User checks out for recurring membership on 2.x
4. Site switches back to 3.0

In this case, the upgrade script would not create a subscription for this user since the upgrade script won't run twice. Without this PR, that means that a subscription object wouldn't be created for this user until an order for the subscription is saved, which likely wouldn't happen until the next renewal. But with this PR, a subscription order only needs to be loaded in order for the subscription to be created, which is much more common and may just happen over time by chance.

Resolves XXX.

### Other information:

### Changelog entry

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -293,6 +293,11 @@
 				} else {
 					$morder = $this->getMemberOrderByCode( $id );
 				}
+
+				// If we found an order, create a subscription if needed.
+				if ( ! empty( $this->id ) ) {
+					$this->get_subscription();
+				}
 			} else {
 				$morder = $this->getEmptyMemberOrder();	//blank constructor
 			}


ideadude:
The only downside I can think here is that we are now preloading the subscription every time we load an order. This is nice for forcing the transition on demand as orders are loaded.

This could have a performance impact on the orders screen in the admin, and a bigger performance impact on order exports. For every order loaded, we would at least have to query the DB and possibly query the gateway.

Do we want to mitigate that in some way?

dparker1005:
Merging as subscription lookup should be constant time when a subscription ID, gateway, and gateway environment is provided.

PR #2747, opened 2023-12-22
Removing unneccessary global $pmpro_giving_level

dparker1005:
### All Submissions:

The `$pmpro_giving_level` global is not needed and adds unnecessary confusion to the code. Instead, we should just check whether `$status` is `changed`, which should only be the case if the user is being assigned a membership level at the same time.

### Other information:

### Changelog entry


ideadude:
I think we need to check for admin_changed as well. e.g. https://github.com/strangerstudios/paid-memberships-pro/blob/3a3ec68cf152f56167e3bdbeba58b5e10a3f8778/includes/profile.php#L296

dparker1005:
Fixed, thanks!

PR #2750, opened 2024-01-04
Fix pmpro_search_filter to support custom taxonomies and archives better.

ideadude:
There is an issue where WooCommerce products are incorrectly being filtered out of searches and archive pages, even the product cat archive pages, when the "filter searches and archives" advanced setting is set.

In reality, we want to just filter out restricted members-only content, and allow everything else.

There was a previous fix here (https://github.com/strangerstudios/paid-memberships-pro/pull/2717) but in reviewing that we realized that this whole function/etc is due for refactoring.

We don't want to refactor for the sake of it, but refactoring this will make it possible to fix this but, some related bugs, and to better maintain this code going forward.

It was hard to limit the scope of the refactor. I think we'll end up rewriting a lot of this. We can merge it into the 3.0 release to help minimize frustrations with this code changing. Even though we should end up with something that is "better" and/or something that fixes the filter to function as intended, people have been running sites off this code and may expect things to keep working "the old way". i.e. people will update and may notice that some things that used to be hidden are no longer hidden... some things that used to be visible are no longer visible.

Okay. So far, we've:
* Removed some redundant code to cast the post_types query var to an array.
* Broken out the logic for when the filter runs into individual checks that return early.
* Fixed a hard coded dev_ DB prefix in the where clause filter code.
* Tweaked the post_type check so that the filter may still run even if the post_type query var was empty.
* Added an archive check to skip the filter if we're viewing an archive that is not in the list to be filtered. 

Those last 2 bullet points basically fix the main issue that was noticed. If you navigate to a WooCommerce product category, the products are no longer being filtered (unless you adjust the pmpro_search_filter_post_types filter to include products).

But we have some other issues and things we might want to look at.

Other issues:
* While the current version of the PR fixed archive views, we are still filtering products out of searches even if products aren't in the list of post types to filter.
* Even if products are in the list of post types to filter, the filter isn't working correctly for WC products or probably other CPTs as well.

To address these, we need to look into the code the sets the `category__in` query var. We only look up post categories to set this var, so it's not going to include any taxonomies that aren't for posts. So basically any other CPT in a cat gets excluded.

(Note: We want to tests posts and products without a query too. These should show up in the search if they aren't otherwise restricted.)

We maybe can switch to using `category__not_in` instead, but it seems like we might have had a reason to do it the other way. We should probably rethink this from scratch.

Other things to look into:
* This code looks like it is potentially making at least 1 additional query for every level, and it may not be caching this stuff well if there are multiple queries on page load that require filtering like this. We should try to combine queries and basically make this as performant as possible. It seems like the query to get cats could be one query for all levels instead of individual queries.
* This is all called the *search* filter. Should we filter archives using a totally separate callback? The functions could start with query checks to see if it looks like a search or archive and return early otherwise. Splitting things up may help.
* We are bailing on the filter if the query is a REST API call. Does this mean that private content could potentially leak through REST API calls? Are we otherwise protecting that? Why do we need to bail on REST API queries anyway? Maybe there is a better check.

### All Submissions:

Resolves: https://github.com/strangerstudios/pmpro-woocommerce/issues/195

1. Install WC, PMPro
2. Create a couple products. Put them in the same product category.
3. Regular setup of PMPro, NO products are marked to require membership or have membership discounts.
4. Make sure the Memberships > Settings > Advanced Settings > Filter Searches and Archives setting is set to "Yes"
5. Visit the category archive page. Note if 2 products are shown.

We also want to test searches for terms that should turn up the products, but currently aren't (in the first version of the PR).

Test both searches and archive pages. Test custom archives and post category archives. Make sure that posts which should be hidden are hidden and posts that should be visible are visible. This can get confusing. Have some posts that are in cats that are restricted. Have some posts that are restricted by level themselves. Have some posts that aren't restricted by level. Have posts in non-member cats. Have posts in no cat. Restrict a product using custom code or the PMPro CPT add on. Make sure they are hidden when expected.

### Other information:

### Changelog entry

> BUG FIX/ENHANCEMENT: Fixes to the code used to filter members-only content from searches and archives.


ideadude:
Some updates.

I've merged in the 3.0 branch and changed it to PR against that branch.

I've moved the code around filtering PMPro pages out of search results to its own callback. This code will now always run. You can unhook it if you want PMPro pages in search results. We thought it might be nice to have specifically the levels page, login page, and maybe account page in the search results... but mostly because these shortcodes don't render well in excerpts, we decided to just remove them. These pages are so important that they often have primary navigation links, and so I'm okay that they get removed from search results. They generally clutter up the results when folks are searching for something else.

The original WooCommerce issue is now fixed. The issue was actually that the pmpro_pages array often has blanks in there, which would pass a '' value to the `post__no_in` query var. This would break the query. We now use array_filter to remove the blanks.

I've changed the for loop that set the my_pages var into a single query.

Some things left to look into:
* We can use static vars to cache the my_pages and hidden_pages vars. Then, if these are set, we don't have to run queries to find them again. This will reduce DB calls and help performance. These values are current_user dependent and shouldn't affect cases multiple queries unless people are getting clever with the pmpro_search_filter_post_types filter... so we could even hash that value to use as a key when caching the vars.
* We should double check the "make this work" section. Is this really the best way to do this.

We still have an edge case where member content that isn't restricted by the memberships_pages table or by category or taxonomy (i.e. restricted by custom code) will still show up in the search results. I think this is okay. You can use a post search filter to filter that stuff out.

ideadude:
Latest commit rewrites the "make it work" stuff, removes the odd string replace in the query in favor of some set math and some big queries to get post ids and set the `post__not_in` value of the query.

This seems to work for archives, searches, categories, tags and even custom taxonomies if you add the taxonomy's post_type through the `pmpro_search_filter_post_types` filter.

To test...

Setup:
* Add 1 level.
* Add a post category (members) and set it to require the level.
* Add a tag (members) and set it to require the level.
* Make sure the advanced setting "Filter restricted posts from searches and archives" is set to Yes.
* It helps to set the show excerpts advanced setting to Yes.

Create some posts:
* Create a post called "Restricted by level". Require the level via the meta box. Add some lorem ipsum.
* Create a post called "Restricted by category". Set the category. Add some lorem ipsum.
* Create a post called "Restricted by tag". Set the tag. Add some lorem ipsum.

Login as a member and load up these pages in tabs:
* /category/members/
* /tag/members/
* /?s=lorem+ipsum

Open an incognito window and stay logged out and view the same 3 tabs.

The logged in member should see all of the restricted posts in the archives and searches. The logged out users should not see the posts at all.

Testing that WooCommerce product cats aren't affected.
* Install WooCommerce
* Create a product category (lorem) - it is NOT set to require membership for this testing.
* Create a product and put it in the "lorem" category.
* visit /product-category/lorem/. You should see the products.
* visit /?s=lorem while logged in or logged out. You should see the products.

Testing CPTs
* Install PMPro CPT. Set it up to restrict WC products.
* Edit a product and mark it to require membership.
* Visit the 3 URLs as a member and non member. Non members should not see the products that require membership.

In theory if you set the product category to be restricted and added the "product" CPT to the post types in the search filter, then products in a restricted category would also not be visible to non members. That takes some custom code to test. Might be worth it.

Testing the use case the "make it work" stuff was written for.
* Keep the above level, category, tag, post setup.
* Add a new level 2.
* Add a new category "level2" that requires level 2.
* Update the "members" category to also check level 2. So this requires level 1 or 2.
* Add a post called "Restricted by multiple categories"
* Add this post to both the members and level2 categories.

So now, the old old logic would say "hey you're not in level 2" so hide this, but "your in level 1" so you should have access to this post. The "make it work" SQL search and replace would fix this for categories but not tags or other taxonomies. This new code should work... in that the new "Restricted by multiple categories" post will show up in the archives and searches for folks with level 1 OR level 2. It should be hidden from folks without any of those levels.

Left to do: Optimize this for performance. Specifically, I think we should cache the $hidden_post_ids values (which are all set based on the current user) and if that is already set, just use that value instead of building it again. This will avoid these expensive queries.

I will also test these queries on the live PMPro site with many thousands of posts and users and meta, to make sure the queries are working within our CPU and memory limits.

Other questions:
* Kim mentioned that the blog index is "kind of an archive"... how does this function with respect to the blog index? I don't know. Worth thinking about and testing.

includes/content.php
```diff
@@ -179,96 +191,139 @@ function pmpro_search_filter($query)
...
-    }
+	/**
+	 * Okay. We're going to filter the search results.
+	 * Some explaination of what we're doing.
+	 * A = All posts hidden by level.
+	 * B = All posts the current user has access to by level.
+	 * C = All posts hidden by category.
+	 * D = All posts the current user has access to by category.
+	 * 
+	 * Then hidden posts = (A - B) + (C - D)
```
dparker1005, line 239: Should this be `(A + C) - (B + D)`? In other words `(All restricted posts) - (All restricted posts that the user has access to)`?

Imagine a case where a post is restricted to level `1` but also in a restricted category `restricted_cat`. If a user had level 1 but did not have access to posts in `restricted_cat`, would we want to show this post?

If we flip the situation and the user has access to `restricted_cat` but they don't have level `1`, would we want to show the post now?

includes/content.php
```diff
@@ -179,96 +191,139 @@ function pmpro_search_filter($query)
...
+	if( ! empty( $level_ids ) ) {
+		// All hidden pages, minus the ones I have access to.
+		$sql = "SELECT mp.page_id
+				FROM {$wpdb->pmpro_memberships_pages} mp
+					LEFT JOIN {$wpdb->posts} p ON mp.page_id = p.ID
+				WHERE mp.page_id NOT IN(
+					SELECT page_id
+					FROM {$wpdb->pmpro_memberships_pages}
+					WHERE membership_id IN(" . implode(',', array_map( 'esc_sql', $level_ids ) ) . ")
+				) AND p.post_type IN('" . implode( "', '", array_map( 'esc_sql', $pmpro_search_filter_post_types ) ) . "')";
```
dparker1005, line 273: Would `SELECT DISTINCT` be more performant than just using `SELECT` and then calling `array_unique`, or is the difference minimal?

includes/content.php
```diff
@@ -179,96 +191,139 @@ function pmpro_search_filter($query)
...
+					LEFT JOIN {$wpdb->posts} p ON tr.object_id = p.ID
+				WHERE tr.term_taxonomy_id IN(
+					SELECT category_id
+					FROM {$wpdb->pmpro_memberships_categories}
+				) AND p.post_type IN('" . implode( "', '", array_map( 'esc_sql', $pmpro_search_filter_post_types ) ) . "')";
+	}
+	$cat_hidden_post_ids = array_values(array_unique($wpdb->get_col($sql)));
+
+	// Merge the hidden post IDs.
+	$hidden_post_ids = array_merge( $hidden_post_ids, $cat_hidden_post_ids );
```
dparker1005, line 321: Do we want to `array_unique()` this or does it not matter?

includes/content.php
ideadude: You're right. I was trying to get this done in 2 queries vs 4. I think we might have to do 4 queries... or (A - B - D) + (C - B - D) which would be 2 queries. Need to try that out and see how complicated those queries get.

includes/content.php
ideadude: Good idea. In general, it's better to do as much as possible in the query.

includes/content.php
ideadude: I think we should use array_unique here. It might not matter, but feels cleaner and helps with any other array stuff that might be done later by others filtering the query.

includes/content.php
ideadude: Latest commit does it in 4 separate queries now. I tried doing just 2 with the (A +C) - (B + D) approach, but this requires joining the relatively smaller pmpro tables with the much larger core WP tables in ways I suspect would make performance worse in a larger % of cases.

Note also that the latest version of the code swaps the B & C queries/arrays with what is explained here.

ideadude:
The latest version of the code uses a 4 query approach and caches the hidden post ids into a static variable to avoid duplicate queries.

Performance could be an issue on sites with a large number of levels, posts, taxonomies, etc. I'm going to test this on our staging site, which should give us a good idea of the potential speed and memory hit.

This feature can be turned off, so we don't need it to be perfect, but we do want it to work well and fast in the majority of cases.

ideadude:
The queries on my dev run in about 1/1000th of a second.

The 2 queries to get restricted posts run in about 2/100 seconds on the PMPro staging site, which has 469994 posts in the DB.

If there is a performance issue, it is likely to scale linearly.

We are also concerned about memory issues if we cache a really big array in the case of sites with lots of restricted posts. A site with 10k restricted posts might need about 200-300kbs to store the array, which also feels okay to ship for me. We could always make improvements to this, but again, very very big sites can turn this feature off or custom code their own search/archive filters.

This feels okay to ship.

Issue #2485, opened 2023-05-26
Allow HTML content in excerpts for membership-restricted content


ipokkel:
**Is your feature request related to a problem? Please describe.**
When a post contains an HTML block, (e.g. a Cover block created with the default WordPress block editor), in the excerpt the HTML is stripped and the excerpt displayed for non-members is not displaying as expected.

**Describe the solution you'd like**
As blocks are being used for content and theme building more frequently displaying the excerpt for non-members with the HTML intact would be a valued feature.

**Screenshots**
(Default Twenty Twenty-Three WordPress Theme)

### Current

<img width="691" alt="2023-05-26_09-52-21" src="https://github.com/strangerstudios/paid-memberships-pro/assets/9744519/447487aa-df94-4cf3-ab0b-81cc5bcc2c15">

### Preferred
<img width="682" alt="2023-05-26_08-54-57" src="https://github.com/strangerstudios/paid-memberships-pro/assets/9744519/fc5d79b6-6607-4a93-94a3-07cadc5bc3f2">

**Describe alternatives you've considered**
Unhooking the `pmpro_membership_content_filter` and hooking in a custom callback based on the `pmpro_membership_content_filter` function with the [$content = wp_strip_all_tags( $content );](https://github.com/strangerstudios/paid-memberships-pro/blob/a45029600127ebe14507bd216b3acda38a95deef/includes/content.php#L348) line removed.

**Additional context**
Moderators: 497573

MaximilianoRicoTabo:
hey @ipokkel submitted a PR to address this request. It's not the same approach you proposed but it works and it's flexible. Give it a look if you want and have time. Thanks

andrewlimaza:
I'm unable to replicate this with the Block Editor.

Steps I took:
1. Created a Cover block and uploaded an image and overlayed text to it.
2. Added some wording below the Cover block.
3. Added a --more-- tag below these two points.
4. Restricted the post to require all memberships
5. Viewed this post as a non-member. 

<img width="725" alt="Screenshot 2023-11-21 at 14 15 41" src="https://github.com/strangerstudios/paid-memberships-pro/assets/12629136/651265b3-1eea-4303-ba22-ec7aa234ac60">

andrewlimaza:
Upon further testing I have been able to replicate this:

Using the "Read More" block renders content correctly. If you do not use this, and automatically let Paid Memberships Pro generate/filter the excerpt it won't show correctly.

[...]

PR #2754, opened 2024-01-09
Use kses_post instead of strip_all_tags

andrewlimaza:
* ENHANCEMENT: Allow Blocks to render in the excerpt automatically.

Resolves: https://github.com/strangerstudios/paid-memberships-pro/issues/2485


ideadude:
There are a couple reasons we removed tags here.

Primarily, when we trim the excerpt a couple lines down, we will very often break the HTML by e.g. removing a closing div or p tag that is started in the early part of the excerpt. We might even break something in the middle of an HTML tag.

Second, the HTML tags count toward the word length when splitting up the content. So something with a lot of HTML tags will come out shorter than something with none.

What we'd really like is a system that somehow counted and split by the visible number of "words" and something that preserves HTML that starts and finishes inside of the excerpt. I unless there are reliably algorithms that do these things, I think these are challenging problems.

Issue #2690, opened 2023-11-28
Endless redirect on a test server without secure https protocol


drmanfred:
**Describe the bug**
I get endless redirects on my test server when the paid-membership-pro plugin is active. 
On this server there is no certificate installed. Therefore access is only possible via http protocol.

**To Reproduce**
Install the plugin in a wordpress environment on a test server without secure https protocol and activate it.

**Isolating the problem (mark completed items with an [x]):**
I debugged that behavior and found that the endless redirect can be stopped with the following change on line 127 in the file /include/https.php:

if( ! str_ends_with(strtolower( $https ), 'off') && strtolower( $https ) != 'false' && is_admin() ) {

In my case the server sends in the $_SERVER['HTTPS'] variable not just 'off' but puts some characters in front of it. So the condition in the original file always meets.

andrewlimaza:
Hi @drmanfred 

Thank you for reaching out. We are looking into this functionality further.

This happens when you set the "Force SSL" option under the "Payment Gateways & SSL Settings" is set to True. Please can you try and set it to not force SSL redirect.

drmanfred:
Hi Andrew,
thank you for looking into this issue.
The "Force SSL" option you mentioned is set to False in my settings.

When I examined the endless redirect, I found that the if-statement in line 127 in /include/https.php doesn't work the way it should. It never skips the redirect in the if-clause even when the $_SERVER['HTTPS'] is set to 'off'.
The statement in line 126: "$https = esc_url_raw( $_SERVER['HTTPS'] );" uses the WordPress function "esc_url_raw()". So the variable "$https" contains the string: "http://off" in case the $_SERVER['HTTPS'] is set to "off". So the condition in that if-clause is always true because "http://off" is not equal "off"

andrewlimaza:
Thanks for the detailed feedback, it helps. I am going to dig into this deeper and will reference this issue once a PR is available or reply if there's further feedback.

andrewlimaza:
Thanks @drmanfred I think it's okay to remove the `esc_url` from the `$https` variable as we are not storing the value anywhere and doesn't need to be escaped. That should resolve the issue of the conditional not working.

drmanfred:
Yes, you are right. It works.

PR #2755, opened 2024-01-09
Remove esc_url for redirecting to HTTPS

andrewlimaza:
* BUG FIX: Fixed an issue of infinite redirect loop in certain cases.

Resolves: https://github.com/strangerstudios/paid-memberships-pro/issues/2690


andrewlimaza:
Removed the `esc_url` function from the data as it was setting the value to `http://off` and the conditional was failing.

See issue: https://github.com/strangerstudios/paid-memberships-pro/issues/2690 for full discussion.

PR #2760, opened 2024-01-10
Revert "Now creating subscription when constructing orders"

dparker1005:
Reverts strangerstudios/paid-memberships-pro#2746

This causes an infinite loop with PayPal Express checkout and any other case where subscription orders are being loaded as the subscription is being created.


kimcoleman:
Merging - do we need something "like this" but approached differently?

PR #2767, opened 2024-01-16
Adding filter `pmpro_member_edit_memberships_panel_memberships_enddate_text`

dparker1005:
### All Submissions:

Adding filter `pmpro_member_edit_memberships_panel_memberships_enddate_text` to allow filtering expiration date shown on edit member memberships page.

### Other information:

### Changelog entry

### DIFF
--- a/adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php
+++ b/adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php
@@ -138,7 +138,7 @@ class PMPro_Member_Edit_Panel_Memberships extends PMPro_Member_Edit_Panel {
 								<?php
 									// Get the expiration date to show for this level.
 									$enddate_to_show = $shown_level->enddate;
-									$enddate_text = __( 'Never', 'paid-memberships-pro' );
+									$enddate_text = esc_html__( 'Never', 'paid-memberships-pro' );
 									if ( ! empty( $enddate_to_show ) ) {
 										$enddate_text = sprintf(
 											// translators: %1$s is the date and %2$s is the time.
--- a/adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php
+++ b/adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php
@@ -138,16 +138,26 @@ class PMPro_Member_Edit_Panel_Memberships extends PMPro_Member_Edit_Panel {
 								<?php
 									// Get the expiration date to show for this level.
 									$enddate_to_show = $shown_level->enddate;
-									if ( empty( $enddate_to_show ) ) {
-										esc_html_e( 'Never', 'paid-memberships-pro' );
-									} else {
-										echo esc_html( sprintf(
+									$enddate_text = __( 'Never', 'paid-memberships-pro' );
+									if ( ! empty( $enddate_to_show ) ) {
+										$enddate_text = sprintf(
 											// translators: %1$s is the date and %2$s is the time.
 											__( '%1$s at %2$s', 'paid-memberships-pro' ),
-											esc_html( date_i18n( get_option( 'date_format'), $enddate_to_show ) ),
-											esc_html( date_i18n( get_option( 'time_format'), $enddate_to_show ) )
-										) );
+											date_i18n( get_option( 'date_format'), $enddate_to_show ),
+											date_i18n( get_option( 'time_format'), $enddate_to_show )
+										);
 									}
+									/**
+									 * Filter the expiration date text to show for this level.
+									 *
+									 * @since 3.0
+									 *
+									 * @param string $enddate_text The expiration date text to show for this level.
+									 * @param WP_User $user The user object.
+									 * @param object $shown_level The level object.
+									 */
+									$enddate_text = apply_filters( 'pmpro_member_edit_memberships_panel_memberships_enddate_text', $enddate_text, $user, $shown_level );
+									echo esc_html( $enddate_text );
 								?>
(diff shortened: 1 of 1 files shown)


adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php
```diff
@@ -138,16 +138,26 @@ protected function display_panel_contents() {
 								<?php
 									// Get the expiration date to show for this level.
 									$enddate_to_show = $shown_level->enddate;
-									if ( empty( $enddate_to_show ) ) {
-										esc_html_e( 'Never', 'paid-memberships-pro' );
-									} else {
-										echo esc_html( sprintf(
+									$enddate_text = __( 'Never', 'paid-memberships-pro' );
```
ideadude, line 141:
```suggestion
									$enddate_text = esc_html__( 'Never', 'paid-memberships-pro' );
```

adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php
```diff
@@ -138,16 +138,26 @@ protected function display_panel_contents() {
...
 									$enddate_to_show = $shown_level->enddate;
-									if ( empty( $enddate_to_show ) ) {
-										esc_html_e( 'Never', 'paid-memberships-pro' );
-									} else {
-										echo esc_html( sprintf(
+									$enddate_text = esc_html__( 'Never', 'paid-memberships-pro' );
+									if ( ! empty( $enddate_to_show ) ) {
+										$enddate_text = sprintf(
 											// translators: %1$s is the date and %2$s is the time.
 											__( '%1$s at %2$s', 'paid-memberships-pro' ),
```
ideadude, line 145:
```suggestion
											esc_html__( '%1$s at %2$s', 'paid-memberships-pro' ),
```

ideadude: Let's fix this one while we're at it too.

ideadude:
From our Slack notes:

> Ok I think the “gold standard” that we shoot for now is one function pmpro_get_membership_expiration_text( $user_id, $level ) that:
> * Applies pmpro_memberslist_expires_column filter on backend for backwards compatibility
> * Applies pmpro_account_membership_expiration_text filter on frontend for backwards compatibility
> * Applies a new filter pmpro_membership_expiration_text which is the new recommended filter to use

ideadude:
Hmm. Here's a stab at the pmpro_get_member_expiration_text function that I would put into functions.php:

I tweaked the parameters from the suggestion. It seems we might want to call this without a user_id. This is problematic for the `pmpro_memberslist_expires_column` filter, since that expects a user object. We pass null if no user is sent in. It's typical to default to the current user, but that felt more dangerous here.

Updating the new memberships panel to use this function is straightforward.

However, when I went to update the members list table class, members list CSV export, and account page to use the new function, it felt very awkward because we have different values available in each time. The admin pages don't have a $level object handy. We have the "enddate" from the query results. And we could pass just the enddate or make a kind of fake level to pass into the function, but that will cause confusion.

We also probably don't want to do anything that add extra queries to set up a level object just in case things get filtered. We purposely make the member list table queries as light as possible.

So maybe we need to rethink if we can combine these and if so, how. Maybe we just pass the the enddate as a string. Maybe we have level_id be an optional value as well. If someone has the user_id and level_id, they can get the user's membership data back if they need to. E.g. the proration plugin will need the level id and user id so it can look for a delayed downgrade to show info for here. Otherwise, you can just filter "Never" into what you want to do.

````
/**
 * Get the text to display a membership's expiration date.
 * @since 3.0
 * @param object $level The level object to get the enddate from.
 * @param int $user_id The user ID this level is for. Passed to filters.		
 * @return string The expiration date text.
 */
function pmpro_get_membership_expiration_text( $level, $user_id = null ) {
	// Make sure a level was passed in.
	if ( ! isset( $level ) || ! is_object( $level ) ) {
		return false;
	}
	
	$enddate = $level->enddate;
	if ( empty( $enddate ) ) {
		$text = esc_html__( 'Never', 'paid-memberships-pro' );
	} else {
		$text = sprintf(
			// translators: %1$s is the date and %2$s is the time.
			esc_html__( '%1$s at %2$s', 'paid-memberships-pro' ),
			date_i18n( get_option( 'date_format'), $enddate ),
			date_i18n( get_option( 'time_format'), $enddate )
		);
	}

	// Apply the pmpro_memberslist_expires_column filter on backend for backwards compatibility.
	if ( is_admin() ) {		
		// This filter expects a user object as the 1 parameter.
		if ( ! empty( $user_id ) ) {
			$user = get_userdata( $user_id );
		} else {
			$user = null;
		}
		$text = apply_filters( 'pmpro_memberslist_expires_column', $text, $user );
	}

	// Apply the pmpro_account_membership_expiration_text filter on frontend for backwards compatibility.
	if ( ! is_admin() ) {
		$text = apply_filters( 'pmpro_account_membership_expiration_text', $text, $level );
	}

	/**
	 * Filter the expiration date text to show for this level.
	 *
	 * @since 3.0
	 *
	 * @param string $text The expiration date text to show for this level.
	 * @param object $level The level object we're getting the enddate for.
	 * @param int $user_id The user ID this level is for.
	 */
	$text = apply_filters( 'pmpro_member_edit_memberships_panel_memberships_enddate_text', $text, $level, $user_id );
	
	return $text;
}
````

ideadude:
Alternative PR here: https://github.com/strangerstudios/paid-memberships-pro/pull/2796

PR #2768, opened 2024-01-16
Actually setting hidden columns on Members List Table and storing for the user

kimcoleman:
### All Submissions:

We wanted to change what columns were hidden by default on the Members List Table. It turns out our code to hide columns wasn't really being respected.

This change makes it so that any new user with access to view this screen has their screen option preferences set on the first load. This is necessary so that we can properly reflect what columns are disabled by default in the Screen Options dropdown panel.

The current columns we hide by default are:
'ID',
'first_name',
'last_name',
'address',
'membership_id',
'joindate',

I used this XWP plugin repo as inspiration here: https://github.com/xwp/stream/blob/develop/classes/class-list-table.php

To test this as a fresh visitor, you must delete this key for your admin user in wp_usermeta: 	managememberships_page_pmpro-memberslistcolumnshidden.

### Other information:

### DIFF
--- a/classes/class-pmpro-members-list-table.php
+++ b/classes/class-pmpro-members-list-table.php
@@ -36,6 +36,21 @@ class PMPro_Members_List_Table extends WP_List_Table {
 				// If true, the parent class will call the _js_vars() method in the footer
 			)
 		);
+
+		// Check for default hidden columns.
+		$this->get_hidden_columns();
+
+		add_filter(
+			'screen_settings',
+			array(
+				$this,
+				'screen_controls',
+			),
+			10,
+			2
+		);
+
+		set_screen_options();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


classes/class-pmpro-members-list-table.php
```diff
@@ -36,6 +36,21 @@ public function __construct() {
 				// If true, the parent class will call the _js_vars() method in the footer
 			)
 		);
+
+		// Check for default hidden columns.
+		$this->get_hidden_columns();
```
dparker1005, line 41: Is this line necessary? `get_hidden_columns()` doesn't look like it does anything besides for return the array of hidden columns, which we don't use here.

ideadude: Technically it would save the user meta if there was nothing there already too. But the method gets called during the prepare_items() method as well, which seems enough to set the defaults on fresh loads (no user meta) and to save/check/uncheck items.

I'm removing it. Nice catch.

PR #2771, opened 2024-01-17
Adjusting capability checks throughout to hide elements where the cap is not present

kimcoleman:
### All Submissions:

This is the start of better capability checks before we render buttons that do things.

On the adminpages/dashboard.php, we should do a bit more here to adjust who sees the "License" section. We need something else to replace it with or that entire "welcome" box could become separate boxes/collapse without that section (CSS fix).

The member-edit and panels have more specific capability checks we need to have a discussion for.

### DIFF
--- a/adminpages/dashboard.php
+++ b/adminpages/dashboard.php
@@ -14,41 +14,51 @@ add_meta_box(
 	'toplevel_page_pmpro-dashboard',
 	'normal'
 );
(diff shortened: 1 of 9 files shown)


ideadude:
Great. Agreed RE the license stuff. People want to hide this. For now they can use custom CSS in the admin.

PR #2773, opened 2024-01-18
Adding a nonce to pmpro_update_level_order()

dparker1005:
### All Submissions:

Adding a nonce to `pmpro_update_level_order()` to prevent CSRF attacks.

### Other information:

### Changelog entry

### DIFF
--- a/adminpages/membershiplevels.php
+++ b/adminpages/membershiplevels.php
@@ -1026,7 +1026,8 @@
 
 		                data = {
 		                    action: 'pmpro_update_level_order',
-		                    level_order: level_order
+		                    level_order: level_order,
+							nonce: '<?php echo wp_create_nonce( 'pmpro_update_level_order' ); ?>'
 		                };
 
 		                $.post(ajaxurl, data, function(response) {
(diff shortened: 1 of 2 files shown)


ideadude:
Merged this.

To be safe we should also run the nonces through sanitize_key before checking them, and use esc_attr or esc_html/etc when echoing them like this.

https://github.com/strangerstudios/paid-memberships-pro/commit/90d80c4e651434b4727f8a9458ca1b95142e670e

Took the opportunity to sanitize nonce parameters in a few more places. https://github.com/strangerstudios/paid-memberships-pro/commit/0830a0d249f166480e8b07e5f4b43d91c6d85f44

I didn't find any cases where we weren't escaping them.

Reminder we want to sanitize as early as possible, and it's required if we ever send the $_REQUEST/GET/POST param to another function, even if it's a core PHP function. We can never know what is going to happen to the param after calling the function. If we're just doing an isset/empty/== check, we don't need to sanitize it.

I also checked if any of the other calls in includes/services.php needed nonces. I don't think so. They were either read only or webhooks that have their own systems for verifying the requests.

PR #2774, opened 2024-01-18
 *  Updates the payment description in Stripe when purchasing a subsc…

### LINKED ISSUE #2770: Customize payment description in Stripe when using Stripe Checkout
When you view a payment in Stripe, we customize the payment description when using on-site Stripe payments, but not when using Stripe Checkout.

There is a gist here with the general steps needed to customize Stripe Checkout payment descriptions:
https://gist.github.com/dparker1005/9e69961ac7f70b2a3daf1afcd6cc7c14

The ideal solution is to take that logic and integrate it directly into the logic that builds the Stripe Checkout Session object. The description for Stripe Checkout payments should be identical to the descriptions created for on-site Stripe payments.

MaximilianoRicoTabo:
…ription using Stripe Checkout.

### All Submissions:

Resolves #2770 

### Other information:

### Changelog entry

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -5311,7 +5311,7 @@ class PMProGateway_stripe extends PMProGateway {
 	 * return string The description to use for the Stripe request.
 	 * @since TBD
 	 */
-	public static function get_order_description( $morder ) {
+	private static function get_order_description( $morder ) {
 		//Bail if no order.
 		if ( empty( $morder ) ) {
 			return '';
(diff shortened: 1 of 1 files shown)


ideadude:
Good start. Let's go for the "ideal version". As David said in the issue:

> The ideal solution is to take that logic and integrate it directly into the logic that builds the Stripe Checkout Session object. The description for Stripe Checkout payments should be identical to the descriptions created for on-site Stripe payments.

Here is where the description is set for onsite orders: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/gateways/class.pmprogateway_stripe.php#L4399

And another one: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/classes/gateways/class.pmprogateway_stripe.php#L4634

One thing we like to avoid is having the same filter used in 2 different places... updating Stripe Checkout to use this filter would make it 3-4 places.

So I suggest we add a get_order_description method that figures out the description, filters it, and returns it. Then we use that in those 2 lines above and the new lines edited by this PR.

Let's confirm with David that this is a good idea.

dparker1005:
I agree, abstracting logic to build the order description is ideal

PR #2780, opened 2024-01-23
Avoid user meta joins in member searches on big sites.

ideadude:
### All Submissions:

On sites with a large numbers of users and/or user meta, members list searches could take a LONG time. The culprit is the join on the user meta table and the checks against user meta in the where clause.

To help with this, we already detect when there is a colon in the search query, and in those cases presearch for a specific user field or meta key. However "non-colon" searches could take a long time or even cause resource issues on the server.

This PR adds a check using the new `wp_is_large_user_count()` added in WP 6.0.0. If this value is true, then a simpler version of the search is done that doesn't join on the user meta table or search any user meta fields.

By default, sites with over 10k users are considered large. This can be adjusted using the `wp_is_large_user_count ` filter. https://developer.wordpress.org/reference/functions/wp_is_large_user_count/

We could consider adding a notice in these cases where the member search query is altered... or even perhaps just suggesting searching with colons, since the colon searches are a hidden feature as is.

1. Create a site with > 10k users.
2. User a tool like the query monitor plugin to track how long it takes to do a search on the members list table (without a colon in it).
3. Note that the search is faster (and doesn't search user meta) once this code is enabled.
4. Sites with < 4k users or with a : in them are unaffected.

### Other information:

### Changelog entry

> ENHANCEMENT: If `wp_is_large_user_count()` is true, searches on the members list table will use a simpler query that does not search user meta. You can still search user meta by using a query with a colon in it, e.g. meta_key:term or first_name: jason. By default WP considers a site with > 10k users "large". You can adjust this using the `wp_is_large_user_count` filter.

### DIFF
--- a/adminpages/memberslist-csv.php
+++ b/adminpages/memberslist-csv.php
@@ -174,6 +174,9 @@
 				}
 				$search = " AND u.ID IN(" . implode( ",", $user_ids ) . ") ";
 			}
+		} elseif( function_exists( 'wp_is_large_user_count' ) && wp_is_large_user_count() ) {
+			// Don't check user meta at all on big sites.
+			$search_query = " AND ( u.user_login LIKE '%" . esc_sql($s) . "%' OR u.user_email LIKE '%" . esc_sql($s) . "%' OR u.display_name LIKE '%" . esc_sql($s) . "%' ) ";
 		} else {
 			// Default search checks a few fields.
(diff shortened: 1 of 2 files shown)


ideadude:
Nice to have is to show some kind of message when this happens, but I'm not so sure people will really notice. We can add that enhancement later.

PR #2782, opened 2024-01-24
Improved capabilities for edit member page

dparker1005:
### All Submissions:

- Improved capabilities for edit member page
- Switched "other" tab to "TOS", only showing if TOS is enabled
- Fixed capabilities for user fields code and now showing group description on edit member tab

### Changelog entry

### DIFF
--- a/adminpages/member-edit/pmpro-class-member-edit-panel-user-fields.php
+++ b/adminpages/member-edit/pmpro-class-member-edit-panel-user-fields.php
@@ -7,13 +7,20 @@ class PMPro_Member_Edit_Panel_User_Fields extends PMPro_Member_Edit_Panel {
 	public function __construct( $field_group_name ) {
 		$this->slug = 'user_fields_' . sanitize_title( $field_group_name );
 		$this->title = $field_group_name;
-		$this->submit_text = __( 'Update Member', 'paid-memberships-pro' );
+		$this->submit_text = current_user_can( 'edit_users' ) ? __( 'Update Member', 'paid-memberships-pro' ) : '';
 	}
 
 	/**
 	 * Display the panel contents.
 	 */
 	protected function display_panel_contents() {
+		// Print the group description.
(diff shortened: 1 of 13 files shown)


adminpages/member-edit/pmpro-class-member-edit-panel-user-fields.php
```diff
@@ -7,20 +7,27 @@ class PMPro_Member_Edit_Panel_User_Fields extends PMPro_Member_Edit_Panel {
...
 	}
 
 	/**
 	 * Display the panel contents.
 	 */
 	protected function display_panel_contents() {
+		// Print the group description.
+		$field_group = pmpro_get_field_group_by_name( $this->title );
+		if ( ! empty( $field_group->description ) ) {
+			echo wp_kses_post( wpautop( $field_group->description ) );
```
kimcoleman, line 20: In core currently, we just use wp_kses_posts and don't wpautop. Do you think we should keep this the same as https://github.com/strangerstudios/paid-memberships-pro/blob/v3.0/includes/fields.php#L642 OR should we adjust that? I'm inclined to make sure they are the same.

ideadude:
Whenever we are using the same filter name in more than one place, we should abstract how that value is gotten (usually by creating a function or method to get it), put the filter in that one place, and then update all occurrences where the filter was used before to use the new function.

I did that here in this commit: https://github.com/strangerstudios/paid-memberships-pro/commit/603142778a09e22c63199c37a6363e91de0bc7ab

I created a pmpro_get_edit_member_capability() function with the pmpro_edit_member_capability filter in it.

I did not update the occurrences of the pmpro_edit_member_capability filter that were in deprecated functions. We generally don't want to update deprecated functions unless we really have to. Also, some of those occurrences were defaulting to the manage_options capability instead of the pmpro_edit_member capability, and I didn't want to change that behavior or account for it in our nice new clean pmpro_get_edit_member_capability() function.

PR #2792, opened 2024-01-30
Show links and previews when viewing a file field as readonly

ideadude:
### All Submissions:

PMPro 3.0 adds a read only view for the new edit member pages. When in this view, we noticed that file type user fields would only show the file name. In most scenarios, it makes sense for folks who are able to "edit_members" and otherwise view members to be able to view the files associated with that user. So we are updating the displayValue method to show previews and links for file type fields.

This should also make it so fields shown through the member shortcode are also shown in this way.

1. Create a user field of type file.
1. As an admin, edit a member to upload a file for them. Use an image to test previews.
2. Edit a user to have a role with the pmpro_edit_members cap but not the edit_users cap
3. Use that user to edit the membership for a member.
4. Go to the tab for the field group that has the file field. A preview and link should display if there is a file.

*Some other things to test:*
* A member shortcode that outputs a file field.
* Member Directory profiles for users with fields.
* Think of cases where file fields might have been output where we WOULDN'T want the link or preview to show.

### Other information:

### Changelog entry

> ENHANCEMENT: Now showing image previews and file links when file fields are viewed as readonly.

### DIFF
--- a/classes/class-pmpro-field.php
+++ b/classes/class-pmpro-field.php
@@ -982,12 +982,12 @@ class PMPro_Field {
 
             //setup date vars
             if(is_array($value) && !empty($value)){
-	    $value = strtotime(implode("/", $value), current_time('timestamp'));
-	}elseif(!is_array($value) && !empty($value)){
-	    $value = strtotime($value, current_time('timestamp'));
-	}else{
-	    $value = strtotime(date('Y-m-d'), current_time('timestamp'));
-	}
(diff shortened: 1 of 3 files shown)


kimcoleman:
If you use a shortcode like `[pmpro_member field="file_field"]`, for the user viewing their content on the frontend it looks like this:

Screenshot 2024-02-05 at 1 02 11 PM

My expectation would be that we show the file link or the image like we do in the admin.

PR #2796, opened 2024-01-31
Added pmpro_get_membership_expiration_text() function

ideadude:
This is an alternative to the PR here: https://github.com/strangerstudios/paid-memberships-pro/pull/2767

We are trying to allow new folks to filter the expiration date wherever it is shown using just one filter, while still supporting older versions of this filter which were used in a few different pages/scenarios.

### All Submissions:

> ENHANCEMENT: Added `pmpro_get_membership_expiration_text()` function and using it whenever an expiration date is shown. This can be filtered by the `pmpro_member_edit_memberships_panel_memberships_enddate_text` and also supports the old `pmpro_account_membership_expiration_text` and `pmpro_memberslist_expires_column` filters from older versions of PMPro.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -590,54 +590,93 @@ function pmpro_getLevelsExpiration( &$levels ) {
 
 /**
  * Get the text to display a membership's expiration date.
+ *
  * @since 3.0
- * @param int|string $enddate The expiration date timestamp or string to show. 
- * @param mixed $level A level object or id (optional) to pass to filters.
- * @param mixed $user A user object or id (optional) to pass to filters.
+ *
+ * @param object|int  $level The level object or ID to get the expiration date for.
+ * @param WP_User|int $user  The user object or ID to get the expiration date for.
+ *
  * @return string The expiration date text.
  */
-function pmpro_get_membership_expiration_text( $enddate, $level = null, $user = null ) {
-	/**
-	 * Filter to include the expiration time with expiration date
-	 * @param bool $pmpro_show_time_on_expiration_date Show the expiration time with expiration date
-	 * @param int $enddate The expiration date timestamp.
-	 * @param mixed $level A level object or id passed into this function.
-	 * @param mixed $user A user object or id passed into this function.
-	 * @return bool $pmpro_show_time_on_expiration_date Whether to show the expiration time with expiration date.
-	 * @since 3.0 Now passes the $enddate, $level, and $user.
-	 *
-	 * Used in adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php
-	 */
-	$show_time = apply_filters( 'pmpro_show_time_on_expiration_date', false, $enddate, $level, $user );
+function pmpro_get_membership_expiration_text( $level, $user ) {
+	// If a user ID was passed, get the user object.
+	if ( is_numeric( $user ) ) {
+		$user = get_userdata( $user );
+	}
 
-	if ( empty( $enddate ) ) {
-		// N/A on the backend. &#8212; dash on the frontend.
+	// Make sure that we have a user object.
+	if ( empty( $user ) || ! is_a( $user, 'WP_User' ) ) {
+		return '';
+	}
+
+	// If a level ID was passed, get the level object.
+	if ( is_numeric( $level ) ) {
+		pmpro_getSpecificMembershipLevelForUser( $user->ID, (int)$level );
+	}
+
+	// Make sure that we have a level object with an enddate property, even if it's empty.
+	if ( empty( $level ) || ! is_object( $level ) || ! isset( $level->enddate ) ) {
+		return '';
+	}
+
+	/**
+	 * Filter to include the expiration time with expiration date.
+	 * Used in adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php.
+	 *
+	 * @param bool $show_time Whether to show the expiration time with expiration date.
+	 *
+	 * @return bool Whether to show the expiration time with expiration date.
+	 */
+	$show_time = apply_filters( 'pmpro_show_time_on_expiration_date', false );
+
+	// Generate the expiration date text.
+	if ( empty( $level->enddate ) ) {
+		// If the level does not have an enddate, show a dash (&#8212;).
 		$text = esc_html_x( '&#8212;', 'A dash is shown when there is no expiration date.', 'paid-memberships-pro' );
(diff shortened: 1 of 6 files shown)


includes/functions.php
```diff
@@ -588,6 +588,67 @@ function pmpro_getLevelsExpiration( &$levels ) {
 	return $expiration_text;
 }
 
+/**
+ * Get the text to display a membership's expiration date.
+ * @since 3.0
+ * @param int $enddate The expiration date timestamp. 
+ * @param mixed $level A level object or id (optional) to pass to filters.
+ * @param mixed $user A user object or id (optional) to pass to filters.
```
dparker1005, line 596: If we want these parameters to be mixed, we should have some code to put both $level and $user into the same format before they are passed to filters. Custom code should know which format to expect data in. It would also be good to have both of these be required since we should always have a level and a user when this function is called.

If we were willing to switch both $level and $user to always be objects, we could also eliminate the $enddate parameter since we will have all the info needed to calculate $enddate from the passed $level and $user.

includes/functions.php
```diff
@@ -588,6 +588,67 @@ function pmpro_getLevelsExpiration( &$levels ) {
...
+		 * @param bool $pmpro_show_time_on_expiration_date Show the expiration time with expiration date
+		 * @param int $enddate The expiration date timestamp.
+		 * @param mixed $level A level object or id passed into this function.
+		 * @param mixed $user A user object or id passed into this function.
+		 * @return bool $pmpro_show_time_on_expiration_date Whether to show the expiration time with expiration date.
+		 * @since 3.0 Now passes the $enddate, $level, and $user.
+		 * 
+		 * Used in adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php
+		 */
+		$show_time = apply_filters( 'pmpro_show_time_on_expiration_date', false, $enddate, $level, $user );
```
dparker1005, line 614: The tough part here is that custom code and add ons filtering end date text would have to code support for this filter as well.

Instead of having this filter, could we instead use the new filter for this function whenever we want to change the format of the function's output?

includes/functions.php
```diff
@@ -588,6 +588,67 @@ function pmpro_getLevelsExpiration( &$levels ) {
...
+		if ( $show_time ) {
+			$text = sprintf(
+				// translators: %1$s is the date and %2$s is the time.
+				esc_html__( '%1$s at %2$s', 'paid-memberships-pro' ),
+				date_i18n( get_option( 'date_format'), $enddate ),
+				date_i18n( get_option( 'time_format'), $enddate )
+			);
+		} else {
+			$text = date_i18n( get_option( 'date_format' ), $enddate );
+		}
```
dparker1005, line 626: The tough part here is that custom code and add ons filtering end date text would have to code support for this filter as well.

Instead of having this filter and logic, could we instead use the new filter for this function whenever we want to change the format of the function's output?

includes/functions.php
```diff
@@ -588,6 +588,67 @@ function pmpro_getLevelsExpiration( &$levels ) {
...
+				date_i18n( get_option( 'time_format'), $enddate )
+			);
+		} else {
+			$text = date_i18n( get_option( 'date_format' ), $enddate );
+		}		
+	}
+
+	// Apply the pmpro_account_membership_expiration_text filter on frontend for backwards compatibility.
+	if ( ! is_admin() ) {
+		$text = '<p>' . $text . '</p>';	// The account shortcode expects this in a p tag.
```
dparker1005, line 631: If a filter is used later, the paragraph tags would be lost. Would it instead be better to wrap the call to pmpro_get_membership_expiration_text() in <p> tags, which would apply the tags after filters are run, instead of adding the <p> tags here before the filters?

includes/functions.php
```diff
@@ -588,6 +588,67 @@ function pmpro_getLevelsExpiration( &$levels ) {
...
+	}
+
+	/**
+	 * Filter the expiration date text to show for this level.
+	 *
+	 * @since 3.0
+	 *
+	 * @param string $text The expiration date text to show for this level.	
+	 */
+	$text = apply_filters( 'pmpro_member_edit_memberships_panel_memberships_enddate_text', $text, $level, $user );
```
dparker1005, line 647: Did we want to change this filter name to something more generic to be used everywhere? For example, just `pmpro_membership_expiration_text`?

shortcodes/pmpro_account.php
```diff
@@ -153,25 +153,12 @@ function pmpro_shortcode_account($atts, $content=null, $code="")
...
-											 * @return bool $pmpro_show_time_on_expiration_date Whether to show the expiration time with expiration date.
-											 *
-											 */
-											if ( apply_filters( 'pmpro_show_time_on_expiration_date', false ) ) {
-												$expiration_text .= ' ' . date_i18n( get_option( 'time_format', __( 'g:i a' ) ), $level->enddate );
-											}
+											$expiration_text = pmpro_get_membership_expiration_text( $level->enddate, $level, $current_user );
 										} else {
-											$expiration_text .= esc_html_x( '&#8212;', 'A dash is shown when there is no expiration date.', 'paid-memberships-pro' );
+											$expiration_text = pmpro_get_membership_expiration_text( esc_html_x( '&#8212;', 'A dash is shown when there is no expiration date.', 'paid-memberships-pro' ), $level, $current_user );
```
dparker1005, line 159: The doc block for `pmpro_get_membership_expiration_text()` currently says that the first parameter must be an int. A string is passed here.

Could we make an assumption that "Never" is shown for an empty expiration date in the Dashboard and `&#8212;` is shown on the frontend? If so, we can abstract this logic directly into the `pmpro_get_membership_expiration_text()` function when an empty enddate is passed.

includes/functions.php
ideadude: I thought of this, but it introduces memory and DB overhead in cases where we're trying to avoid it (e.g. members list exports).

includes/functions.php
ideadude: Nice. Moved the filter outside the conditional so it's always used set up. Also passing it to the main/bottom filter.

includes/functions.php
ideadude: Okay. We knew we'd update some filter users with this. Better to remove the jank and put the p tags outside the filter. Folks can use CSS to minimize the affect of the p tags.

includes/functions.php
ideadude: Nice catch. Went with pmpro_membership_expiration_text

PR #2797, opened 2024-01-31
Hide non-PMPro notices on PMPro settings pages.

ideadude:
### All Submissions:

Inspired by [this tweet](https://twitter.com/verygoodplugins/status/1751228984065446001) and other discussions, this PR removes non-PMPro notices from PMPro settings pages.

This PR needs some discussion.

1. Do we really want to do this?
2. Are we sure we aren't filtering out PMPro callbacks that might not start with pmpro_ or be part of a class that starts with PMPro_?
3. Should we only do this on certain pages?
4. Should we even hide some of our notices?
5. We had some stricter code already running on the wizard that could be merged with this new code: https://github.com/strangerstudios/paid-memberships-pro/blob/74d80891d63a3b65fccb448b52461e7c525b6132/includes/admin.php#L217-L226

> ENHANCEMENT: Hiding non-PMPro notices from PMPro settings pages.

### DIFF
--- a/includes/admin.php
+++ b/includes/admin.php
@@ -346,21 +346,32 @@ function pmpro_hide_non_pmpro_notices() {
 		// Loop through the callbacks and remove any that aren't PMPro.
 		foreach ($wp_filter[$hook]->callbacks as $priority => $callbacks) {
 			foreach ($callbacks as $key => $callback) {				
-				if ( is_array( $callback['function'] ) ) {					
-					// Skip if the class starts with pmpro_.
-					$class = get_class( $callback['function'][0] );
-					if ( strpos( strtolower( $class ), 'pmpro_' ) === 0 ) {
-						continue;
-					}
+				if ( is_string( $callback['function' ] ) ) {
+					// Check the function name.
+					// Ex. add_action( 'admin_notices', 'pmpro_admin_notice' );
+					$name_to_check = $callback['function'];
+				} elseif ( is_array( $callback['function' ] ) && is_string( $callback['function'][0] ) ) {
+					// Check the class name for the static method.
+					// Ex. add_action( 'admin_notices', array( 'PMPro_Admin', 'admin_notice' ) );
+					$name_to_check = $callback['function'][0];
+				} elseif ( is_array( $callback['function' ] ) && is_object( $callback['function'][0] ) ) {
+					// Check the class name for the non-static method.
+					// Ex. add_action( 'admin_notices', array( $some_object, 'admin_notice' ) );
+					$name_to_check = get_class( $callback['function'][0] );
 				} else {
-					// Skip if the callback function starts with pmpro_.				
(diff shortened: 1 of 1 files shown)


ideadude:
New commit has this checking both class names (for namespaced stuff, it checks the start of the namespace) that start with pmpro_ and function names that start with pmpro_. Both are case insensitive checks.

https://github.com/strangerstudios/paid-memberships-pro/pull/2797/commits/61cdb8a2bb43bd1dec0b4713b2b5ea11f0f52ae8

This will still stop a method like pmpro_something that is in a class that doesn't start with PMPro. Not sure if there are many of those in the wild that we should worry about.

We can tweak this and add exceptions as needed.

includes/admin.php
```diff
@@ -320,3 +320,49 @@ function pmpro_admin_footer_text( $text ) {
...
+					}
+				} else {
+					// Skip if the callback function starts with pmpro_.				
+					if ( strpos( strtolower( $callback['function'] ), 'pmpro_') === 0 ) {
+						continue;
+					}
+				}
+
+				// Probably not a PMPro notice, remove it.
+				unset($wp_filter[$hook]->callbacks[$priority][$key]);
```
dparker1005, line 363:
```suggestion
				if ( is_string( $callback['function' ] ) ) {
					// Check the function name.
					// Ex. add_action( 'admin_notices', 'pmpro_admin_notice' );
					$name_to_check = $callback['function'];
				} elseif ( is_array( $callback['function' ] ) && is_string( $callback['function'][0] ) ) {
					// Check the class name for the static method.
					// Ex. add_action( 'admin_notices', array( 'PMPro_Admin', 'admin_notice' ) );
					$name_to_check = $callback['function'][0];
				} elseif ( is_array( $callback['function' ] ) && is_object( $callback['function'][0] ) ) {
					// Check the class name for the non-static method.
					// Ex. add_action( 'admin_notices', array( $some_object, 'admin_notice' ) );
					$name_to_check = get_class( $callback['function'][0] );
				} else {
					// Ex. add_action( 'admin_notices', function() { echo 'Hello World'; } );
					// We don't use closures in PMPro, so we don't need to check for them.
					$name_to_check = '';
				}

				// Trim slashes for namespaces and lowercase the name.
				$name_to_check = strtolower( trim( $name_to_check, '\\' ) );

				// If the function name starts with 'pmpro', then we don't want to remove it.
				// Not checking for 'pmpro_' because we have class names like PMProGateway_stripe and want to keep notices from add ons.
				if ( strpos( $name_to_check, 'pmpro' ) !== 0 ) {
					unset( $wp_filter[$hook]->callbacks[$priority][$key] );
				}
```

dparker1005 (changes requested):
The current implementation throws an error when a static method is hooked or an anonymous function is hooked. The current code would also hide notifications from some PMPro classes (ex. `PMProGateway_stripe`) and PMPro Add Ons (ex. `pmproabc_`).

PR #2798, opened 2024-01-31
Fix/sub sync

ideadude:
Adding a supports() method to the base gateway class and stripe class that we can use on the subscriptions tab to figure out if we should show the sync button on subscriptions.

This method is nice because we can extend it in the future for other similar things to detect difference between the gateways.

If we like this approach:

1. We need to update the other gateways that support sub sync to have their own methods like Stripe does. (Braintree, Authorize.net, PayPal Express, others in core?, other gateways added through plugins? PayFast? PayStack?)
2. We should probably recode this similar check for payment method updates: https://github.com/strangerstudios/paid-memberships-pro/blob/v3.0/classes/gateways/class.pmprogateway_stripe.php#L2227-L2237

### DIFF
--- a/classes/gateways/class.pmprogateway.php
+++ b/classes/gateways/class.pmprogateway.php
@@ -1,7 +1,7 @@
 <?php	
 	//require_once(dirname(__FILE__) . "/class.pmprogateway.php");
 	class PMProGateway
-	{
+	{	
 		function __construct($gateway = NULL)
 		{
 			$this->gateway = $gateway;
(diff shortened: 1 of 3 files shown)


dparker1005:
An alternative solution could be to check if the child class overwrites the `update_subscription_info()` method?

For example, something like:
`
// If we can sync the subscription with the gateway, show the button.
$update_subscription_info_method = new ReflectionMethod( $subscription->get_gateway_object(), 'update_subscription_info' );
if ( $update_subscription_info_method->getDeclaringClass()->getName() !== 'PMProGateway' ) {
	// Show the button...
}
`

classes/gateways/class.pmprogateway.php
```diff
@@ -215,17 +215,18 @@ function getTransactionStatus(&$order)
...
-		 *
+		 * Check if the gateway supports a certain feature.
+		 * 
 		 * @since 3.0
-		 *
-		 * @return string|false 'individual' if the gateway allows for payment method updates for individual subscriptions, 
-		 *                      'all' if the gateway updates all subscriptions, or false if the gateway does not support payment method updates.
+		 * 
+		 * @param string $feature The feature to check for.
+		 * @return bool
```
dparker1005, line 226:
```suggestion
		 * @return bool|string Whether the gateway supports the requested. A string may be returned in cases where a feature has different types of support.
```

classes/gateways/class.pmprogateway.php
```diff
@@ -215,17 +215,18 @@ function getTransactionStatus(&$order)
...
 			//this looks different for each gateway, but generally an array of some sort
 			return array();
-		}
+		}		
 
 		/**
-		 * Returns whether the gateway allows for payment method updates.
-		 *
+		 * Check if the gateway supports a certain feature.
+		 *
```
dparker1005, line 222:
```suggestion
		 * 
		 * $feature options:
		 *    'subscription_sync' - Whether this gateway can pull subscription data from the gateway.
		  * 
```

PR #2807, opened 2024-02-02
Query args filtering simplified. isset() is enough to still support empty strings. Removed null !== comparison

mircobabini:
### All Submissions:

Doing a isset() test vs. null !== strict comparison has only one difference: isset does not error in case the array key is actually not set.

Avoid phpcs "Condition is always 'true' because 'isset( $args['user_id'] )' is already 'true' at this point".

Use the following snippet to see that it's the exactly same thing.

```php
$args = [];

var_dump( isset( $args['a'] ) );
var_dump( null !== @$args['a'] );
var_dump( isset( $args['a'] ) && null !== $args['a'] );

echo '--------' . PHP_EOL;

$args['a'] = '';

var_dump( isset( $args['a'] ) );
var_dump( null !== $args['a'] );
var_dump( isset( $args['a'] ) && null !== $args['a'] );

echo '--------' . PHP_EOL;

$args['a'] = 'test';

var_dump( isset( $args['a'] ) );
var_dump( null !== $args['a'] );
var_dump( isset( $args['a'] ) && null !== $args['a'] );
```

### Other information:

### DIFF
--- a/classes/class-pmpro-subscription.php
+++ b/classes/class-pmpro-subscription.php
@@ -336,13 +336,10 @@ class PMPro_Subscription {
 
 		/*
 		 * Now filter the query based on the arguments provided.
-		 *
-		 * isset( $arg ) && null !== $arg is meant to deal with $args['arg'] = null usage
-		 * while still supporting $args['arg'] = ''.
 		 */
 
 		// Filter by ID(s).
-		if ( isset( $args['id'] ) && null !== $args['id'] ) {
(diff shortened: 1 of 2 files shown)


ideadude:
Thanks. Makes for _even cleaner_ code.

I think the case the extra check was meant to account for was something like:
`$args = array( 'a' => null, 'b' => 'something' );`

I still get 3 falses when I do the same var dumps with a var setup that way in all versions of PHP I can test.

PR #2811, opened 2024-02-05
fix: use user ids resulting from custom query by pmpro_members_list_s…

rollsappletree:
…ql filter

### All Submissions:

Using `pmpro_members_list_sql` filtered query to actually generate the CSV to. 
I implemented a custom query (with other filters based on user meta) using the  pmpro_members_list_sql filter.
I can see the users by the filters I setted in the membership table via UI. If I click the export csv button, the code actually use the custom query to generate an array of users, and then it uses "between" with those array. 

If the users are not consecutives, the member UI data and the one extracted via CSV are not equal. 
With this change it'll be the same.

1. go to the membership page
2. create some new filters and implement it via `pmpro_members_list_sql`
3. try to download CSV

### Other information:

### Changelog entry


dparker1005:
This PR breaks the assumption that `$theusers` (an array of user IDs to output) is ordered. This may be the case if the `ORDER BY` statement is altered by the `pmpro_members_list_sql` filter to show data in a custom order.

To break the assumption that `$theusers` is in order, this PR switches from querying between a set of user IDs to splicing the `$theusers` array to get the specific user IDs to show during each iteration. This is much safer to avoid duplicate and missing records in cases where `pmpro_members_list_sql` is out of order.

PR #2814, opened 2024-02-06
Cast level_id to be integer in the pmpro_getLevelAtCheckout

andrewlimaza:
* BUG FIX: Fixes cases where passing a string, but valid, level_id to the `pmpro_getLevelAtCheckout` would return NULL.

The proposed fix is to case the level ID to integer as late as possible. This only occurs when you pass through the level ID as a string, a good example of this is the Advanced Levels Page Add On using this function and passing through a string.

### All Submissions:

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2427,8 +2427,11 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
 		$level_id = apply_filters( 'pmpro_default_level', intval( $default_level ) );
 	}
 
+	// Cast level ID to int to be safe.
+	$level_id = (int) $level_id;
+
 	// If we still don't have a level, bail.
-	if ( empty( $level_id ) || $level_id < 1 || ! is_int( $level_id ) ) {
+	if ( empty( $level_id ) || $level_id < 1 ) {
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -2427,8 +2427,11 @@ function pmpro_getLevelAtCheckout( $level_id = null, $discount_code = null ) {
 		$level_id = apply_filters( 'pmpro_default_level', intval( $default_level ) );
 	}
 
+	// Cast level ID to int to be safe.
+	$level_id = (int) $level_id;
+
 	// If we still don't have a level, bail.
-	if ( empty( $level_id ) || $level_id < 1 || ! is_int( $level_id ) ) {
+	if ( empty( $level_id ) || $level_id < 1 ) {
```
ideadude, line 2434: Remove the cast and comment above.

Even though something like "4" < 1 still evaluates to false and "0" < 1 evaluates to true in the versions of PHP I tested, it feels safe to future proof this against any strict typing coming down the pipe in PHP by casting to an in here in the check.

https://3v4l.org/NPn5Z

`if ( empty( $level_id ) || (int)$level_id < 1 ) {`

PR #2817, opened 2024-02-06
Fix/member shortcode

ideadude:
### All Submissions:

The [pmpro_member] shortcode is a powerful tool that can be used to display any user meta value for the current user or if a user_id is specified any user meta value for another user.

Because of this, the shortcode used improperly could allow any user able to edit a post or certain widgets to find potentially sensitive user meta for themselves or other users. This is a privacy issue and a security issue.

We considered creating a block list to stop certain fields from being displayed this way. This doesn't feel like a good complete solution as we can never tell what meta from other plugins or user fields would be deemed sensitive by the site owner. So inevitably some fields would be missed.

We considered creating an allow list, but that option does not feel great either. One of the use cases for this shortcode is to embed non-PMPro meta fields into pages. Adding the requirement to allow list the field through custom code or compatibility code in other plugins limits the usefulness of the shortcode and makes it confusing to use.

Instead, we have decided to take the approach similar to how WP core handles unsafe HTMl like script tags in WP posts. We hook in before any content is saved. Instead of checking the "unfiltered_html" capability, we check for the "edit_users" capability. If the current user saving the content does not have the "edit_users" capability, we strip any use of the shortcode from the content.

We feel the "edit_users" capability is the correct one to check here. By default, only administrators have this permission. Any user with this permission is potentially able to find this information through other means. Core WP developers have called the "edit_users" capability the "keys to the castle". It should be handled with care.

1. Log in as an admin. 
2. Edit a post and add a shortcode like [pmpro_member field="user_email"]. Save.
3. Note that the current user's email is shown when you view the page.
4. Change that post's author to a different user with the author role.
5. Log in as the author.
6. Edit the post with the shortcode or try to add a similar one to a post. Save.
7. Note that the shortcode has been removed.

### Other information:

To see if the pmpro_member shortcode is currently in use on your site, you can search for "[pmpro_member " (note the space at the end, which avoids including results for the pmpro_membership shortcode) from the post table of your admin dashboard. You should doublecheck any page using the shortcode to make sure that only approved data is being shown through the shortcode.

Some notes on related changes we are not including in this PR.
* We considered showing a note on the edit post page as a warning before or after saving a post with the shortcode added. This is very complicated to add since the block editor saves posts through Javascript and the Classic Editor (and every other page builder) handles things differently. We do not want to hinder performance by adding complicated checks. When WP core strips out script tags, there is no notice. We followed suit here. We may reconsider this if it seems users are often confused by this.
* We considered showing a notice if we found the shortcode in use in the site's posts. We feel this message has the potential to be overly alarming and confusing. Instead of an insite notification, we will include information about the vulnerability in the installation notes and related blog posts.

### Changelog entry

> SECURITY: Only users with the "edit_users" capability may add the pmpro_member shortcode to posts and widgets now. (Thanks, Scott Kingsley Clark.)


shortcodes/pmpro_member.php
```diff
@@ -153,3 +153,36 @@ function pmpro_member_shortcode($atts, $content=null, $code='')
...
+ * @param array $instance
+ * @return array
+ */
+function pmpro_maybe_strip_member_shortcode_from_widgets( $instance ) {
+	if ( ! current_user_can( 'edit_users' ) ) {
+		$instance = pmpro_strip_shortcode( 'pmpro_member', $instance );		
+	}
+	return $instance;
+}
+add_filter( 'widget_update_callback', 'pmpro_maybe_strip_member_shortcode_from_widgets' );
```
dparker1005, line 188: Since `pmpro_strip_shortcode()` is clever enough to take an array or a string, we could eliminate one of these functions and just move the `add_filter()` to point to the remaining one.

And then if we only have a single function that uses the `pmpro_strip_shortcode()` function, maybe the abstraction is unnecessary at that point and this entire PR could just be a single function specifically for the `[pmpro_member]` shortcode.

Neither change is necessary, but just wanted to put them out there in case you think it would be cleaner.

dparker1005 (approved):

PR #2825, opened 2024-02-08
Automatically select first edit member panel/tab

andrewlimaza:
* ENHANCEMENT: Automatically select the first Edit member panel and tab that's available.

This helps improve UX when only certain tabs are shown to a specific user that may not have full capabilities or shown certain tabs only in the edit member screen.

### All Submissions:

1. Give a non-admin role the ability to all `pmpro_` prefixed capabilites, make sure they don't have `edit_users` capability.
2. Filter the `pmpro_edit_member_capability` to be `edit_users` (Note: The Membership Manager Add On does this automatically).
3. While logged-in as this manager/non-admin edit a member and see that the Orders panel shows by default due to the `pmpro_orders` capability given in step 1 and is automatically "active".

### Other information:


ideadude:
As it is now, if you cannot edit users, you get a read only view of the top user info tab.

So we don't need this now, but might be handy in the future.

PR #2829, opened 2024-02-08
Adding PMPro update to fill data for all subscriptions

dparker1005:
### All Submissions:

Gives users an active step to pull in subscription data instead of having to wait for subs to be loaded organically.

Also fixes bugs in the default migration data code that may cause migrations to fail.

Resolves XXX.

1. Drop subscriptions table if you have one
2. Run 3.0 migration script
3. Got to PMPro dashboard
4. Click on the button to start the DB migration process
5. Wait as the subscription data is loaded

### Other information:

### Changelog entry


dparker1005:
The `'active' !== $this->status` check was removed so that cancelled subscriptions also have accurate end date and billing amount data pulled from them. This was originally disabled for performance reasons, but this migration tends to run fairly quickly. It is getting correct data for ALL subscriptions, active and cancelled.

The changes to stop using `pmpro_isLevelRecurring()` is because we don't want the results to be filtered. We want a level that has a billing amount and cycle number, that is it. `pmpro_isLevelRecurring()` also doesn't check for a cycle number.

The change to create a blank `$subscription_level` is to avoid infinite loops when a level no longer exists on a website. The level has `cycle_number` set to `1`,  which is what would break the infinite loop.

PR #2854, opened 2024-02-20
Fixing error handling on cancel page

dparker1005:
### All Submissions:

`! empty()` fixed warning when `$worked` is not defined.

`$pmpro_error` change avoids overwritting any error messages that may already be set.

`$_REQUEST['confirm']` change is needed to re-show the cancellation form on error.

### Other information:

### Changelog entry

### DIFF
--- a/adminpages/levels/delete-level.php
+++ b/adminpages/levels/delete-level.php
@@ -34,7 +34,7 @@ if($ml_id > 0) {
             // Couldn't delete the subscription or the membership.
             // We should probably notify the admin
             $pmproemail = new PMProEmail();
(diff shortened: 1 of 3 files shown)


dparker1005:
Recent change to stop checking `$pmpro_error` is because there are cases where a cancellation is still successful even though $pmpro_error is set. Ex if a subscription fails to sync with the gateway. In this case, we still try to cancel the membership, but also send an email to the admin.

dparker1005:
Recent change to no longer return `false` from `pmpro_cancelMembershipLevel()` on failed subscription cancel needed to make the return consistent with that function's docblock and also to avoid situations on the cancel page where the cancellation shows as failed even though the membership level was removed.

PR #2899, opened 2024-03-21
Remove deprecated code from update script.

kimcoleman:
This code was breaking when zeroing out the db version.

New installs won't have > 100 orders and so wouldn't have queued up the code.

There shouldn't be too many folks upgrading to 3.0 from 1.8.7 or earlier.

But some folks might zero out the DB version for various other reasons and we don't want to load up broken updates.

We should plan to figure out these old updates soon. It does slow down new installs a little bit and can just lead to odd issues.

For now, we can potentially just merge this and any others that break.


includes/updates/upgrade_1_8_8.php
```diff
@@ -16,94 +16,8 @@ function pmpro_upgrade_1_8_8() {
 	$sqlQuery = "UPDATE $wpdb->pmpro_membership_orders SET status = 'success' WHERE gateway = 'authorizenet' AND status = ''";
 	$wpdb->query($sqlQuery);
 	
-	//Fixing old $0 Stripe orders. (Sets up update via AJAX)
-	$orders = $wpdb->get_col("SELECT id FROM $wpdb->pmpro_membership_orders WHERE gateway = 'stripe' AND total = 0");			
-	if(!empty($orders))
-		pmpro_addUpdate('pmpro_upgrade_1_8_8_ajax');
+	// Removed the Stripe update, which relied on deprecated code.
```
ideadude, line 19:
```suggestion
	// Since 3.0: Removed the Stripe update, which relied on deprecated code.

```

PR #2908, opened 2024-03-26
Fixing expired/cancelled dates for old members in Members List

dparker1005:
### All Submissions:

Fixed issue where the expired/cancaellation date would be empty for old members in the Members List.

### Other information:

### Changelog entry


ideadude:
Is this a case where the pmpro_sanitize_with_safelist function could work? How does is that function handled by the Plugin Checker security check now. If we need to sanitize the input or add a comment to negate the scan, we kind of lose the efficiency of using our function. I still like the readability of using our function RE what is happening.

I'd be fine either way. Just wanted to call this out.

https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/functions.php#L3668-L3681

PR #2909, opened 2024-03-26
pmpro_no_quotes - Add null coalescing operator

ipokkel:
### All Submissions:

When null is passed to the `str_replace` arguments used in `pmpro_no_quotes` the follow error got logged in debug:
```
PHP Deprecated: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated
Resolves XXX.
```

Adding a null coalescing operator to the search string third argument returns an empty string which resolves this issue.

1. Activate the Custom Level Cost Text Add On, navigate to Memberships > Settings > Advanced and check the debug log for the deprecated warning.
2. Apply the patch, reload the page, and confirm that no new log entries were added.

### Other information:

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1978,7 +1978,7 @@ function pmpro_checkDiscountCode( $code, $level_id = null, $return_errors = fals
 }
 
 function pmpro_no_quotes( $s, $quotes = array( "'", '"' ) ) {
-	return str_replace( $quotes, '', $s );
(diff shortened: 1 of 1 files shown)


ideadude:
This works but would raise the PHP version requirement to 7.

Inline with our intention to only raise the version requirement when necessary, we can rewrite this to avoid the null in another way.

Something like `if ( empty( $s ) ) { $s = ''; }` above the return.

PR #2923, opened 2024-03-29
Adding setting to enable/disable custom page templates

dparker1005:
### All Submissions:

Screenshot 2024-03-29 at 8 40 35 AM

Added a setting to the Page Settings page that allows admins to choose to:
- Always use custom page templates
- Use custom page templates only if they are compatible with their current version of PMPro
- Not use custom page templates

By default, this is set to only using custom page templates that have versions that match the core page template versions. This should help to avoid issues where sites update to the newest version of PMPro and functionality on frontend pages breaks as a result.

For sites that are using a theme or 3rd party plugin that includes custom page templates, this can also protect them from authors that do not update their page templates or allow them to avoid using those templates for any other reason.

For sites that have built their own template and are ok being responsible for keeping them up to date, they can choose to always use page templates, regardless of version.

### Other information:

### Changelog entry


includes/page-templates.php
```diff
@@ -190,14 +218,23 @@ function pmpro_page_template_notices() {
...
 		?>
-		<div class="notice notice-warning">
-			<p>
-				<?php
-					esc_html_e( 'Outdated page templates have been detected in your theme or a plugin. You should update the following templates to ensure compatibility with the Paid Memberships Pro plugin:', 'paid-memberships-pro' );
-					echo '<tr>' . $outdated_templates_string. '</tr>'; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-				?>
-			</p>
+		<div class="notice notice-error pmpro_notification pmpro_notification-error">
+			<button type="button" class="pmpro-notice-button notice-dismiss" value="hide_pause_notification"><span class="screen-reader-text"><?php esc_html_e( 'Dismiss this notice.', 'paid-memberships-pro' ); ?></span></button>
```
kimcoleman, line 222: If we aren't going to allow dismiss in the banner, we should delete this close button.

kimcoleman:
Since this page can get long, we should duplicate the save settings button in all the section boxes. If we don't like that approach, we instead need to move the button outside the last box.

dparker1005:
Pushed those changes, thank you!

PR #2937, opened 2024-04-04
Using abstracted checkout completion function

dparker1005:
### All Submissions:

For the past year or so, we have used an abstracted "checkout completion" function to finish processing checkouts using the Stripe Checkout gateway. Now that this approach has been tested, this PR updates the PMPro checkout preheader code to use that abstracted (and renamed) function to use instead of having all the logic inline. The goal of this is to make the checkout flow more manageable to make a refactor easier down the road.

This also brings a couple of documentation updates and "hour" membership fixes from the inline preheader completion flow to the abstracted function.

There is also a bug where the $order->discount_code property is sometimes set to the discount code string instead of the discount code object. This could break some emails and is fixed in this PR.

Finally, the abstracted checkout completion function loses the logic to save billing address data in user meta, which is expected and the direction that we plan on moving.

Checkouts have been tested with the free gateway, on-site Stripe, Stripe Checkout, and PayPal Express. For the paid tests, a discount code was used with all of them.

### Other information:

### Changelog entry


includes/checkout.php
```diff
@@ -205,11 +226,6 @@ function pmpro_complete_async_checkout( $order ) {
 		'enddate'         => $enddate
 	);
 
-	global $pmpro_error;
```
ideadude, line 208: This seems like a good thing to remove. I can imagine it might fail for onsite checkouts where ipnlog is not defined. At the same time, was this at all useful for debugging ipn/webhook logs?

preheaders/checkout.php
```diff
@@ -562,242 +567,40 @@
...
+		$rurl = apply_filters( "pmpro_confirmation_url", $rurl, $user_id, $pmpro_level );
+		wp_redirect( $rurl );
+		exit;
+	} else {
 
-				//send email to admin
-				$pmproemail = new PMProEmail();
-				$pmproemail->sendCheckoutAdminEmail( $current_user, $order );
-			}
+		//uh oh. we charged them then the membership creation failed
```
ideadude, line 595: It's not clear to me here without looking inside pmpro_complete_checkout() that an error during "membership creation" is the only failure state.

This is all meant as an in progress refactor, so I don't think we need to change any of the code, but perhaps update the comment here to be more clear what pmpro_complete_checkout() is doing and why we are doing the tests we are below. i.e. if it returned false, we know user creation failed.

ideadude (commented):
Just some comments/questions. Great stuff. I'm doing a few more tests and will let you know how they go.

includes/checkout.php
dparker1005: The `ipnlog()` function was only ever used to log error messages for PayPal Standard checkout completions. I'm not concerned about removing it, and we can add it back to the PayPal Standard code specifically if needed.

preheaders/checkout.php
dparker1005: Users are always created before pmpro_complete_checkout() is called. pmpro_complete_chekcout() returns true if `pmpro_changeMembershipLevel( $custom_level, $order->user_id, 'changed' )` completes successfully, false otherwise. So in the pmpro_msg, "membership creation" is assigning a membership level. Maybe we need clearer wording in the error message

dparker1005: Maybe it would be better to say "Something went wrong while processing your checkout."

dparker1005: Improved the comment and error message

PR #2938, opened 2024-04-05
Default to "View: My Access"

andrewlimaza:
* ENHANCEMENT: Default to "View with my access" when no option is found. Fixes an issue where defaulting to View with access would confuse admins and not highlight the code green as the CSS class is missing.

### All Submissions:


dparker1005:
Just added some quick changes to enforce "if the setting is not `yes` or `no`, default to `current`" logic in case the value somehow gets set to something else

PR #2954, opened 2024-04-11
Additional email template variables for the expired memberships email

MaximilianoRicoTabo:
* Add membership id and membership level name to data array object.
 * Tweak WPCS and validations

image
image

### All Submissions:

Resolves XXX.

### Other information:

### Changelog entry

### DIFF
--- a/classes/class.pmproemail.php
+++ b/classes/class.pmproemail.php
@@ -1286,7 +1286,7 @@
 				'user_login' => $user->user_login,
 				'sitename' => get_option( 'blogname' ), 				
 				'membership_id' => $membership_level->id,
-				'membership_level_name' => $membership_level->name, 
+				'membership_level_name' => $membership_level->name,
 				'siteemail' => get_option( 'pmpro_from_email' ), 
 				'login_link' => pmpro_login_url(),
 				'login_url' => pmpro_login_url(),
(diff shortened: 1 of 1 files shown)


classes/class.pmproemail.php
```diff
@@ -1306,26 +1306,70 @@ function sendTrialEndingEmail( $user = NULL, $membership_id = NULL )
...
+			if( !$user ) {
+				return false;
+			}
+			//If we have a membership ID, get the level.
+			if (! empty( $membership_id ) ) {
+				$membership_level = pmpro_getLevel( $membership_id );
+			//If not let's find out what's the last ended membership level.
+			} else {
+				//Get all levels for a specific member
+				$all_levels = pmpro_getMembershipLevelsForUser( $user->ID );
```
dparker1005, line 1332: `false` can be passed as a second parameter to `pmpro_getMembershipLevelsForUser()` to get inactive levels in addition to the user's active records. If we are looking for the most recently ended membership, having inactive levels in the `$all_levels` array seems necessary.

classes/class.pmproemail.php
```diff
@@ -1306,26 +1306,70 @@ function sendTrialEndingEmail( $user = NULL, $membership_id = NULL )
...
+					return false;
+				}
+				//get the older membership
+				$membership_level = array_reduce( $all_levels, function ( $oldest_level, $current_level ) {
+					// Check if the level status is not "active" and compare end dates
+					if ( $current_level->enddate !== null && ( $oldest_level === null || $current_level->enddate > $oldest_level->enddate ) ) {
+						return $current_level;
+					}
+					return $oldest_level;
+				});
```
dparker1005, line 1344: The level status is not currently being checked here. `$current_level->enddate` may be filled even if the membership is still active.

`$current_level->enddate` is either the date that a membership was ended if the membership is not active, or is the expiration date for the membership if the membership is still active.

dparker1005 (changes requested):

classes/class.pmproemail.php
MaximilianoRicoTabo: why would be the best way to check status then ? compare enddate  with now() function  besides the null scenario ?

dparker1005: Pushed a commit with a direct SQL query instead of using pmpro_getMembershipLevelsForUser(). Should be more efficient and cleaner to read

PR #2968, opened 2024-04-18
Switching the new content admin menu item to respect the same capability as Members List

kimcoleman:
### All Submissions:

Adjusted the capability for the admin + New > Member menu item to respect the same capability we use to check for this on the Members List: edit_users.

### DIFF
--- a/includes/adminpages.php
+++ b/includes/adminpages.php
@@ -273,7 +273,7 @@ function pmpro_admin_bar_menu() {
 	}
 
 	// Add menu item for adding a new member.
-	if ( current_user_can( 'manage_options' ) ) {
+	if ( current_user_can( pmpro_get_edit_member_capability() ) ) {
 		$wp_admin_bar->add_menu(
 			array(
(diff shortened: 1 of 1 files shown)


dparker1005 (commented):
Should we be using the `pmpro_get_edit_member_capability()` function here? It looks like this is used elsewhere in the code for the Edit User page.

kimcoleman:
@dparker1005 I don't think so - we use edit_users here, since [adding a member](https://github.com/strangerstudios/paid-memberships-pro/blob/dev/adminpages/member-edit/pmpro-class-member-edit-panel-user-info.php#L57-L58) also requires this:

https://github.com/strangerstudios/paid-memberships-pro/blob/dev/adminpages/memberslist.php#L24-L26

But looking at this, perhaps we should change this cap (can be a separate PR just noticed). https://github.com/strangerstudios/paid-memberships-pro/blob/dev/adminpages/memberslist.php#L24-L26. That one could be pmpro_get_edit_member_capability().

Issue #2958, opened 2024-04-16
Enhancement: Show payment gateway on the Subscriptions and Orders tabs of the Member Edit page.


ipokkel:
**Is your feature request related to a problem? Please describe.**
When inspecting a member's subscriptions or orders it requires a few extra clicks to determine which payment gateway was used for the subscription or the order.

**Describe the solution you'd like**
Including the payment gateway and environment, e.g. Stripe (test) or Stripe, as additional data in an existing column or a separate column for the gateway used.

**Describe alternatives you've considered**
The class for the orders tab has hooks available to achieve this.
https://gist.github.com/ipokkel/e2384e9b2815dbb6bcd194b3e9dbb8fb
The class for subs does not.

**Additional context**
Related to PMPro 3.0 and newer Member Edit pages.
Maybe for the orders list, we can squeeze the gateway in parentheses after the subscription transaction ID.
And for the subscriptions list, maybe we can add a new column “Gateway” which includes both the gateway and ID (ex `Stripe (sub_flksajdflksjf)` )

PR #2970, opened 2024-04-22
Show "Gateway" on Member Edit

andrewlimaza:
* ENHANCEMENT: Show the payment gateway on the member edit orders panel to make it easier to differentiate where the payments are coming from when using more than one gateway.

Resolves: https://github.com/strangerstudios/paid-memberships-pro/issues/2958

### All Submissions:

1. Navigate to the Member Edit Screen and choose "Orders"
2. See the gateway name included in the table.

Screenshot 2024-04-22 at 11 27 09

### DIFF
--- a/adminpages/member-edit/pmpro-class-member-edit-panel-orders.php
+++ b/adminpages/member-edit/pmpro-class-member-edit-panel-orders.php
@@ -15,7 +15,7 @@ class PMPro_Member_Edit_Panel_Orders extends PMPro_Member_Edit_Panel {
 	 * Display the panel contents.
 	 */
 	protected function display_panel_contents() {
-		global $wpdb;
+		global $wpdb, $pmpro_gateways;
 
 		//Show all invoices for user
 		$invoices = $wpdb->get_results( $wpdb->prepare( "SELECT mo.*, du.code_id as code_id FROM $wpdb->pmpro_membership_orders mo LEFT JOIN $wpdb->pmpro_discount_codes_uses du ON mo.id = du.order_id WHERE mo.user_id = %d ORDER BY mo.timestamp DESC", self::get_user()->ID ) );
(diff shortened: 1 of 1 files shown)


dparker1005 (approved):
Agreed that showing the gateway here would be nice. We probably want this change in the Subscriptions tab as well, along with including the Subscription Transaction ID there.

Long-term, it would be nice to have a system for showing/hiding individual columns similar to the "Screen Options" dropdown on List Tables

adminpages/member-edit/pmpro-class-member-edit-panel-orders.php
```diff
@@ -104,6 +105,25 @@ protected function display_panel_contents() {
...
+								<?php
+								// Logic to get the gateway name and output it neatly.
+									if ( ! empty( $invoice->gateway ) ) {
+										if ( ! empty( $pmpro_gateways[$invoice->gateway] ) ) {
+											$gateway = esc_html( $pmpro_gateways[$invoice->gateway] );
+										} else {
+											$gateway = esc_html( ucwords( $invoice->gateway ) );
+										}
+										if ( $invoice->gateway_environment == 'sandbox' ) {
+											$gateway .= ' (test)';
```
dparker1005, line 118:
```suggestion
											$gateway .= ' (' . esc_html__( 'test', 'paid-memberships-pro' . ')';
```

adminpages/member-edit/pmpro-class-member-edit-panel-orders.php
```diff
@@ -104,6 +105,25 @@ protected function display_panel_contents() {
...
+										}
+
+										echo $gateway;
+									} else {
+										esc_html_e( '&#8212;', 'paid-memberships-pro' );
```
dparker1005, line 123:
```suggestion
										echo  '&#8212;';
```
I don't think we need to localize this, a dash should be shown for all languages.

kimcoleman (approved):
I approve of this - we should be cautious since there is less horizontal space on this panel and also since admins cannot change visible columns. We can tackle some way to have columns hidden or shown if we feel this gets to a place where there are too many columns. For now, we have the space and nothing is wrapping too much even on my Macbook.

PR #2971, opened 2024-04-23
Fixed always redirect to login page

andrewlimaza:
* BUG FIX: Fixed an issue where the redirect would always redirect to the login page.

The issue here was we were ommitting the redirect value entirely from the $login_form_array which would then default to setting a $redirect_to value of the REQUEST URI and it would always redirect to that URL based on the wp_login_form defaults when that parameter was missing.

This fix, fixes issues for custom redirects (snippets) as well as the Member Homepages Add On.

Setting the `redirect` argument to false by default as Paid Memberships Pro handles it's own redirects during the login process and this returns the expected functionality.

### All Submissions:

### DIFF
--- a/includes/login.php
+++ b/includes/login.php
@@ -501,11 +501,8 @@ function pmpro_login_forms_handler( $show_menu = true, $show_logout_link = true,
 					<?php
 						$login_form_array = array(
 							'value_username' => esc_html( $username ),
-							'redirect' => false
+							'redirect' => empty( $redirect_to ) ? false : esc_url( $redirect_to ),
 						);
-						if ( ! empty( $redirect_to ) ) {
-							$login_form_array['redirect'] = esc_url( $redirect_to );
-						}
 						pmpro_login_form( $login_form_array );
 						pmpro_login_forms_handler_nav( 'login' );
 					?>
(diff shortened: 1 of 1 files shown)


dparker1005 (approved):

dparker1005 (changes requested):
According to the WP developer docs, the `redirect` argument should always be a string:
https://developer.wordpress.org/reference/functions/wp_login_form/

In the `wp_login_form()` function itself, it is clear that `redirect` is expected to be a string as well, so passing `false` may throw warnings in recent versions of PHP:
https://github.com/WordPress/wordpress-develop/blob/a2bb04fc352f0b58a822ff83f3bd8ca6fed11757/src/wp-includes/general-template.php#L627

Is there a better "default" redirect argument that we can pass here?

andrewlimaza:
> According to the WP developer docs, the `redirect` argument should always be a string: https://developer.wordpress.org/reference/functions/wp_login_form/
> 
> In the `wp_login_form()` function itself, it is clear that `redirect` is expected to be a string as well, so passing `false` may throw warnings in recent versions of PHP: https://github.com/WordPress/wordpress-develop/blob/a2bb04fc352f0b58a822ff83f3bd8ca6fed11757/src/wp-includes/general-template.php#L627
> 
> Is there a better "default" redirect argument that we can pass here?

I moved over to passing an empty string instead of false. This will still work with the `esc_url` call that eventually happens by the `wp_login_form`.

andrewlimaza:
Added blocking next release tag back to this as it's interfering with Member Homepages and a bunch of custom code that redirects members to specific pages on login.

dparker1005 (approved):

PR #2972, opened 2024-04-24
Fix renewing recurring subscription with PPE

dparker1005:
### All Submissions:

Letting the checkout preheader save the MemberOrder during a successful recurring PPE checkout instead of saving sooner.

When an order is saved in "success" status, a PMPro Subscription is created for the user if needed. By saving the order early, the PMPro sub was created early. Then, when the user's level is changed and all of their active subs are deleted, this new subscription is deleted too.

With the fix, the order isn't saved until after the level change occurs. So now, any previous subscriptions are deleted on level change, and the  new subscription is set up in PMPro afterwards.

Resolves XXX.

1. Log in as a user with a membership level
2. Purchase that same membership level again with a recurring membership via PPE
3. See that before the fix, the new subscription was immediately cancelled. After the fix, the subscription remains as expected

### Other information:

### Changelog entry


andrewlimaza (approved):
Fix resolves the issue, thanks for catching this.

PR #2989, opened 2024-05-14
Removing the accepted credit cards option and select field on billing forms. Deprecating the filter: `pmpro_include_cardtype_field`

kimcoleman:
### All Submissions:

This PR removes the option and display of "We accept {card types}" from billing format.

The option was not actually validating whether an "accepted" card was used. We do not need to show this on the frontend billing forms.

This PR also removes the filter to optionally include a dropdown for "card type" on billing fields. We will now always have card type get set as a hidden field using JavaScript and the credit card validator. We might want to adjust this so it isn't even a hidden field at checkout and is rather just handled when orders are saved. I'm open to talking through that.

### Other information:

### Changelog entry
* DEPRECATED: No longer storing an option for accepted card types or allowing a card type select field in payment forms.
* DEPRECATED: Deprecated the filter `pmpro_include_cardtype_field`.


kimcoleman:
TO DO: I think we also need to remove the default setting in upgrade_1.php to set card types. https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/updates/upgrade_1.php#L22-L23

kimcoleman:
@dparker1005 fixed merge conflicts here.

PR #3013, opened 2024-06-19
Fix a small content warning when trying to add the CSS body class dynamically.

andrewlimaza:
* BUG FIX: Fixed an issue in some cases where a warning would be shown when trying to dynamically add the membership level body CSS class.

NOTE: I cannot replicate this but I assume there is some sort of missing data/stray data in the database that's adding to the array. A customer was facing this issue on a specific page on their site.

### All Submissions:

### DIFF
--- a/includes/content.php
+++ b/includes/content.php
@@ -568,7 +568,9 @@ function pmpro_body_classes( $classes ) {
 		if( ! empty( $post_levels[1] ) ) {
 			$classes[] = 'pmpro-body-level-required';
 			foreach( $post_levels[1] as $post_level ) {
-				$classes[] = 'pmpro-body-level-' . $post_level[0];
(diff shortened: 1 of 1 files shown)


dparker1005 (commented):
In my tests, `$post_levels[1]` looks to be an array of IDs and `$post_level` is just a level ID as a string. Therefore referencing `$post_level[0]` doesn't really make sense.

Maybe the solution here is to just rename the variable `$post_level` to be `$post_level_id` for clarity, and then to switch `$classes[] = 'pmpro-body-level-' . $post_level[0];` to `$classes[] = 'pmpro-body-level-' . $post_level_id;`.

dparker1005:
@andrewlimaza I just want to confirm, do you have any concerns about this new approach?

andrewlimaza:
This does fix the warning, thanks!

PR #3018, opened 2024-06-21
Added filter for subscription cost text

louiswol94:


classes/class-pmpro-subscription.php
```diff
@@ -1041,6 +1045,15 @@ public function set( $property, $value = null ) {
 		$this->{$property} = $value;
 	}
 
+	/**
+	 * Get the membership Level ID for this subscription.
+	 *
+	 * @since 3.1
+	 */
+	public function get_membership_level_id() {
+		return $this->membership_level_id;
```
louiswol94, line 1054: We need the Level ID in the callback function. It was either adding this getter, or changing the property's access to `public` instead of `protected`.

louiswol94:
@dparker1005 @andrewlimaza Please review :)

dparker1005:
Sorry, I had missed this submission before putting together and merging #3021. They are essentially the same solution, but let me know if you have any concerns with the implementation of #3021.

RE the `get_membership_level_id()` function, you should be able to call `$subscription->get_membership_level_id()` without it being explicitly defined in the PMPro_Subscription class as a result of the magic `__call()` method. If that is not working, can you please link me to the code that you are using and I can take a closer look?

dparker1005:
(Closing as #3021 has already been merged, but we can continue discussing `$subscription->get_membership_level_id()` here)

Issue #2463, opened 2023-05-04
Enhancement: CSS to only show the invoice for print media on the PMPro Confirmation page.


ipokkel:
**Is your feature request related to a problem? Please describe.**
When clicking on the "Print" button by the invoice on the confirmation page the complete confirmation page is printed instead of just the invoice.

**Describe the solution you'd like**
To be able to print only the invoice when clicking on the "Print" button on the confirmation page.

**Describe alternatives you've considered**
Add custom CSS to achieve this: https://gist.github.com/ipokkel/f39a2225043f237d5c3c01c196422801

PR #3023, opened 2024-06-26
Single order and order admin print updates

kimcoleman:
### All Submissions:

This is a total overhaul of our frontend order (formerly "invoice" page). Many of the decisions we've made collectively, but I want to point out a few things I made the call on myself:

1. For free orders / $0 orders, we are keeping relatively the same format as a paid order. The payment type will use our common "—" to show that it is not a payment.
2. I'm adding the refunded order "total refunded" as a line item. We can consider to add this to the price parts function instead so we don't duplicate this code + make it filterable. Right now, it duplicates the order total. There are certainly cases where this is not correct, but an order in refunded in the admin side assumes that total value should not count toward reporting, so I think also the best guess for frontend.
3. If the user's billing address is provided, we show that in the "Bill to" area. If not, the "Bill to" area instead shows the user's display name and email address.
4. If the user's billing address is shown in the "Bill to" area, we instead add the user's account display name and email within the order description in the table.
5. We created a new function to show the order's subscription period end date. This is used to show the "period" that order covers for their membership. Right now, if the end date is = to the order date, we do not show anything. This was a tough call, but we maybe need a way rather to store what "would have been" the next payment date for this specifc case where a) the sub is cancelled, b) the membership is active through next pyment date. We did not want to look this up on the membership (too complicated).
6. We moved the check payment instructions to instead hook in via that gateway class. There is still some gateway = 'check' checks - specifically where we show the "payment method" in the invoice.php frontend page template. I am not certain of the best way to get this moved into the class.
7. We kept the pmpro_invoice_bullets_top and pmpro_invoice_bullets_bottom hooks, but they are now just in their own section. It's not how we would do it if we started over. We could consider a new single hook, mark old ones as deprecated and map to the new one. We should test this with plugins that hook in here like group accounts or invite only.
8. We added a new filter `pmpro_order_single_meta` to adjust the meta details of the order shown first on the page. This is stuff like: order date, payment method, pay to (sitename but eventually this will become a full store address), and bill to. I don't want to combine this filter with the invoice bullets old hooks. The formatting of this section is super specific.

We should probably also test this with VAT and anything else that has an order display impact.

I'm open to being challenged on all these decisions and re-evalute/rework as needed. 

### Other information:


kimcoleman:
This is updated now to also support printing the invoice / save as PDF from the Membership Confirmation. This is an interim solution until we consider supporting a standalone frontend order PDF download system. 

Resolves #2463

PR #3045, opened 2024-07-03
Updating invoice wording to order in emails

dparker1005:
### All Submissions:

Updating our wording from "invoice" to "orders" in emails.

Old email template variables with `!!invoice` will still work due to compatibility code in the sendEmail() function of the PMProEmail class.

The tricky case here was with the Stripe-specific emails which reference Stripe invoices. These are truly invoices, so the email contents and variables for that template were not changed.


kimcoleman (approved):
Approved as-is. Let's consider whether we should add an upgrade script to swap instances of variables like like `!!invoice_` to `!!order_``.

PR #3049, opened 2024-07-05
Fix issue of field not hiding

andrewlimaza:
* BUG FIX: Fixed an issue where classes would override the pmpro_hidden styling, so this honors pmpro_hidden classes always.

Using strpos instead of identical conditonal as there may be more than one added custom class. This will always give a priority to `pmpro_hidden` class to hide the field.

### All Submissions:

### DIFF
--- a/classes/class-pmpro-field.php
+++ b/classes/class-pmpro-field.php
@@ -1226,8 +1226,10 @@ class PMPro_Field {
 			$this->divclass .= " ";
 		}
 
-		// Add a class to the field based on the type.
-		$this->divclass .= "pmpro_form_field pmpro_form_field-" . $this->type;
+		// Add a class to the field based on the type and skip if we're trying to hide it with pmpro_hidden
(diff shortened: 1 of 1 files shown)


kimcoleman:
I did a deep review of our Add Ons and Shipping was the only one using a User Fields "setting" / code-based field setup to string in the pmpro_hidden divclass. There is an open PR to approach this in a different way.

Every other time we add pmpro_hidden to a field, we do not include other CSS selectors that are "not respecting" that the field should be "display: none". 

I did raise this issue with Andrew, so this is why we came to have this PR. I'm second guessing that this is needed now. We should not expect anyone else outside our own code to add a divclass like "pmpro_hidden" - if they want to have a field hidden, the "hidden" field type is the correct one to use.

I tested this as is (without the PR) with conditional fields and this doesn't impact them at all. I think this is just a single isolated case where the Shipping Add On needs a different approach and we otherwise will not use the pmpro_hidden class in this way.

PR #3069, opened 2024-07-17
Workaround to use get_date_from_gmt to format into 'U'

mircobabini:
### All Submissions:

Fix the next payment date TIME (and the cancel date TIME if using CONPD in core) to consider the timezone.
When usign get_date_from_gmt with format = 'U', the result does not consider the timezone.
Then, a solution is to call that function twice. One to add the timezone, then to format the datetime.

1. Create a new subscription
2. Check the TIME of the next payment/cancellation

It will be a cycle ahead (MINUS or PLUS the timezone).

### Other information:

### DIFF
--- a/classes/class-pmpro-subscription.php
+++ b/classes/class-pmpro-subscription.php
@@ -783,6 +783,12 @@ class PMPro_Subscription {
 
 		// Get date in WP local timezone.
 		if ( $local_time ) {
+			if( 'U' === $format ){
+				// When formatting using the epoch, the date must already consider the timezone offset.
+				// Then, first apply a simple format to add the timezone.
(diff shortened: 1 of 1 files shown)


dparker1005 (approved):
Overall, looks good to me.

I hesitated a bit when initially reviewing this PR since 'U' is, by definition, the seconds since the Unix Epoch (January 1 1970 00:00:00 GMT). Altering the date to a "local timestamp" like this PR breaks that definition.

With that being said, creating a "local timestamp" like this is something that we already do elsewhere in PMPro, so this change would bring the PMPro Subscriptions class more in line with other PMPro code.

I think we should merge this change and, down the line, consider setting stronger standards about how we handle timestamps in PMPro.

PR #3099, opened 2024-07-31
Exploratory PR to allow people to adjust the style of the mounted Stripe Elements fields.

kimcoleman:
### All Submissions:

People with dark themes have no way to target and style the Stripe elements fields that get added via JS. This PR adds a new filter `pmpro_stripe_card_element_style` so that custom styles can be added.

Here is an example of using this hook to make the placeholder and field values white. You may still need custom CSS to override core PMPro style variations and how background colors and borders are set.

https://gist.github.com/kimcoleman/ab5282830d18252ec39b9dab124d6e7a


kimcoleman:
This is the list of variables I think you can use. While this doc shows them mounted to the entire element render itself, they do also work when added to the single inputs.

dparker1005 (approved):

PR #3104, opened 2024-08-01
Now sending users to Stripe Checkout during the process method

dparker1005:
### All Submissions:

Previously, we had to send users to Stripe Checkout on the `pmpro_checkout_before_change_membership_level` hook since user objects would not exist during the gateway's `process()` method. Now, however, users are created before payments are taken.

Currently, if sending a user to Stripe Checkout fails (due to not being able to connect to Stripe, for example), the checkout will complete successfully without collecting payment. Switching to using the `process()` method allows us to halt checkout on these errors and show a relevant error message at checkout.

The easiest way to replicate is to set up a local dev environment and check out with Stripe Checkout with your machine's wifi disabled.

### Other information:

### Changelog entry


dparker1005:
This should wait for a minor release in case custom code is calling our Stripe gateway's `pmpro_checkout_before_change_membership_level()` method. This is extremely unlikely, but technically breaking

andrewlimaza (approved):
This works out well, thanks David!

PR #3118, opened 2024-08-15
pmpro-buddypress field sync support in GUI

DarkLotus:
### All Submissions:

* [X] Have you followed the [Contributing guideline](https://github.com/strangerstudios/paid-memberships-pro/blob/master/.github/CONTRIBUTING.md)?

Brings across the abilty to enable Syncing PMPRO Fields to  BP XProfile fields via the GUI instead of just via register helper code defined fields.
Adds a new input thats only visible in the user field settings if pmpro-buddypress is enabled, that lets you set the $field->buddypress value via GUI.

### Other information:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?

### Changelog entry

> Add support for GUI user fields to support BP sync if pmpro-buddypress is installed.

### DIFF
--- a/includes/fields.php
+++ b/includes/fields.php
@@ -1342,11 +1342,12 @@ function pmpro_get_field_html( $field = null ) {
         $field_type = $field->type;
         $field_required = $field->required;
         $field_readonly = $field->readonly;     	
-        $field_profile = $field->profile;
+		$field_buddypress = $field->buddypress;    	
         $field_wrapper_class = $field->wrapper_class;
         $field_element_class = $field->element_class;
         $field_hint = $field->hint;
         $field_options = $field->options;
+		$field_profile = $field->profile;
     } else {
         // Default field values
         $field_label = '';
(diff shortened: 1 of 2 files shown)


dparker1005:
Hey @DarkLotus, thanks for putting this together.

After looking into this further, we're going to close this one without merging.

Coded user fields have this built in due to legacy code, but moving forward we want to keep integration-specific code inside the corresponding integrations rather than in core. To support this kind of thing properly from the GUI, there are some core changes we'd need to make first, and then matching development work on the pmpro-buddypress add-on to use them. No ETA on either piece for now.

In the meantime, there's an existing recipe that works as a workaround today — admins can map their PMPro field names to XProfile field names via the pmpro_add_user_field filter and BP sync picks it up from there: https://github.com/strangerstudios/pmpro-snippets-library/blob/master/add-ons/pmpro-buddypress/sync-buddypress-fields.php

Really appreciate the contribution.

PR #3141, opened 2024-09-17
User Field Condtions (AKA "Depends")

andrewlimaza:
This adds functionality and settings for conditional User Fields.

The following conditions that have been added:

* is empty
* is not empty
* is equal to
* is not equal to
* contains
* does not contain


dparker1005:
Closing since this work never made it past draft status. If this is still a feature that we want to release, we can use AI to open a new PR off of the up-to-date dev branch.

PR #3144, opened 2024-09-23
Standardizing Page Builders - Elementor

JarrydLong:
### All Submissions:

Standardises how our page builder settings look, using the same controls throughout. 

image

1. Create a page or post and edit in Elementor. Add an element
2. Click on the Advanced tab and then on the Paid Memberships Pro tab
3. Adjust settings as needed

### Other information:


includes/compatibility/elementor/class-pmpro-elementor-content-restriction.php
```diff
@@ -19,27 +19,61 @@ protected function content_restriction() {
...
+				'default' => 'no',
+            )
 		);
 
-		$element->add_control(
+        $element->add_control(
+			'pmpro_invert_restrictions', array(
+				'type' => \Elementor\Controls_Manager::SELECT,
+				'options' => array(                
+					'0' => esc_html__( 'Show content to...', 'textdomain' ),
```
andrewlimaza, line 40: 'textdomain' needs to be updated to 'paid-memberships-pro'

andrewlimaza:
We also need to add backwards compatibility to this logic. The data is available but lost when upgrading to this script because of the new settings added, once changing to "Specific Membership Levels" it shows - we can automate this by checking for the values for backwards compatibility. It does need a bit more thought.

dparker1005:
Moving back to draft as we need to rethink the migration approach.

andrewlimaza:
Simplified this code where possible and be as performant as possible. This needs more testing, mainly when migrating from a fully built site (with multiple pages) and variations of content restrictions on the container and element Elementor settings.

To test this we need:

1. Create various pages using Elementor with Elementor specific restrictions: Container and Element level.
2. The restrictions should be varied such as restrict for non-members, show no access message to multiple levels and not showing the no access message.
3. Once these pages are set up correctly - pull this PR into your environment, visiting the pages as either the frontend or the backend _should_ upgrade/update the settings without even noticing that settings were updated.

**Note**: Once you upgrade, you cannot downgrade. This removes the 'legacy' settings from the post meta/element settings and is irreversible. Downgrading would mean that you need to reconfigure your levels and restrictions.

PR #3163, opened 2024-10-11
Comment out cancel_subscription method in the pmprogateway class

andrewlimaza:
This is a temporary fix to remove the update_subscription method from the base gateway class as we have checks to see if the method_exists for a gateway object. I've commented this out versus entirely removing it from the class so developers can reference this.

This will always return true since we are extending the base class and `method_exists` isn't smart enough to know if we're overriding the method or not within our actual gateway used class.

Gateways may still declare the `cancel_subscription` method and use this as the preffered/newer way of doing things and gateways that are still using the deprecated `cancel()` method would still work.

### DIFF
--- a/classes/gateways/class.pmprogateway.php
+++ b/classes/gateways/class.pmprogateway.php
@@ -102,7 +102,7 @@
 		/**
 		 * Cancel the subscription associated with the passed order.
 		 *
-		 * @deprecated 3.2 Use cancel_subscription insetad.
+		 * @deprecated 3.2 Use cancel_subscription instead by overriding it in each gateway class.
 		 *
 		 * @param MemberOrder $order The order object associated with the subscription to cancel.
 		 * @return bool True if the subscription was canceled successfully, false otherwise.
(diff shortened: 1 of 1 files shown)


andrewlimaza:
To test this with PayFast is a good way to test before and after merging this.

1. Navigate to sandbox.payfast.co.za (You do not need an account, your email address is your unique identifier for sandbox)
2. Install PMPro PayFast integration - https://wordpress.org/plugins/pmpro-payfast/
3. Create a monthly membership level for any amount and checkout with Payfast.
4. Inside the sandbox account, you should see under "Subscriptions" a token. Take note of the timestamp
5. Cancel your membership and see that it says it's successful but it's not really cancelled in the gateway.
6. Go to Edit Member for the account you used to take the above steps and choose "Sync With Gateway", see that the subscription returns that it's still active.

Repeat these steps after pulling this code and see that the subscription is cancelled correctly.

ideadude:
Closing in favor of this one: https://github.com/strangerstudios/paid-memberships-pro/pull/3164

I'm worried other code (pay by check?) expects the cancel_subscription method to be available in the base class. In any case, we put it there for a reason. PR 3164 instead creates a pmpro_method_defined_in_class() function that performs what the original code intended, checks if the cancel_subscription method is specifically defined in the gateway class.

We may want to switch this code up in the future, but I feel the newer PR fixes the issue, while avoiding new issues.

PR #3178, opened 2024-10-25
Updated Stripe cancelSubscriptionAtGateway for subs table

dparker1005:
### All Submissions:

Updated the `cancelSubscriptionAtGateway()` method of the Stripe Gateway clas to reference subscriptions instead of orders.

This PR also fixes an issue when PMPro tries to cancel a subscription in Stripe, but the subscription is already "scheduled for cancellation" in Stripe. In this case, `$subscription->canceled_at` would be set for the future and this method would bail early. This may happen in cases where a customer requests cancellation via the Stripe Customer Portal and Stripe tries to use their own CONPD behavior.

### Other information:

### Changelog entry

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -3369,25 +3369,14 @@ class PMProGateway_stripe extends PMProGateway {
 			return false;
 		}
 
-		// If this is already cancelled, return true.
-		if ( ! empty( $subscription->canceled_at ) ) {
-			return true;
-		}
-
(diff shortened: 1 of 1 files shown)


kimcoleman (approved):
Tested where we were seeing this error on our live site and it was successfully cancelled. Approved!

ideadude:
Looks good, but I'm thinking about how this relates to the update code when updating from PMPro 2 to 3.

If we avoid instantiating the order during this, is it possible that the subscription data may not be in the new table yet and so using the subscription class would fail?

Can you test this or confirm in the code that this case doesn't happen?

PR #3184, opened 2024-10-30
Marking PayFlow, PayPal Standard, and Braintree as deprecated

dparker1005:
### All Submissions:

Sites already using these gateways can continue using them, but they will not be shown as options on new sites.

### Other information:

### Changelog entry


ideadude (approved):
Looks good. Ready to merge.

FYI it seems we were loading the Braintree class early so the dependency checks would run even if not being used, which is okay to skip now that we are deprecating.

PR #3190, opened 2024-11-08
Now setting the discount code ID on orders before saving

dparker1005:
### All Submissions:

Currently, discount codes are set for orders after the orders have been saved. This means that the discount code attributes are not correct during the `pmpro_added_order` hook and the `pmpro_updated_order` hook.

This PR updates the logic when saving orders at checkout and on the Edit Order page to set the discount code during the order save process to make sure that those actions work as expected.

Note: When using offiste gateways (such as Stripe Checkout or PayPal Express), the discount code will still not be available during the `pmpro_added_order` hook. This is because the order is initially added in `token` status and is later updated to `success` with the discount code added once the checkout is complete. In these cases, it is best to use the `pmpro_after_checkout` or `pmpro_order_status_success` hooks.

### Other information:

### Changelog entry

### DIFF
--- a/adminpages/orders.php
+++ b/adminpages/orders.php
@@ -180,6 +180,11 @@ if ( ! empty( $_REQUEST['save'] ) ) {
 		}
 	}
 
+	// Set the discount code.
+	if( isset( $_REQUEST['discount_code_id'] ) ) {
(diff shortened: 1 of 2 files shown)


ideadude:
Looks good to ship.

PR #3191, opened 2024-11-08
Adding user ID to getRandomCode() in case other hash parameters are identical

dparker1005:
### All Submissions:

Adding user ID to getRandomCode() in case other hash parameters are identical. Specifically for cases where to orders are purchased at the exact same time and `rand()` produces the same result for the `$auth_code` and `secure_auth_code`.

### Other information:

### Changelog entry

### DIFF
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -1573,7 +1573,7 @@
 		 * Get a random code to use as the order code.
 		 */
 		function getRandomCode() {
-			global $wpdb;
+			global $wpdb, $current_user;
 
 			// We mix this with the seed to make sure we get unique codes.
 			static $count = 0;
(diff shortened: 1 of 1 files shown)


ideadude:
Do we always have a user ID here, especially in the cases where we had conflicts (I think 2 people checking out at the exact same time)? Or the rare case the same user is creating 2 orders at the same time (fairly unlikely, but maybe when running test scripts for unit testing, or some other case like that).

Would something like session_id() work as well or better? Let's also test that session_id() gets a unique value. Or could use IP address. Or some combination of user id, session id, ip address, and anything else we can think of. If we add one extra md5, no big deal performance-wise, as long as getting these values is constant time.

dparker1005:
Agreed that we should brainstorm a way to make this 100% secure, but this PR is sufficient to avoid issues in 98%  of legitimate cases. We can always tweak that function however we want, there isn’t any code debt for making incremental improvements there

PR #3200, opened 2024-11-14
WordPress 6.7 compat. Fix doing_it_wrong notice related to early translations

mircobabini:
### All Submissions:

Some setups are showing a doing_it_wrong notice starting from WP6.7 (related to https://make.wordpress.org/core/2024/10/21/i18n-improvements-6-7/)

To investigate on your local env and see all the errors, modify your core `function _load_textdomain_just_in_time` this way:
Screenshot 2024-11-14 at 12 23 14

Causes:
1. Reports registration. Fixed 2e9bfe32341f07547188b0d5716adf2f5e578680
2. (not fixed in this PR) User fields default group name (which is also tricky to be filtered because it runs way before other plugins and themes)
Screenshot 2024-11-14 at 12 26 46

Tests to be done:
- [ ] Backward compatibility for the admin reports from third party plugins/themes.

### Other information:

* [ ] TEST TO BE DONE!!!

### Changelog entry

### DIFF
--- a/adminpages/reports.php
+++ b/adminpages/reports.php
@@ -6,6 +6,15 @@
 global $pmpro_reports;
 
 /**
+ * Filter the admin reports provided by our core, our Add Ons, and other plugins/themes.
+ *
+ * @since TBD
+ *
+ * @param array  $pmpro_reports The registered reports.
+ */
+$pmpro_reports = apply_filters( 'pmpro_registered_reports', $pmpro_reports );
+
+/**
 * Load the Paid Memberships Pro dashboard-area header
 */
 require_once( dirname( __FILE__ ) . '/admin_header.php' ); ?>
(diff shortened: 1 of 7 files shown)


dparker1005:
Added a commit to fix the "default group name" in `includes/fields.php`. The fix that we are going with is no longer translating the default group name. This functionality will be restored in a future version of PMPro. Until then, site admins will need to change the name of the group on new sites when adding user fields (which they are likely already doing anyways).

dparker1005:
Note that this new approach for registering reports is not compatible with the Mobile Report Dashboard Add On. I'm going to look into a fix for this

ideadude:
This feels like the correct solution, even though it may break some other code (like the mobile reports dashboard) that expects the $pmpro_reports global to be set up before init. There could be race conditions on some code we have e.g. that tries to rearrange report positions on init and stuff like that.

I still think we should ship it and fix the issues that come up as they come up.

PR #3211, opened 2024-11-22
Custom translation updates (GlotPress)

andrewlimaza:
* ENHANCEMENT: Added functionality to look for updates via our own translations server.

### All Submissions:

### DIFF
--- a/includes/lib/glotpress-helper.php
+++ b/includes/lib/glotpress-helper.php
@@ -1,0 +1,270 @@
+<?php
+/**
+ * Main Traduttore Registry library.
+ *
+ * @since 1.0.0
+ *
+ * Copyright (c) 2018-2020 required (email: info@required.com)
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License, version 2 or, at
+ * your discretion, any later version, as published by the Free
+ * Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+namespace Required\Traduttore_Registry;
+
+use DateTime;
+
+const TRANSIENT_KEY_PLUGIN = 'traduttore-registry-plugins';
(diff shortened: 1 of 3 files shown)


andrewlimaza:
Update Traduttore still works with older PHP versions (without fatal erroring) and our recommended PHP version is 7.4+, so this is okay.

includes/localization.php
```diff
@@ -56,3 +56,41 @@ function pmpro_translate_billing_period($period, $number = 1)
...
+ * Handle translation updates from our own translation server.
+ * @since TBD
+ */
+function pmpro_check_for_translations() {
+	// Run it only on a PMPro page in the admin.
+	if ( ! current_user_can( 'update_plugins' ) ) {
+		return;
+	}
+
+	$is_pmpro_admin = ! empty( $_REQUEST['page'] ) && strpos( $_REQUEST['page'], 'pmpro' ) !== false;
```
ideadude, line 70: We do a similar check here: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/includes/admin.php#L327-L331

Maybe we should have a function in includes/admin.php or adminpages/functions.php for this?

includes/lib/glotpress-helper.php
```diff
@@ -0,0 +1,270 @@
...
+
+	if ( ! \is_object( $translations ) ) {
+		return;
+	}
+
+	/*
+	 * Don't delete the cache if the transient gets changed multiple times
+	 * during a single request. Set cache lifetime to maximum 15 seconds.
+	 */
+	$cache_lifespan   = 15;
```
ideadude, line 154: Can we document how the caching works? Maybe WP core is doing it's own caching too.

How often will the site hit our servers? Ideally we don't want the admin pages to be hitting our servers every time they load a plugins page or our settings pages.

includes/localization.php
andrewlimaza: I like the idea of having an admin function that checks is a pmpro setting page. We're starting to do this more.

includes/lib/glotpress-helper.php
andrewlimaza: We can get way more aggressive with the cache. I'd be okay to bump it up by a day or two (we have a way to manually download the translation as well or they could delete the transient to ping the server when needed).

dparker1005: I spent some time looking into this. TLDR: our `clean_translations_cache()` here that clears caches older than 15 seconds should only be run twice daily (by default) on the core `set_site_transient_update_plugins` hook or whenever plugins are manually rechecked or a plugin is updated triggering the `delete_site_transient_update_plugins` hook.

So in a sense,  we do have some built-in caching from WP. It looks like the 15 second delay is really just there to prevent updating translations for a plugin multiple times on a single page load

PR #3216, opened 2024-11-28
[BUGFIXING] * Rename spanish translation files to get the completed one.

MaximilianoRicoTabo:
* Add few remaining translations

### All Submissions:

Resolves an issue reported by a freescout user

### Other information:

### Changelog entry


andrewlimaza (approved):

dparker1005:
This PR deletes the `languages/paid-memberships-pro-es.mo/.po` files and replaces those with `languages/paid-memberships-pro-es_ES.mo/.po`. I just want to confirm that we definitely want to delete the former instead of including both. @andrewlimaza Can you confirm?

andrewlimaza:
Digging into this a bit more, we should not delete the `languages/paid-memberships-pro-es.mo/.po` as this would be a fallback for Spanish locales/translation files don't exist.

For example if we do not have Spanish (Argentina) file present it would use this Spanish file generically. We should ensure that the `languages/paid-memberships-pro-es.mo/.po` is up to date and correct.

This may have been my bad in saying we don't actually need it, but it won't hurt since it's a fallback file for this locale. Same goes for any other locale that may have colloquial terms per country (i.e. English - UK, English - South Africa, French - Canada, French - Belgium etc).

These files can omit the location like `languages/paid-memberships-pro-fr.mo/.po` and `languages/paid-memberships-pro-es.mo/.po`. We should make sure these are up-to-date and fix it as they come through.

andrewlimaza:
We shouldn't delete this file but make sure both `es_ES` and the `es` match for now and our translation server will resolve future cases once that goes out.

andrewlimaza:
@dparker1005 I've made the change by putting the `es.mo/.po` files back from the `es_ES.mo/.po` versions. It's worth to have a fallback.

PR #3217, opened 2024-11-28
[ENHANCEMENT]  Add Copy to clipboard suport to discount codes, order ids, payment ids and more.

MaximilianoRicoTabo:
…nstead of open edit page

image
image
image
image
image
image

### All Submissions:

This PR adds small button/icon next to the codes (discount codes, subscripton ids, order id, payment id) and event handlers in the JS admin file to allow users copy to the clipboard those codes.

To test the scenario where the user's browser doesn't have access to clipboard API just start an instance without HTTPS.

To test the happy path scenario just go through different places with buttons and either click to copy and paste it somewhere else. Also it's possible to navigate with tab and copy hitting enter key for accessibility sake.

### Other information:

### Changelog entry


kimcoleman:
This link isn’t broken as is. It doesn’t open the edit page. It duplicates the discount code as a new code with same settings. 

Let’s discuss whether the word copy is misleading and duplicate is better.

MaximilianoRicoTabo:
Should I close this one then @kimcoleman  ?

kimcoleman:
We can chat about this in a dev meeting. ‘click to copy’ would be nice some other places in the plugin.

dparker1005:
When testing, the `! navigator.clipboard` check was failing on my local testing site on both Chrome and Safari. It looks like this may be due to the site not being HTTPS. Maybe we should put up an alert if the copy fails for some reason instead of just returning. Or even better, maybe we should only render the "copy" button if the browser supports copying.

If we're adding this behavior for copying the "discount code" code, maybe we should also add this for copying order codes as well. Or maybe we want a more universal solution that could even be used for payment/subscription IDs.

MaximilianoRicoTabo:
Hey @dparker1005 I took navigator.clipboard support from here -> https://caniuse.com/mdn-api_navigator_clipboard  It seems pretty supported to me, but fair enough to hide the button if the browser doesn't support.

Also, we can try a more universal solution, happy to add more effort to this PR to bring functionality to orders and payment / subs ids

andrewlimaza:
I feel that the clipboard icon should follow a more standardized icon.
Screenshot 2025-01-21 at 12 00 58

<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><path d="M384 336l-192 0c-8.8 0-16-7.2-16-16l0-256c0-8.8 7.2-16 16-16l140.1 0L400 115.9 400 320c0 8.8-7.2 16-16 16zM192 384l192 0c35.3 0 64-28.7 64-64l0-204.1c0-12.7-5.1-24.9-14.1-33.9L366.1 14.1c-9-9-21.2-14.1-33.9-14.1L192 0c-35.3 0-64 28.7-64 64l0 256c0 35.3 28.7 64 64 64zM64 128c-35.3 0-64 28.7-64 64L0 448c0 35.3 28.7 64 64 64l192 0c35.3 0 64-28.7 64-64l0-32-48 0 0 32c0 8.8-7.2 16-16 16L64 464c-8.8 0-16-7.2-16-16l0-256c0-8.8 7.2-16 16-16l32 0 0-48-32 0z"/></svg>

kimcoleman:
We're going to hold off on this for 3.4. There's discussion about whether showing these copy links on the list tables vs. the single views only is a better route. I want to explore other platforms and click to copy functionality to decide whether we should pepper these around so much on the list tables and/or the accessibility and expected language/icons.

dparker1005:
We are going to close this PR for now. If this functionality is requested again, we will likely take a different approach in the future.

PR #3230, opened 2024-12-13
Test wp_options values for notifications

andrewlimaza:
* ENHANCEMENT: Added notification check for wp_option (general) option which will help for compatibility plugins AND any of our own pmpro related options that are arrays.

This has been tested with a custom JSON file, as well as various option values both single values and arrays. The structure of the JSON required follows the pmpro_setting logic and extended into non-pmpro options.

### All Submissions:

### DIFF
--- a/includes/notifications.php
+++ b/includes/notifications.php
@@ -436,6 +436,40 @@ function pmpro_notification_test_pmpro_setting( $data ) {
 }
 
 /**
+ * Test any wp_option setting for a specific value.
+ * @param array $data Array from the notification with [0] option name to check [1] value to check for.
+ * @returns bool true if an option if found with the specified name and value.
+ */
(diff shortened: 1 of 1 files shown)


ideadude:
Our original pmpro option check could only check for exact matches. It would be nice to be able to do a couple other checks:
* empty or not
* numerically greater than, less than, equal
* "contains" text searches

Another nice to have would be support for serialized data. So if we stored an option called settings1 like { option1: value1, option2: value2 } we would want to be able to check settings1.option2 value.

Another nice to have is checking user meta.

This all doesn't have to be supported now, but it would be nice to have it in this version of the API ASAP since we're introducing it.

We could switch to using the new and improved wp_options check for things instead of the limited pmpro options check.

andrewlimaza:
I'm reworking newer logic and will reference this PR instead of updating it.

PR #3234, opened 2024-12-17
Abstract email templates into individual classes

dparker1005:
### All Submissions:

Creating an abstract class for email templates will allow us to avoid duplicate code that is common across every email template. Separating each template into a separate class will then make it easier to modify existing emails and add additional email templates down the road.

### Other information:

### Changelog entry


classes/email-templates/class-pmpro-email-template-billing-admin.php
```diff
@@ -0,0 +1,212 @@
...
+	public static function get_default_subject() {
+		return __( 'Billing information has been updated for !!user_login!! at !!sitename!!', 'paid-memberships-pro' );
+	}
+
+	public static function get_default_body() {
+		return __( '<p>The billing information for !!display_name!! at !!sitename!! has been changed.</p>
+<p>Account: !!display_name!! (!!user_email!!)</p>
+<p>
+	Billing Information:<br />
+	!!billing_name!!<br />
```
kimcoleman, line 81: This is the only template that seems to build out the address using individual variables. Is now an ok time to swap this to:

Billing Information:<br />
!!billing_address!!

dparker1005: Done

kimcoleman:
In looking at this deeper now it seems dev does have "duplicate" variables like login_link, orders_link, and levels_link. They seem to be just populated in the PHP, not shown as "variables" you can use. Probably in an effort to reduce confusion. Maybe we had to keep these because of old merged in Add On, like CONPD? 

IF we keep both the _link and _url versions just to make sure the populate, let's only show the _url versions, which seem to be what we showed in our variable reference from.

dparker1005:
> In looking at this deeper now it seems dev does have "duplicate" variables like login_link, orders_link, and levels_link. They seem to be just populated in the PHP, not shown as "variables" you can use. Probably in an effort to reduce confusion. Maybe we had to keep these because of old merged in Add On, like CONPD?
> 
> IF we keep both the _link and _url versions just to make sure the populate, let's only show the _url versions, which seem to be what we showed in our variable reference from.

Addressed in my last commit

adminpages/emailtemplates-edit.php
```diff
@@ -18,7 +18,7 @@
 
 	// Email variables.
 	$email_variables = [
-		'general' => [
+		__( 'General Settings / Membership Info', 'paid-memberships-pro' ) => [
```
ideadude, line 21: All of the __() here should be esc_html__()... unless they have HTML in them, then they need to be refactored to escape the translation and keep the required HTML.

ideadude: Of course, this was a .org requirement and we will have more control over our translations and thus more trust of them. If it's a hassle, this can probably wait, but it's a good practice.

classes/email-templates/class-pmpro-email-template-admin-change-admin.php
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Register the email template.
+ *
+ * @since TBD
+ *
+ * @param array $email_templates The email templates (template slug => email template class name)
+ * @return array The modified email templates array.
+ */
+function pmpro_email_templates_admin_change_admin( $email_templates ) {
```
ideadude, line 160: Just a note here to document that we could use a closure (anonymous function) here and the code would be a bit tighter and we would avoid creating yet another prefixed function for the namespace...

However, using a closure here like this would make it harder to unhook an email template using remove_filter, which folks may want to do.

So well done. I agree with the use of the named function to add the template here, even if it cluttered the namespace a bit.

adminpages/emailtemplates-edit.php
dparker1005: Is this still the case when the content isn't being echo'd immediately? Currently, we are running this data through esc_html() during output as per "sanitize early, escape late":
`<th><code><?php echo esc_html( $email_variable ); ?></code></th>`

dparker1005: It's probably best to escape anyways. Best to be overly cautious

PR #3247, opened 2025-01-08
Using the discount code set on the level object when set

dparker1005:
### All Submissions:

During asynchronous checkouts, we are currently saving the discount code used at checkout in an order meta field and also on the checkout level object. We have seen issues where those discount codes do not match which may cause discount code uses not to be saved.

To streamline our data storage and avoid duplicate (and potentially conflicting) information, this PR makes the discount code set on the checkout level the source of truth. In a future breaking update, discount code data will only be pulled from the checkout level object when completing checkout.

### Other information:

### Changelog entry

### DIFF
--- a/includes/checkout.php
+++ b/includes/checkout.php
@@ -69,6 +69,7 @@ function pmpro_calculate_profile_start_date( $order, $date_format, $filter = tru
 	update_pmpro_membership_order_meta( $order->id, 'checkout_level', $pmpro_level_arr );
 
 	// Save the discount code.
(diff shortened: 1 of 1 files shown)


ideadude (approved):
Good. Ready to merge.

I thought maybe now is the right time to introduce a PMProDiscountCode class, which could among other things abstract and cache the queries to get the code by id/etc, but we can put that off until later. I don't see any big performance concerns here. Would be nice to include in 4.0 though for cleaner code.

PR #3278, opened 2025-02-03
Improved notification checks.

andrewlimaza:
* ENHANCEMENT: Added a new notification conditional callback for `_options` table with more conditionals ( =, !=, <, <=, >, >=, contains, notcontains, empty and notempty).

### All Submissions:

### DIFF
--- a/includes/notifications.php
+++ b/includes/notifications.php
@@ -453,6 +453,67 @@ function pmpro_notification_test_site_url_match( $string ) {
 }
 
 /**
+ * Check against the WordPress options table.
+ * @param array $data Array from notification with plugin_file, comparison, and version to check. $data[0] is the option name, $data[1] is the comparison operator, and $data[2] is the value to compare against.
+ * @returns bool true if plugin is active and version comparison is true, false otherwise. *
+ */
+function pmpro_notification_test_check_option( $data, $option = NULL ) {
+
+    if ( ! is_array( $data ) ) {
+        return false;
+    }
+    
+    if ( ! isset( $data[0] ) || ! isset( $data[1] ) || ! isset( $data[2] ) ) {
+        return false;
+    }
+
+	// No option passed in, assume it's the first passthrough.
+    if ( empty( $option ) ) {
+        $option = get_option( $data[0] );
+    }
+
+    // If $option is an array, recursively check all its values.
+    if ( is_array( $option ) ) {
+        foreach ( $option as $opt ) {
+            if ( pmpro_notification_test_check_option( $data, $opt ) ) {
+                return true;
+            }
+        }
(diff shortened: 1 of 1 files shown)


ideadude:
This is great and addresses our desire to ask questions like "does the value of option1 contain some string".

Often we are saving settings into one option as a serialized array, or less commonly as a serialized object. We may want to check the value of a sub element of that option, so e.g. in the pmpro-buddypress plugin we store settings like this: https://github.com/strangerstudios/pmpro-buddypress/blob/dev/includes/pmpro-buddypress-settings.php#L42-L55

````
$pmpro_bp_options = array(
	'pmpro_bp_restrictions'				=> $pmpro_bp_restrictions,
	'pmpro_bp_group_creation'			=> $can_create_groups,
	'pmpro_bp_group_single_viewing'		=> $can_view_single_group,
	'pmpro_bp_groups_page_viewing'		=> $can_view_groups_page,
	'pmpro_bp_groups_join'				=> $can_join_groups,			
	'pmpro_bp_private_messaging'		=> $pmpro_bp_private_messaging,
	'pmpro_bp_public_messaging'			=> $pmpro_bp_public_messaging,
	'pmpro_bp_send_friend_request'		=> $pmpro_bp_send_friend_request,
	'pmpro_bp_member_directory'			=> $pmpro_bp_member_directory,
	'pmpro_bp_group_automatic_add'		=> array(),
	'pmpro_bp_group_can_request_invite'	=> array(),
	'pmpro_bp_member_types'				=> array());
update_option('pmpro_bp_options_users', $pmpro_bp_options, 'no');
````

So if we wanted to see if the pmpro_bp_groups_join value of the pmpro_bp_options was 1, we can maybe allow the option name parameter of the rule to be "pmpro_bp_options>pmpro_bp_groups_join" or "pmpro_bp_options:pmpro_bp_groups_join". Since > and : are not allowed in option names, we can look for that character and if found pull out the "sub option" and check that.

I like : better than > or -> for this.

andrewlimaza:
For this PR, we have to thoroughly test as much cases that we can:

* Conditions: `=, ==, !=, >, <, >=, <=, contains, notcontains, empty, notempty`
* We need to test "sub-options" such as: `pmpro_bp_options:pmpro_bp_groups_join`
* Values should be single strings, integers as well as arrays or object values (i.e. $pmpro_level->ID) etc.

While this may be a bit difficult to test, it's best to create a locally hosted JSON file with custom notification rules and run them and show notifications. It's best to create one notification with very specific conditions and see if those conditions are met and the banner shows. 

After every test, we should clear the JSON and any server-side caching (transients) to see if the new notification shows. This will be the best way to ensure conditions are working.

PR #3285, opened 2025-02-06
Additional Order Filters

JarrydLong:
### All Submissions:

You can now filter the orders table by selecting:

1. A date
2. Membership Level
3. Discount Code
4. Paid/Free Orders or a custom range (eg $0 - $50)
5. Gateway paid (only for sites with multiple gateways)
6. Order status (success, token, error etc)

### Other information:

### DIFF
--- a/adminpages/orders-csv.php
+++ b/adminpages/orders-csv.php
@@ -30,73 +30,91 @@ $max_orders_per_loop = apply_filters( 'pmpro_set_max_orders_per_export_loop', 20
 global $wpdb;
 
 //get users
-if ( isset( $_REQUEST['s'] ) ) {
+if ( ! empty( $_REQUEST['s'] ) ) {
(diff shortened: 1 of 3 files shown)


dparker1005:
The backend logic here feels like it's in the right direction, but after seeing how having all the filters visible simultaneously looks, we may want to approach UI differently. This is especially true with some filter selections adding additional fields, such as "min" and "max" total fields.

Here is a screenshot demonstrating how complicated having all the fields visible can get:
Screenshot 2025-03-17 at 2 27 06 PM

Perhaps a system where there is an "add filter" button that allows creating a list of filters one at a time? This would keep the UI clean and minimal while making adding new filters down the road more straightforward.

This will likely need more discussion before we begin working on a revised approach as a cycle project.

dparker1005:
Closing as a separate PR added filters to this table in PMPro v3.7

PR #3333, opened 2025-03-07
Add "Select Country" at the beginning.

andrewlimaza:
* ENHANCEMENT: Added "Select Country" at the beginning. This is a great fallback if no default country is set instead of choosing the first valid country.

There are some Add Ons that would benefit from this, any that add a country field to checkout, member profiles etc. While it would work to squeeze the "Select country" in anytime we want to use $pmpro_countries, it's best to solve this at it's core, otherwise we will have to constantly squeeze it in everytime we add this (it may not be often but it seems best practice to do this in core). 

This also makes logic around Javascript onchange type of code easier to manage, and work with and also reduces errors as it would default to "Afghanistan" or any filtered value that was the first value of the array when the default country is not detected at all, as well as required field logic to work and not be bypassed because it would default to a non-zero'd value.

One thing to note here, the State Dropdowns Add On added in logic to "Select Country" and squeeze it in when enqueuing the JavaScript files and values/logic, we will need to do an update once this is merged in and released to avoid duplicate "Select country" options. 

This is better as people can still remove it via the `pmpro_countries` filter with custom development whereas squeezing it in here and there doesn't allow developer's to easily remove it.

### All Submissions:

### DIFF
--- a/includes/countries.php
+++ b/includes/countries.php
@@ -256,6 +256,6 @@
 	asort($pmpro_countries);
 
 	// Add the blank option at the start of the array, always.
-	$pmpro_countries = array( ''   => __( 'Select country', 'paid-memberships-pro' ) ) + $pmpro_countries;
+	$pmpro_countries = array( ''   => __( 'Select Country', 'paid-memberships-pro' ) ) + $pmpro_countries;
 
 	$pmpro_countries = apply_filters("pmpro_countries", $pmpro_countries);
--- a/includes/countries.php
+++ b/includes/countries.php
@@ -255,4 +255,7 @@
 
 	asort($pmpro_countries);
 
+	// Add the blank option at the start of the array, always.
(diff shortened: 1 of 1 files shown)


dparker1005 (commented):
Would we want to consider moving away from the `global $pmpro_country` and instead use helper functions when pulling this kind of data? Maybe something like `pmpro_get_countries( bool $include_choose_country = false )`?

This approach would have the benefit of not creating duplicate entries in Add Ons that are already "squeezing in" the "Select Country" option. Long term, having the `$include_choose_country` argument would also make it easy for devs to include or exclude the "Select Country" option when pulling the countries list.

andrewlimaza:
We should move to a function that gets the countries, and would love to have the `$include_choose_country` option. Tried to bandaid this solution for now until we rewrite/refactor this down the line.

I want to avoid doing this in each Add On that touches the countries list.

PR #3348, opened 2025-03-20
[BUGFIXING] Header and Footer are logging a warning due te lack of subject when save template on edit template page.

MaximilianoRicoTabo:
image

### All Submissions:

Check the `$_REQUEST['subject']` isset

1. Without this patch, go to the header or footer email template and save the template.
2. Check the logs.
3.

### Other information:

### Changelog entry

### DIFF
--- a/includes/email.php
+++ b/includes/email.php
@@ -200,10 +200,7 @@ function pmpro_email_templates_save_template_data() {
 	}
 
 	$template = sanitize_text_field( $_REQUEST['template'] );
-	$subject = '';
-	if( isset( $_REQUEST['subject'] ) ) {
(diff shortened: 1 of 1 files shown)


andrewlimaza (approved):

ideadude:
I'm going to merge this in.

It's maybe a bit confusing that we are saving a blank '' subject for the header and footer templates.

We could update the code to treat the header and footer templates differently and avoid trying to load or save that field. That might be a good idea, but also complicates the code a bit. I can't imagine we would ever want to save a subject for the header/footer, but I do think we might add other settings and want to overahaul this to save everything into one setting or at least one setting per email template.

For now, let's keep it as is with the updated code here to avoid the warning.

PR #3353, opened 2025-03-25
Adding functions for getting countries

dparker1005:
### All Submissions:

Adding getter functions for the list of countries and the default country instead of relying on globals. The new `pmpro_get_countries()` function also takes an optional parameter to include an empty "Choose a country" entry, useful for select country dropdowns.

### Other information:

### Changelog entry

### DIFF
--- a/includes/countries.php
+++ b/includes/countries.php
@@ -1,8 +1,14 @@
 <?php
-	//thanks JigoShop
-	global $pmpro_countries, $pmpro_default_country;
-	$pmpro_default_country = apply_filters("pmpro_default_country", "US");
 
+/**
+ * Get the list of PMPro countries.
+ *
+ * @since TBD
+ *
+ * @param bool $include_empty Whether to include an empty option in the list. Useful for select fields.
+ * @return array The list of countries with their codes as keys and names as values.
+ */
+function pmpro_get_countries( $include_empty = false ) {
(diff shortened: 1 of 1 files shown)


andrewlimaza (approved):

includes/countries.php
```diff
@@ -255,4 +261,37 @@
 
 	asort($pmpro_countries);
 
-	$pmpro_countries = apply_filters("pmpro_countries", $pmpro_countries);
+	/**
+	 * Filter the list of countries.
+	 *
+	 * @param array $pmpro_countries The list of countries with their codes as keys and names as values.
+	 */
+	$pmpro_countries = apply_filters( 'pmpro_countries', $pmpro_countries );
```
andrewlimaza, line 269: May be worth moving this before the return so the "select a country" is included.

dparker1005: I thought about that, but what held me back is that then we'd probably want to pass `$include_empty` to the `pmpro_countries` filter as well so that custom code knows whether the empty option should be here.

I thought that this solution was ok to ensure that the empty option is always present and at the beginning of the list when needed, but I'd be open to making that swap

andrewlimaza: That's okay with me leaving it as is. Excited for this one 👍

PR #3361, opened 2025-04-01
Fixing case where the report called in URL does not exist

kimcoleman:
### All Submissions:

If a user tries to access a report (via URL attribute) that doesn't actually exist, we would get a fatal error.

This PR makes sure the report slug called exists before trying to load the report function.

1. Add a non-existent report slug to your URL, like: `/wp-admin/admin.php?page=pmpro-reports&report=nope`
2. See the fatal error.
3. Pull this code and see no error.

### Other information:

### Changelog entry
* BUG FIX: Prevents a fatal error when accessing a non-existent report in the admin area.

### DIFF
--- a/adminpages/reports.php
+++ b/adminpages/reports.php
@@ -13,10 +13,14 @@ require_once( dirname( __FILE__ ) . '/admin_header.php' ); ?>
 <hr class="wp-header-end">
 
 <?php
-// View a single report if requested.
+$report_exists = false;
 if ( ! empty( $_REQUEST[ 'report' ] ) ) {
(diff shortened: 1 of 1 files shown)


ideadude:
Merging.

We weren't showing a warning before, but now that we are checking, we can consider showing a helpful message if the report is not found.

In practice this doesn't happen too often, but could happen especially for custom reports or ones added by Add Ons if that custom code or Add On is not active when someone visited a bookmarked URL or something.

I think for now we can just fail quietly here as we do in the PR.

PR #3364, opened 2025-04-04
Add new filter for login response messages

andrewlimaza:
* ENHANCEMENT: Adding in `pmpro_login_forms_handler_response_message` to allow filtering of success and error messages when logins failed or are successful.

Sample code:
```
function my_test_pmpro_login_message( $message, $msgt ) {
	if ( $msgt == 'pmpro_error' ) {
		$message = 'Whoops! Something went wrong. Please try again.';
	}

	return $message;
}
add_filter( 'pmpro_login_forms_handler_response_message', 'my_test_pmpro_login_message', 10, 2 );
```

Here is the screenshot with the filtered result.
Screenshot 2025-04-04 at 15 25 36

### All Submissions:

### DIFF
--- a/includes/login.php
+++ b/includes/login.php
@@ -457,6 +457,17 @@ function pmpro_login_forms_handler( $show_menu = true, $show_logout_link = true,
 		}
 	}
 
+	/**
+	 * Allow filtering of the form handler success and error messages. 
+	 * This includes the login form, password reset form, and the lost password form messages.
+	 * 
+	 * @since TBD
+	 * 
+	 * @param string $message The displayed message that is related to the form's response.
+	 * @param string $msgt The message type we can check against to know if it's a success or error message - (pmpro_error or pmpro_success).
+	 */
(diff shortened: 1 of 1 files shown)


dparker1005:
After further discussion, we are not adding this filter at this time.

The primary motivation for this PR was to be able to easily modify login error messages such as "The password you entered for the username %s is incorrect." which could confirm whether or not a username exists on a website; however, this information could also be obtained by beginning a checkout process with a username and seeing whether the checkout moves past the user creation step.

For sites that still want to alter login error messages, this can be done using the `gettext` filter:
https://developer.wordpress.org/reference/hooks/gettext/

For reference if we do consider this again in the future, it is worth considering moving this filter up to after the `switch ( sanitize_text_field( $_GET['action'] ) ) {` statement completes so that it specifically targets login failed error messages instead of also filtering log out, password reset, and other errors. This would allow us to pass the specific error reason into the filter (ex. `incorrect_password`) so that wording can easily be adjusted per error type.

PR #3368, opened 2025-04-08
Fix warning message when updating a plugin

### LINKED ISSUE #3323: HTML symbol not showing up because of esc_html()
**Describe the bug**
Some encoded HTML shows up because of esc_html() when you try to update a "Paid" Add On.

**To Reproduce**
Image

**Expected behavior**
Since esc_html() is required and other escape solutions are not real solutions, we can:
- change the > to something like a dash or a comma (not so cool, but...)
- or use str_replace after the escape to restore &gt; => `>` - hacky but cool.

- [x] This bug happens with a default WordPress theme active, or [Memberlite](https://memberlitetheme.com).

andrewlimaza:
* BUG FIX: Fixed a warning message that was encoding the `>` on output as well make the message more clear where the settings are to add a license key to PMPro.

It looks like it wasn't just the `esc_html` causing this, tried various other methods like wp_kses/wp_kses_post - WordPress itself may have been interfering with this message. Decoding the $msg after escaping it works. This will still keep the string safe but just decode allowed characters. `esc_html` will still strip out non-allowed HTML entities.

### All Submissions:

Resolves #3323 

Screenshot 2025-04-08 at 10 29 10

### Other information:

### DIFF
--- a/includes/addons.php
+++ b/includes/addons.php
@@ -516,8 +516,8 @@ function pmpro_admin_init_updating_plugins() {
 		$slug = str_replace( '.php', '', basename( $plugin ) );
 		$addon = pmpro_getAddonBySlug( $slug );
 		if ( ! empty( $addon ) && pmpro_license_type_is_premium( $addon['License'] ) && ! pmpro_can_download_addon_with_license( $addon['License'] ) ) {
-			$msg = sprintf( __( 'You must enter a valid PMPro %s License Key under Memberships > License to update this Add On.', 'paid-memberships-pro' ), ucwords( $addon['License'] ) );
-			echo '<div class="error"><p>' . html_entity_decode( esc_html( $msg ) ) . '</p></div>';
+			$msg = sprintf( __( 'You must enter a valid PMPro %s License Key in the PMPro Settings to update this Add On.', 'paid-memberships-pro' ), ucwords( $addon['License'] ) );
+			echo '<div class="error"><p>' . esc_html( $msg ) . '</p></div>';
 
 			// can exit WP now
 			exit;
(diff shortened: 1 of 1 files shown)


includes/addons.php
```diff
@@ -516,8 +516,8 @@ function pmpro_admin_init_updating_plugins() {
 		$slug = str_replace( '.php', '', basename( $plugin ) );
 		$addon = pmpro_getAddonBySlug( $slug );
 		if ( ! empty( $addon ) && pmpro_license_type_is_premium( $addon['License'] ) && ! pmpro_can_download_addon_with_license( $addon['License'] ) ) {
-			$msg = sprintf( __( 'You must enter a valid PMPro %s License Key under Settings > PMPro License to update this add on.', 'paid-memberships-pro' ), ucwords( $addon['License'] ) );
-			echo '<div class="error"><p>' . esc_html( $msg ) . '</p></div>';
+			$msg = sprintf( __( 'You must enter a valid PMPro %s License Key under Memberships > License to update this Add On.', 'paid-memberships-pro' ), ucwords( $addon['License'] ) );
+			echo '<div class="error"><p>' . html_entity_decode( esc_html( $msg ) ) . '</p></div>';
```
ideadude, line 520:
```suggestion
			$msg = sprintf( __( 'You must enter a valid PMPro %s License Key in the PMPro Settings to update this Add On.', 'paid-memberships-pro' ), ucwords( $addon['License'] ) );
			echo '<div class="error"><p>' . esc_html( $msg ) . '</p></div>';
```
Let's just rewrite to avoid the issue. High agency users know how to find the PMPro settings page and the license settings are called out pretty well in the header and sidebar.

PR #3372, opened 2025-04-10
Add function_exists checks to addons.php

andrewlimaza:
* BUG FIX: Fixes a race condition when activating Paid Memberships Pro after the Update Manager is already activated.

These functions were only wrapped in `if ( ! function_exists...` calls and does not change code whatsoever. The git diff shows that code was changed but it's a coincidence that the lines changed were the same conditionals/checks in 2 areas of the code base/change.

This resolves an issue with a future release of Update Manager being active and then activating Paid Memberships Pro.

### All Submissions:

1. Pull the latest `pmpro-update-manager` plugin and install it.
2. Deactivate PMPro core, and try to activate it again (see there's a fatal error).
3. Apply this patch, and repeat step 2.

### Other information:

### DIFF
--- a/includes/addons.php
+++ b/includes/addons.php
@@ -21,62 +21,64 @@ add_action( 'admin_init', 'pmpro_setupAddonUpdateInfo' );
  *
  * @since  1.8.5
  */
-function pmpro_getAddons() {
-	// check if forcing a pull from the server
-	$addons = get_option( 'pmpro_addons', array() );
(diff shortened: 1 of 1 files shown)


ideadude:
We still have code in place that will try to load the PMPro versions of these first, but this catches the case where the update manager is already active and PMPro is being activated. I think it's okay if we're running the "wrong" versions of these functions in this singular case.

PR #3376, opened 2025-04-25
Site change detection: hash current site url and start comparing hashes

mircobabini:
### All Submissions:

Previously, if the site change was made using wp-cli search-replace or similar (hosting migration tools working the same way), even the "current_site_url" option saved w/o obfuscation/hashing was updated. Then, the site change was not detected, and the pause mode not triggered.

Now, we are hasing the current site url, and comparing hashes.

1. Migrate a site using wp search-replace 'old url' 'new url' --all-tables --precise
2. The site change pause mode is not triggering, because the current site url option has been changed as well
3. Now, retrying after upgrading.

### Other information:

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -4680,26 +4680,32 @@ function pmpro_activating_plugin( $plugin = null ) {
 }
 
 /**
- * Compare the stored site URL with the current site URL
+ * Compare the stored site URL with the current site URL hash
  *
  * @since 2.10
- * @return bool True if the stored and current URL match
+ * @return bool True if the stored and current URL hashes match
  */
 function pmpro_compare_siteurl() {
-	$site_url = get_site_url( null, '', 'https' ); // Always get the https version of the site URL.=
-	$current_url = get_option( 'pmpro_last_known_url' );
+	$site_url = get_site_url( null, '', 'https' ); // Always get the https version of the site URL.
+	$site_url = str_replace( 'http://', 'https://', $site_url ); // We don't want to consider scheme, so just force https for this check.
 
-	// If we don't have a current URL yet, set it to the site URL.
-	if ( empty( $current_url ) ) {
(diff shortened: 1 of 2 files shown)


mircobabini:
The commit https://github.com/strangerstudios/paid-memberships-pro/pull/3376/commits/9a1a38818ed60ec6f3167f884943addc8515f735 is a lifesaver.
Otherwise, we have to deal with http/https comparisons at runtime... 

But, maybe still useful to have this in place for a while, just in case the db update fails for any reason?

```php
  if( $site_url_hash !== $current_url_hash ){
    // Check if the difference is just because of the http/https scheme.
      if( $current_url_hash === hash( 'sha256', str_replace( 'https://', 'http://', $site_url ) ) ){
        // Silent update, we now have the correct URL with https hashed for comparison.
	      update_option( 'pmpro_last_known_url', $site_url_hash );
	      return true;
      }
  }
```

mircobabini:
Of course the message must be changed. We don't have a readable copy of the old message now.
Or we can just change the encryption method to something we can reverse. No security related issues involved.
Maybe just base64?

Screenshot 2025-04-25 at 23 15 14

dparker1005:
I agree that using a reversible hashing method, like base64, is preferred so that we can reverse the encoding for the "URL Change Detected" message.

To avoid breaking any place that we get/update the `pmpro_last_known_url` option, what if we leave the `pmpro_compare_siteurl()` function unchanged and instead implement this with two filters:
- `pre_update_option_pmpro_last_known_url` - base64 encodes the value being set before saving in the database
- `option_pmpro_last_known_url` - Checks if the current value appears to be base64 encoded. If not, saves the value again so that it gets base64 encoded. Then decodes the URL and returns the decoded value

This way, any code that is already interacting with the `pmpro_last_known_url` option does not need changes to support the new encoded approach. Happy to discuss, let me know what you think

PR #3377, opened 2025-04-26
Support linking to a filtered result via URL attributes on the Add Ons admin screen.

kimcoleman:
### All Submissions:

Updates the Add Ons admin page to support passing the `s` parameter in the URL. You can pass in a single Add On slug or a query. Examples:

https://pmprokimfork.local/wp-admin/admin.php?page=pmpro-addons&s=restrict+access

https://pmprokimfork.local/wp-admin/admin.php?page=pmpro-addons&s=pmpro-courses

### Other information:

### Changelog entry
* ENHANCEMENT: Added support for passing a search term or Add On slug to pre-filter results on the Memberships > Add Ons admin page using a URL parameter `s`.

### DIFF
--- a/adminpages/addons.php
+++ b/adminpages/addons.php
@@ -59,7 +59,10 @@
 			</ul>
 			<div class="search-form">
 				<label class="screen-reader-text" for="search-plugins"><?php esc_html_e( 'Search Add Ons', 'paid-memberships-pro' ); ?></label>
-				<input type="search" name="s" id="search-add-ons" data-search="content" class="wp-filter-search" placeholder="<?php esc_attr_e( 'Search Add Ons...', 'paid-memberships-pro' ); ?>">
+				<?php
+					$pmpro_addon_search = isset( $_GET['s'] ) ? sanitize_text_field( wp_unslash( $_GET['s'] ) ) : '';
+				?>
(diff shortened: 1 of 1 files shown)


adminpages/addons.php
```diff
@@ -410,16 +415,13 @@
 					var view_items = $(`[data-search-${view}]`);
 					var view_val = $(this).data('view');
 
-					// Update the URL hash.
-					$( this ).attr( 'href' ).replace( /#/, '' );
```
andrewlimaza, line 414: Removed this line of code as it wasn't running anywhere and we used the `view_val` instead (retrieves the same value).

andrewlimaza:
@kimcoleman I submitted a commit to this PR to fix the clear search issue, and fixed another issue where the search was being cleared on key up every time until the `MIN_SEARCH_LENGTH` was met, this now looks at the raw value of the input and clears it out when there is 0 inputs. 

I also noticed that the search is cleared when you press 'enter' while the search field is in focus, this is due to the jQuery 'search' event and is now working correctly. It confused me a little but it is expected behavior.

To run these tests, and ensure the search input field is cleared take the following steps:

1. Enter "approvals" in the search.
2. Click on any of the terms "Premium" or "Free" and see that the search term "approvals" is removed.

The Add On results do reset also if doing the above steps, but for step 1 you use &s=approvals in the URL. However, the &s=approvals stays within the URL but does not negatively affect the search thereafter. We can use jQuery to clear any `&s=xxxx` from the URL within the clearSearch function.

PR #3387, opened 2025-05-14
allow single option select2 and remember the value if using ajax

johnorourke:
DRAFT!  Probably requires some discussion because:

- the documentation says 'multiple' defaults to false - not true for select2 fields
- the `select2options` option is undocumented

### All Submissions:

The `select2` custom field type allows for options to be fetched via ajax, but when doing this, the value is not preserved because it's not in the options array at the time the field is rendered.

Additionally, select2 defaults to being a multi-select field with no option to remove that.  The fix here is to allow the user to add `multiple: false` in the `select2options` setting making it backwards compatible with existing behaviour.  The documentation says the `multiple` field setting defaults to false, but this has no effect on `select2` fields.

### How to test the `multiple` change

1. create a `select2` field programatically, including an array of options
2. observe that it allows multiple options to be selected
3. in your field's settings, add `"select2options" => " multiple: false "`
4. observe that your field now only allows one option

### How to test the ajax change

1. create a `select2` field programatically with no `options` setting
2. add this setting in the field's parameters:

```
'select2options' => ' ajax: {
    url: "https://run.mocky.io/v3/1cbd61ca-ac99-4c22-a956-7045f790bc87",
},
minimumInputLength: 3 
 ',
```

3. in your field, begin typing "opt" and you should see it populate with data from [this mock](https://run.mocky.io/v3/1cbd61ca-ac99-4c22-a956-7045f790bc87)

### Improvements to this PR:

- should it support setting `multiple:false` in the actual field settings, instead of just `select2options` ?  (not backwards compatible - could break existing installations)
- should it be 2 separate PRs?
- should `select2options` be documented in the public documentation?

### Changelog entry

> Allow select2 fields to be non-multi-select by specifying " multiple:false " in the select2options field setting
> Allow select2 fields to have selected value preserved when fetching results via ajax

### DIFF
--- a/classes/class-pmpro-field.php
+++ b/classes/class-pmpro-field.php
@@ -1003,7 +1003,7 @@ class PMPro_Field {
 
 			//build multi select
 			$r = '<select id="' . esc_attr( $this->id ) . '" name="' . esc_attr( $this->name );
-			if(empty($this->select2options) || !preg_match('/multiple: *false/', $this->select2options)) {
+			if( empty( $this->select2options ) || ! preg_match( '/multiple: *false/', $this->select2options ) ) {
 				// backwards compatible - default to multi-select
 				$r .= '[]" multiple="multiple" ';
 			} else {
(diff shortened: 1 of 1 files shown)


dparker1005:
Thanks for opening this @johnorourke and for surfacing the doc/behavior mismatch. You're right that it was confusing.

For now, we're going to keep the current behavior (select2 and multiselect are always multi-select) and have updated the docs to match. The docs are now updated to clarify that multiple only applies to the select field type, and that select2 and multiselect always allow multiple selections:
https://www.paidmembershipspro.com/documentation/user-fields/field-types-attributes/#select

On the ajax value-preservation piece: in general we're trying not to add functionality to code-defined fields that isn't available through the Member Fields UI, so we're going to pass on that change. select2options is an advanced setting for custom functionality, and any further customization (like handling preselected values when loading options via ajax) is something developers can handle on the JS side as needed.

I'm going to close this out, but we really appreciate the writeup since it's what prompted the doc fix.

johnorourke:
Thanks for the detailed response @dparker1005 - that sounds like a good solution for keeping the product less complex and reduce future maintenance 👍

PR #3403, opened 2025-06-10
Sync stripe application fees before payment

dparker1005:
### All Submissions:

Removes the Stripe application fee for payments after a PMPro license is activated and adds the fee when a license is removed.

### Other information:

### Changelog entry


ideadude:
We should show a warning to folks who are either using the fee filter or have the legacy 1% fee.

Different messages for each set of folks. The legacy folks probably never even thought about it, and so we should have some more care with them.

The filter folks aren't bad people, but they (or their developer) did either install some code or a plugin to reduce the fees. (Really doubt anyone raised the fees through code.)

Something like: 

> Accounts connected through Stripe Connect require a 2% fee on all transactions OR a premium PMPro license.
> 
> Your site is currently paying a __% fee. This is due to custom code on your site. (Or This is a lower fee that was in place at the time you connected.)
> 
> A future update will force this fee to 2%. To reduce the fee to 0, you may either purchase and activate a premium PMPro license or disconnect and set up your own restricted keys. Or you can continue to use Stripe through our Connect Account as usual with the new fee.

PR #3419, opened 2025-06-26
Fixing object cache issues

ideadude:
### All Submissions:

1. Set up object caching with something like W3 Total Cache.
2. Add the login form widget to the sidebar of a page.
3. Clear all cache.
4. Visit that page while logged in.
5. Visit that page while logged out.
6. Note that the logged in version of the widget will show up instead of the logged out one.

### Other information:

Note that I removed the escaping of the cached version of the widget. We will want to add some escaping back that supports form elements.

We can't rely on wp_cache_flush_group to work.

We maybe shouldn't be caching this at all. It's not too heavy and probably already covered by page caches.

Do we want some other way to track the cache key/groups we create so we can flush better?

Are there other times to flush here?

There are some other spots in our code that are using wp_cache_get/set. We should make sure those are working well too.

### Changelog entry

> BUG FIX/ENHANCEMENT: Fixed issues with the login form widget when using object caching.

### DIFF
--- a/includes/widgets.php
+++ b/includes/widgets.php
@@ -26,8 +26,15 @@ class PMPro_Widget_Member_Login extends WP_Widget {
 
 	function widget( $args, $instance ) {
 		$cache = array();
+		$cache_key_parts = array(
+			'args'            => $args,
+			'logged_in'       => is_user_logged_in(),
+			'current_user_id' => get_current_user_id(),
+			'is_login_page'   => pmpro_is_login_page(),
+		);
+		$cache_key = 'pml_' . md5( wp_json_encode( $cache_key_parts ) );
 		if ( ! $this->is_preview() ) {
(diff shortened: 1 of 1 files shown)


ideadude:
Thinking to just remove the cache. We likely included it because it was part of the widget code we referenced when writing this, but the login form and info is not an intense bit of code that would really need to be cached, even though it potentially shows up on every page. Honestly, checking and loading the cache might end up taking more time than checking if the user is logged in and grabbing user info that is likely loaded anyway. It also will get cached well by page caches.

Here is a branch that simply removes the cache: https://github.com/strangerstudios/paid-memberships-pro/tree/fix/remove-login-widget-cache

We're using that on our live site now.

kimcoleman:
Closing in favor of the merged code here that Jason mentions: https://github.com/strangerstudios/paid-memberships-pro/pull/3423

PR #3424, opened 2025-07-03
Dashboard: Tweaks for small screens

dalemugford:
Tweaks for welcome and quick links boxes, and postbox buttons on mobile.

Screenshot 2025-07-03 at 1 44 55 PM


kimcoleman (approved):
Looking great - we could tweak this endlessly but right now I am happy with how it looks as the screen size changes. Thank you!

PR #3461, opened 2025-08-06
WP Fusion Compatibility

JarrydLong:
### All Submissions:

Implements compatibility with WP Fusion. 

If WP Fusion Lite is enabled, we'll enable this by default. 

If WP Fusion Pro is enabled, we'll default to the Pro version's PMPro functionality and not load our own.

Resolves XXX.

1. Install WP Fusion Lite
2. Select a CRM (I tested with MailPoet)
3. Fill in settings on a level under the WP Fusion section and run a test checkout with that level
4. WP Fusion should subscribe the member to a Mailpoet list in this case

### Other information:

### DIFF
--- a/includes/compatibility.php
+++ b/includes/compatibility.php
@@ -78,6 +78,9 @@ function pmpro_compatibility_checker() {
 			include_once( PMPRO_DIR . '/includes/compatibility/' . $value['file'] ) ;
 		}
 	}
+
+    include_once( PMPRO_DIR . '/includes/compatibility/wp-fusion.php' );
+
 }
 add_action( 'plugins_loaded', 'pmpro_compatibility_checker' );
 
--- a/includes/compatibility/wp-fusion.php
+++ b/includes/compatibility/wp-fusion.php
@@ -1,0 +1,37 @@
+<?php
+/**
+ * WP Fusion Compatibility
+ * 
+ * @since TBD
+ */
+
+// Running this after plugins_loaded to ensure WP Fusion is loaded.
+function pmpro_wpfusion_compatibility() {
+
+    /**
+     * WP Fusion Lite is not active, we shouldn't run anything WP Fusion related for now.
+     */
+    if( ! class_exists( 'WP_Fusion_Lite' ) ) {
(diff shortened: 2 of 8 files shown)


andrewlimaza (approved):

includes/compatibility/wp-fusion-lite.php
```diff
@@ -0,0 +1,22 @@
...
+ * @since TBD
+ */
+
+// WP Fusion Pro is active, then let's give WP Fusion Pro priority to include their own PMPro integration code.
+if ( class_exists( 'WP_Fusion' ) ) {
+	return;
+}
+
+// Include the WP Fusion PMPro specific classes.
+include_once( PMPRO_DIR . '/includes/lib/wp-fusion/class-base.php' );
```
dparker1005, line 17: Looking at the `wp-fusion-lite` code, it looks like the `WPF_Integrations_Base` class exists in that plugin and is loaded under rare circumstances via the `WP_Fusion_Lite->integrations_includes()` method.

I don't think that method would ever actually run in practice, but if it did, we could fatal error for redeclaring the `WPF_Integrations_Base` class. This could also happen if any other plugin integrates in the same way that we are by redeclaring the `WPF_Integrations_Base` class separately.

To be safe, I think it would be worth either first checking that the class does not exist before declaring our own (which fixes the potential conflict assuming WP Fusion Lite loads their file first) or, even better, calling `require_once WPF_DIR_PATH . 'includes/integrations/class-base.php';` to only ever load `wp-fusion-lite`'s version of the file exactly once (in which case the load order does not matter)

andrewlimaza: After discussion, we're going to load the class-base.php locally, and prefix/rename the class to avoid any conflicts.

andrewlimaza: See commit - https://github.com/strangerstudios/paid-memberships-pro/pull/3461/commits/e529eea05fb64b34bc4cbddb1d5aa52cdd6d58b8

PR #3467, opened 2025-08-15
Initial work toward the admin view of orders

kimcoleman:
### All Submissions:

Adds a new View for orders in admin before the "edit" step.

### Other information:

### Changelog entry
* FEATURE: Added new View of orders in the admin.


kimcoleman:
I don't really have a strong feeling that it's the right or wrong move. I think green would be good. Do you think green and first? Green and where it is? It's nice if "most" things are the same order to order, then "extra" things before "delete" always? If we don't expect anything else to squeeze in, and since once mark as paid is complete there is no "undo", I think putting it first and green is a good option.

I won't do that step yet, but we probably need the CSS for it. I don't know of a WP-native "green" button (like the destructive selector for the delete button there).

PR #3510, opened 2025-10-15
Defensive coding in PMPro_AddOns class to simplify duplicating code in Update Manager Add On

dparker1005:
### All Submissions:

Checking for the `PMPRO_DIR` constant before referencing it in case this code is being loaded in the Update Manager Add On where that constant will not always be set.

### Other information:

### Changelog entry

### DIFF
--- a/classes/class-pmpro-addons.php
+++ b/classes/class-pmpro-addons.php
@@ -354,6 +354,8 @@ class PMPro_AddOns {
 			if ( version_compare( $plugin_data['Version'], $addon['Version'], '<' ) ) {
 				$value->response[ $plugin_file ]              = $this->get_plugin_API_object_from_addon( $addon );
 				$value->response[ $plugin_file ]->new_version = $addon['Version'];
+
+				// If we have an icon to show, add it to the response. Otherwise let it show the default icon.
 				$icon = $this->get_addon_icon( $addon['Slug'] );
 				if ( ! empty( $icon ) ) {
 					$value->response[ $plugin_file ]->icons = array( 'default' => esc_url( $icon ) );


classes/class-pmpro-addons.php
```diff
@@ -907,9 +910,14 @@ private function send_ajax_result( $result ) {
 	 * @since 2.8.x
 	 *
 	 * @param string $slug The identifying slug for the addon (typically the directory name).
-	 * @return string $plugin_icon_src The src URL for the plugin icon.
+	 * @return string|false $plugin_icon_src The src URL for the plugin icon or false if PMPRO_DIR is not defined.
 	 */
 	public function get_addon_icon( $slug ) {
+		// If PMPRO_DIR is not defined, bail. This may happen in the Update Manager Add On.
+		if ( ! defined( 'PMPRO_DIR' ) ) {
+			return false;
```
ideadude, line 920: Instead of returning false, we could return the default in the get_addon_icon method. That way we don't have to check for the false return above.

classes/class-pmpro-addons.php
```diff
@@ -354,7 +354,10 @@ public function update_plugins_filter( $value ) {
 			if ( version_compare( $plugin_data['Version'], $addon['Version'], '<' ) ) {
 				$value->response[ $plugin_file ]              = $this->get_plugin_API_object_from_addon( $addon );
 				$value->response[ $plugin_file ]->new_version = $addon['Version'];
-				$value->response[ $plugin_file ]->icons       = array( 'default' => esc_url( $this->get_addon_icon( $addon['Slug'] ) ) );
+				$icon = $this->get_addon_icon( $addon['Slug'] );
```
ideadude, line 359: Seems like this would not set `$value->response[ $plugin_file ]->icons` if PMPro was active??

Issue #767, opened 2018-10-15
Help people transition from plugins they've downloaded on GitHub (-master)


ideadude:
When you download a repo branch on github (vs a release), it downloads into a folder called pmpro-addon-master instead of pmpro-addon.

We can try to get better about maintaining tags/releases for all add-ons and direct people to use that, but I think most will still click the first download link they see on the homepage.

Maybe we can add code to PMPro core to detect these -master add-ons and either rename the directory (and reactivate the plugin since it will look like it moved) -- which is scary because deactivating a plugin can cause fatal errors -- or at least show a notice somewhere to the admin saying hey "you should probably rename this directory and reactivate the plugin". Maybe show that message along with a button to try doing it (might not have permission anyway) and a big warning RE possibly causing a fatal error.

dparker1005:
Thanks for opening this. Closing this as addressed.

PMPro 3.1 added detection for installed Add Ons with incorrect folder names, including cases where an Add On was installed from a GitHub branch download into a `-master` style folder. PMPro 3.7 also added a Site Health check that flags these incorrect folder names and points admins to the PMPro Site Health details.

We are not planning to automatically rename plugin folders from core because that can be risky on active sites, but the current checks should help admins identify and fix the issue manually.

PR #3553, opened 2025-12-23
Added Site Health Notice for incorrect folder names

andrewlimaza:
* ENHANCEMENT: Added Site Health critical issue notification if plugins are using development/GitHub slugs.

Resolves: https://github.com/strangerstudios/paid-memberships-pro/issues/767

Marking this as a draft PR as we will want to reword it and possibly add more notices for other conflicts we find.

Screenshot 2025-12-23 at 13 02 59


classes/class-pmpro-site-health.php
```diff
@@ -585,7 +594,14 @@ function get_add_ons_with_incorrect_folder_name() {
 			$errors[] = "{$addon['Name']} ( {$addon['plugin']} => {$installed_name} )";
 		}
 
-		return empty( $errors ) ? __( 'No add ons with incorrect folder names detected.', 'paid-memberships-pro' ) : implode( " | \n", $errors );
+		if ( ! empty( $errors ) ) {
+			self::$file_errors = true;
```
dparker1005, line 634: Instead of using a static variable to track whether there was a file error, how do we feel about updating the `get_add_ons_with_incorrect_folder_name()` method to take a parameter `$return_bool = false` that, if enabled, will return true if there are file errors and false if there are no errors? This would allow our new code to just call `self::get_add_ons_with_incorrect_folder_name( true )` instead of checking the static variable

dparker1005: I'm actually going back on this. The fact that the $file_errors var is private means that there is no code debt. Adding a function parameter would add code debt since that method is public. I'm happy sticking with the private variable.

PR #3554, opened 2025-12-23
Add week-passed cleanup to Library Conflicts

dalemugford:
### All Submissions:

Kim Coleman found this issue and surfaced it on Slack on Dec 23rd. Essentially she was seeing conflicts listed that pointed to inactive plugins, revealing that our tracker was holding on to stale conflicts.

Library conflicts are tracked prior to PMPro loading its libraries, and stored in an option in the database. They are referenced/displayed in Site Health. However, _there's no explicit code to cleanup stale library conflicts that may have been resolved through deactivation, or plugin update, etc._

This PR adds a small amount of code to loop through and unset entries older than 7 days when `pmpro_track_library_conflict` is invoked.

If we're particularly concerned about doing too much work here, we could alternatively schedule a cleanup to run separately either with Action Scheduler or hook to a Site Health action.

```
	// Remove conflicts if older than 7 days.
	foreach ( $library_conflicts as $lib_name => $paths ) {
		foreach ( $paths as $lib_path => $info ) {
			if ( ! empty( $info['timestamp'] ) && ( strtotime( $info['timestamp'] ) < strtotime( '-7 days', strtotime( $now ) ) ) ) {
				unset( $library_conflicts[ $lib_name ][ $lib_path ] );
			}
		}
	}
```

### Other information:

### Changelog entry

> Paid Memberships Pro now removes plugin library conflicts tracked over a week ago from its conflict tracker.

### DIFF
--- a/includes/compatibility.php
+++ b/includes/compatibility.php
@@ -199,6 +199,7 @@ function pmpro_track_library_conflict( $name, $path, $version ) {
 	$library_conflicts[ $name ][ $path ]['timestamp'] = $now;
 
 	// Cleanup.
+	$seven_days_ago = strtotime( '-7 days' );
 	foreach ( $library_conflicts as $lib_name => $paths ) {
 		// If no paths, remove the library entry.
 		if ( empty( $paths ) ) {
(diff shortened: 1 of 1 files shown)


includes/compatibility.php
```diff
@@ -196,6 +196,21 @@ function pmpro_track_library_conflict( $name, $path, $version ) {
...
+		if ( empty( $paths ) ) {
+			unset( $library_conflicts[ $lib_name ] );
+			continue;
+		}
+		// Check each path's timestamp. If older than 7 days, remove it.
+		foreach ( $paths as $lib_path => $info ) {
+			if ( ! empty( $info['timestamp'] ) && ( strtotime( $info['timestamp'] ) < strtotime( '-7 days' ) ) ) {
+				unset( $library_conflicts[ $lib_name ][ $lib_path ] );
+			}
+		}
```
ideadude, line 214: If all paths are old and `$library_conflicts[ $lib_name ]` becomes empty from that foreach, do we want to unset `$library_conflicts[ $lib_name ]` at the bottom here?

dalemugford: Yes, good catch. After the foreach removes all the stale paths, we'd be left with an empty array for that library. Adding a cleanup check after the loop would keep things tidy, e.g.

```
if ( empty( $library_conflicts[ $lib_name ] ) ) {
    unset( $library_conflicts[ $lib_name ] );
}
```

dalemugford: Will add that.

PR #3576, opened 2026-02-10
Fix login error messages not displaying when redirect_to is present

sacrefizz:
## Summary

When a user is redirected to the PMPro login page with a `redirect_to` parameter (e.g., after trying to access a members-only page), failed login attempts do not show error messages on the PMPro login page.

### Root cause

The `redirect_to` URL is passed directly to `wp_login_form()` via the `redirect` argument, which embeds it as a hidden `redirect_to` field in the form. When the form POSTs to `wp-login.php`, this `redirect_to` value interferes with the authentication flow, causing the error messages to be lost instead of being displayed on the PMPro login page.

### Fix

This PR separates PMPro's redirect handling from `wp_login_form()`'s `redirect_to` hidden field:

- **`pmpro_login_form()`**: Extracts the redirect URL from the args before calling `wp_login_form()`, and adds it as a separate `pmpro_redirect_to` hidden field via the `login_form_bottom` filter.
- **`pmpro_login_redirect()`**: Checks `pmpro_redirect_to` first for successful login redirects.
- **`pmpro_login_failed()`**: Checks `pmpro_redirect_to` first when preserving the redirect URL across failed login attempts, with a fallback to `redirect_to` for backwards compatibility.

### How to reproduce

1. Set up a PMPro login page and a members-only page
2. Visit the members-only page while logged out — you get redirected to `/login/?redirect_to=%2Fmembers-page%2F`
3. Enter an incorrect password
4. **Before fix**: No error message is displayed
5. **After fix**: The appropriate error message is displayed (e.g., "The password you entered for the username X is incorrect")

## Test plan

- [x] Visit a members-only page while logged out, verify redirect to login page
- [x] Enter wrong password — error message should display
- [x] Enter wrong username — error message should display
- [x] Enter correct credentials — should redirect to the original members-only page
- [x] Visit login page directly (no redirect_to) — verify login still works normally
- [x] Test with login widget — verify widget login still works

### DIFF
--- a/includes/login.php
+++ b/includes/login.php
@@ -17,6 +17,13 @@ function pmpro_login_redirect( $redirect_to, $request = NULL, $user = NULL ) {
 
 	$is_logged_in = ! empty( $user ) && ! empty( $user->ID );
 
+	// Check for PMPro's separate redirect field, which is used to prevent
+	// login error messages from being lost on failed login attempts.
+	// See pmpro_login_form() for where this field is added.
+	if ( $is_logged_in && ! empty( $_REQUEST['pmpro_redirect_to'] ) ) {
+		$redirect_to = esc_url_raw( wp_unslash( $_REQUEST['pmpro_redirect_to'] ) );
+	}
+
 	if ( $is_logged_in && empty( $redirect_to ) ) {
 		// Can't use the pmpro_hasMembershipLevel function because it won't be defined yet.
 		$is_member = $wpdb->get_var( "SELECT membership_id FROM $wpdb->pmpro_memberships_users WHERE status = 'active' AND user_id = '" . esc_sql( $user->ID ) . "' LIMIT 1" );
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
@/tmp/pr-review-body.md

flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3576 — Fix login error messages not displaying when redirect_to is present**
*sacrefizz → dev | 1 file, +37 -1*
*https://github.com/strangerstudios/paid-memberships-pro/pull/3576*

**Summary**
Correct root cause diagnosis and a clean fix. Separating PMPro's redirect tracking from `wp_login_form()`'s native `redirect_to` field is the right approach. The round-trip logic holds: failed login embeds the URL as a query param back to the PMPro login page, which re-injects it into the `pmpro_redirect_to` hidden field on re-render. Ready to merge with two minor items noted below.

**Issues**

- **Minor** `includes/login.php:~20` (`pmpro_login_redirect`) — `pmpro_redirect_to` is accepted from `$_REQUEST` and overwrites `$redirect_to` without validating the host. WordPress core downstream uses `wp_safe_redirect()` which prevents the open redirect in practice, but the `pmpro_login_redirect_url` filter exposes the raw value to callers before that safety net kicks in. Defensive fix: wrap with `wp_validate_redirect( $value, home_url() )` before assigning.

- **Minor** `includes/login.php:~1095` (`pmpro_login_failed`) — The `pmpro_redirect_to` read applies `wp_unslash()`, but the old `redirect_to` fallback added two lines later does not. The pre-existing code didn't have it either, so this isn't a regression, but since you're touching this line, worth making consistent:
  ```php
  $redirect_to = ( ! empty( $_REQUEST['redirect_to'] ) ) ? esc_url_raw( wp_unslash( $_REQUEST['redirect_to'] ) ) : '';
  ```

**Looks Good**
- The closure approach for `login_form_bottom` is correct — storing a reference to the closure and passing it to `remove_filter()` is exactly how anonymous filter cleanup works in PHP. No filter leakage between multiple `pmpro_login_form()` calls on the same page.
- `unset( $args['redirect'] )` before passing to `wp_login_form()` is the cleanest way to prevent the hidden field collision without patching core behavior.
- Backwards compatibility preserved in `pmpro_login_failed()` via the `redirect_to` fallback.

dparker1005:
Hi there @sacrefizz,

I just wanted to follow to confirm whether you have been able to replicate this behavior on your site with only PMPro active as we still have not been able to replicate the missing error message issue locally.

Without being able to replicate this issue, we are unable fully understand the cause of this behavior to confirm that whatever fix that we ship addresses the core issue.

dparker1005:
Closing this out since we haven't been able to reproduce the issue and haven't heard back.

@sacrefizz, if you're able to reproduce with only PMPro active, we can reopen this PR and take another look.

PR #3620, opened 2026-03-20
Add Divi 5 compatibility for PMPro module-level restrictions

JarrydLong:
### All Submissions:

Adds support for Divi 5 and migrates the old values from Divi 4 to the new fields

- Detects Divi 5 via et_builder_d5_enabled() and registers D5-specific hooks
- Server-side condition evaluation via divi_module_options_conditions_is_custom_condition_true
- No-access message injection via divi_module_wrapper_render
- REST endpoint GET /divi/v1/pmpro/membership-levels for the VB level selector
- pmpro-divi-5.js registers the pmproMembershipLevel condition type in the D5 VB conditions panel
- D4 to D5 migration converts legacy PMPro attrs to native D5 conditions on upgrade
- D4 hooks remain active in both modes to support shortcode-module fallback content

image

1. Restrict a row in Divi 4 as normal
2. Upgrade to Divi 5
3. Edit the same element. It should move the old settings into the new Advanced > Conditions section in the Divi 5 Editor

### Other information:

### DIFF
--- a/includes/compatibility/divi.php
+++ b/includes/compatibility/divi.php
@@ -1,18 +1,348 @@
 <?php
 
-class PMProDivi{
+class PMProDivi {
 
-	function __construct(){
+	function __construct() {
 
-		if ( empty( $_GET['page'] ) || 'et_divi_role_editor' !== $_GET['page'] ) {
-			add_filter( 'et_builder_get_parent_modules', array( __CLASS__, 'toggle' ) );
-			add_filter( 'et_pb_module_content', array( __CLASS__, 'restrict_content' ), 10, 4 );
-			add_filter( 'et_pb_all_fields_unprocessed_et_pb_row', array( __CLASS__, 'row_settings' ) );
-			add_filter( 'et_pb_all_fields_unprocessed_et_pb_section', array( __CLASS__, 'row_settings' ) );			
+		$is_d5 = function_exists( 'et_builder_d5_enabled' ) && et_builder_d5_enabled();
+
+		if ( $is_d5 ) {
+
+			// Divi 5: use the native Display Conditions system for hide/show logic.
+			// The condition name 'pmproMembershipLevel' is evaluated server-side here.
+			add_filter( 'divi_module_options_conditions_is_custom_condition_true', array( __CLASS__, 'd5_condition_check' ), 10, 4 );
+
+			// Divi 5: intercept module wrapper output to inject the no-access message.
+			// This runs before the displayable check so we can replace the content rather
+			// than return an empty string when showNoAccessMessage is enabled.
+			add_filter( 'divi_module_wrapper_render', array( __CLASS__, 'd5_no_access_message' ), 1, 2 );
+
+			// Divi 5: enqueue the VB conditions UI script and register the membership
+			// levels REST endpoint used by that script.
+			add_action( 'divi_visual_builder_assets_before_enqueue_scripts', array( __CLASS__, 'd5_enqueue_vb_script' ) );
+			add_action( 'rest_api_init', array( __CLASS__, 'd5_register_rest_routes' ) );
+
+			// Divi 5: D4 → D5 migration. Mark our D4 attrs as legacy so they don't
+			// get preserved as unknownAttributes (which would keep the module as a
+			// shortcode-module fallback). Then convert them to native D5 conditions.
+			add_filter( 'divi.conversion.legacyAttributeNames', array( __CLASS__, 'd5_legacy_attribute_names' ) );
+			add_filter( 'divi.conversion.postConvertAttrs', array( __CLASS__, 'd5_migrate_d4_attrs' ), 10, 4 );
+
 		}
-		
+
+		// Always register the D4 hooks for two reasons:
+		//   1. When Divi 4 is the active builder (! $is_d5).
+		//   2. When Divi 5 is active but a page contains migrated D4 content — Divi 5 wraps
+		//      rows/sections that have unknown third-party attributes (like paid-memberships-pro)
+		//      in a `divi/shortcode-module` fallback block which bootstraps the full D4 shortcode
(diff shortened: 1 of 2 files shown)


flintfromthebasement (changes requested):
**PR: strangerstudios/paid-memberships-pro#3620 — Add Divi 5 compatibility for PMPro module-level restrictions**
*JarrydLong → dev | 2 files, +699 -45*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3620>

**Summary**

Solid implementation of Divi 5 conditions integration. The overall architecture — condition check hook, wrapper render swap, REST endpoint, JS VB component, and D4→D5 migration — is well thought out. Copilot's bot review already flagged several issues inline (REST namespace, hidden levels, empty `levelIds` fallback, `response.ok` guard, multi-select staleness, `strpos` strict check). Those are all real; this review focuses on severity assessment and two findings not yet flagged.

**Issues**

- **Critical** `includes/compatibility/divi.php:107-112` — The pattern of returning `true` from `d5_condition_check` when `$should_display === false && showNoAccessMessage === 'on'` is a content-leak risk. `d5_no_access_message` (the hook that replaces the output) is scoped to `divi/row` and `divi/section` only, but `d5_condition_check` has no such scope restriction. If a user applies a `pmproMembershipLevel` condition with `showNoAccessMessage=on` to any other module type, the condition check lies to Divi (`true` = access granted), Divi renders the full restricted content, and `d5_no_access_message` never fires to swap it out. The fix: either guard `d5_no_access_message` to handle all module types, or if that's not feasible given Divi's render pipeline, only return `true` for module types where you can guarantee the wrapper filter runs. Accept that `showNoAccessMessage` will not work outside row/section and document that constraint. Copilot's inline comment (#7) flags the same issue but undersells the severity — this is a content disclosure bug, not just a UX gap.

- **Major** `includes/compatibility/divi.php:196-205` — REST route registered under `divi/v1` namespace. PMPro owns `pmpro/v1` — use that (or `pmpro-divi/v1`). Registering under a third-party namespace is fragile: Divi could claim that path, another plugin could conflict, and it obscures ownership. Also flagged by Copilot.

- **Major** `includes/compatibility/divi.php:224-227` — `pmpro_getAllLevels()` defaults to `$include_hidden = false`, so the REST endpoint omits hidden/inactive levels. A site may have row restrictions set to a hidden level; those levels won't appear in the D5 VB selector and will be effectively uneditable without knowing the ID. Should be `pmpro_getAllLevels( true )`. Also flagged by Copilot.

- **Minor** `includes/compatibility/divi.php` / `js/pmpro-divi-5.js` — `enableCondition` is included in `conditionSettings` and exposed in the VB UI, but the PHP `d5_condition_check` never reads it. If a user sets `enableCondition = 'off'`, the condition still evaluates. Either enforce it server-side in `d5_condition_check` or remove the field — right now it's a UI control that does nothing.

- **Minor** `includes/compatibility/divi.php:152` — `d5_no_access_message` replaces the full module wrapper output (hooked into `divi_module_wrapper_render`) with raw `pmpro_get_no_access_message()` HTML. The no-access message HTML won't include Divi's wrapper `<div>` structure, which could break column/grid layout depending on where the row sits. D4's `restrict_content` operated on inner module content, not the wrapper. Worth testing with a restricted row inside a multi-column section.

**Questions**

1. Does `divi_module_wrapper_render` fire before or after Divi has already determined the condition result is `true`? The commit comment says Divi skips rendering entirely if the condition fails — but is there a separate Divi hook that runs for failed conditions, perhaps for injecting fallback content, that would let you drop the `return true` workaround?

2. For the D4→D5 migration in `d5_migrate_d4_attrs`: is this a one-time conversion that fires during post save, or does it run on every page load? If it runs repeatedly it should be idempotent (i.e., if a D5 condition for PMPro already exists, don't add a duplicate).

andrewlimaza:
To test this functionality in Divi 5:

1. Navigate to a page that is already built with Divi or create a new page using the Divi Builder
2. Select the settings "cog" icon for a row or element. If this doesn't show up, on the left hand sidebar while in edit mode there is an option "Action Icons on Hover" that needs to be selected to show this cog icon per element.
3. On the right hand side sidebar, once selecting the cog icon from step 2, choose "Advanced" > "Conditions". There should be a prefilled condition when upgrading from Divi 4 to Divi 5 that already had PMPro membership content restriction as a "custom condition".

Note that you can create multiple conditions within a single element as part of Divi 5's functionality, we will definitely see customers running into issues with this and not totally knowing where or what is being restricted correctly. This is more of a Divi UI issue but so far the PR works in keeping/migrating content restriction by moving it to custom conditions. We'll work on Flint's feedback as well.

dparker1005:
To clean up this code, we should be able to leverage the existing `pmpro_apply_block_visibility()` function vs re-coding similar logic specifically for Divi. That function was initially built for the block editor, but is generally applicable for these kinds of situations. For example, our Elementor integration uses that function when determining if a user has access to a specific element or when rendering that element.

I'll be taking a closer look at this over the next couple days

dparker1005:
**Edge case worth documenting:** the no-access message behavior when `pmproMembershipLevel` is combined with *other* Divi conditions (AND / OR / date-time / user-role / etc.) is driven purely by the PMPro condition's own pass/fail state, not by Divi's aggregate displayability. So a logged-in user who doesn't have Level 1 will see the PMPro no-access message regardless of whether some OR'd sibling condition would have let the module through, or whether an AND'd condition is also failing. That's arguably the correct read of "Show No Access Message" — the admin explicitly opted in to telling users about the PMPro restriction — but it's worth mentioning in the docs so admins know the message is scoped to the PMPro rule, not to the module's overall visibility.

**Minor quirk:** if a module has two `pmproMembershipLevel` conditions OR'd together with different `showNoAccessMessage` settings, the first one in array order drives the injected message, not the one Divi's conflict-resolution would pick (same-type OR → later wins). Rare setup; not fixing.

PR #3628, opened 2026-03-31
Hide recurring trial settings behind gateway supports check

dparker1005:
## Summary
- Add `recurring_trials` declaration to the gateway `supports()` method for gateways that support charging a different amount for N billing cycles (PayPal Express, PayPal WPP, Authorize.net).
- Hide the Custom Trial settings on the edit level page unless the active gateway declares `recurring_trials` support or the level already has a trial configured.
- When a level has an existing trial but the gateway does not support it, show a generic warning replacing the old hardcoded per-gateway messages.

## Context
The PMPro "Custom Trial" settings allow charging a different recurring amount for the first N billing cycles (e.g., $40/month for 3 months, then $50/month). This is distinct from a simple free trial delay.

The only gateways that ever supported this were:
- **PayPal Express** — deprecated
- **PayPal Website Payments Pro (WPP)** — deprecated
- **Authorize.net** — deprecated

Other gateways like Stripe, Braintree, and Payflow Pro only ever supported free trial delays (`trial_period_days`, `trialPeriod`, profile start date), not different amounts for N cycles. The edit level page previously showed per-gateway warnings for these, but since all gateways that actually use recurring trials are now deprecated, this is the first step toward cleaning up the existing behavior so that we can potentially rebuild trials fresh in an upcoming update.

## Test plan
- [ ] Set gateway to Stripe — verify trial settings are hidden on a new level
- [ ] Set gateway to Authorize.net — verify trial settings are visible
- [ ] Set gateway to Stripe with an existing level that has trial settings — verify the settings show with the red warning
- [ ] Verify the `pmpro_membership_level_after_trial_settings` action still fires regardless of gateway

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/adminpages/discountcodes.php
+++ b/adminpages/discountcodes.php
@@ -6,7 +6,7 @@
 	}
 
 	//vars
-	global $wpdb, $pmpro_currency_symbol, $pmpro_stripe_error, $pmpro_braintree_error, $pmpro_payflow_error, $pmpro_twocheckout_error, $pmpro_pages, $gateway;
+	global $wpdb, $pmpro_currency_symbol, $pmpro_stripe_error, $pmpro_braintree_error, $pmpro_pages, $gateway;
 
 	$now = current_time( 'timestamp' );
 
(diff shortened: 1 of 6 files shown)


ideadude:
I tested this on `flintdev.test` with the site pointed at this PR branch (`review-pr-3628`) and the active gateway set to Stripe. I can confirm a few issues that should be addressed before merge:

1. **Discount code save can fatal when trial fields are hidden.**
   With Stripe active, the discount code edit screen does not render `trial_amount[]` / `trial_limit[]`. Saving a discount code then hits `adminpages/discountcodes.php:145-146`, where those request keys are passed directly into `array_map()`:

   ```text
   Undefined array key "trial_amount" in discountcodes.php:145
   Fatal error: array_map(): Argument #2 ($array) must be of type array, null given
   ```

   There is also a related alignment risk here: the save loop uses `$n = array_search( $level_id, $all_levels_a )` and then reads `$trial_amount_a[$n]`. Since the PR conditionally removes the trial inputs while `all_levels[]` still includes every level, mixed rendered/unrendered rows can misalign those parallel arrays.

2. **Membership level save logs undefined-key warnings.**
   With Stripe active, the level edit screen correctly hides `custom_trial` and `trial_amount`, but `adminpages/levels/save-level.php:29-30` still reads these directly:

   ```text
   Undefined array key "trial_amount" in save-level.php:29
   Undefined array key "trial_limit" in save-level.php:30
   ```

   This can probably be fixed by defaulting those request values before sanitizing, or by keeping hidden default inputs when the UI is hidden.

3. **The trial settings table still has invalid raw markup.**
   With Authorize.net active so the trial section renders, the raw admin response outputs:

   ```html
   <table class="form-table">
       <tr class="recurring_info">
       ...
       </tbody>
   </table>
   ```

   Browsers insert the missing `<tbody>` in the DOM, but the source markup still has a closing `</tbody>` without a matching opening tag.

I like the overall direction of gating custom recurring trials behind gateway support. The main thing I’d change is making the hidden-field/save behavior explicit so hiding unsupported settings does not break normal saves or discount-code array indexing.

ideadude:
RE the bigger question, I do agree with this direction as an incremental step toward our future trial setting updates. Nice work.

PR #3629, opened 2026-04-01
Fixes the _load_textdomain_just_in_time was called incorrectly error

JarrydLong:
### All Submissions:

Fixes the notice related to loading our textdomain too early

```
PHP Notice: Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the paid-memberships-pro domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the init action or later.
```

1. Switch to a different language
2. Check error logs, nothing should come up related to this

### Other information:

### Changelog entry

### DIFF
--- a/includes/localization.php
+++ b/includes/localization.php
@@ -20,7 +20,7 @@ function pmpro_load_textdomain() {
 	//load via plugin_textdomain/glotpress
 	load_plugin_textdomain( 'paid-memberships-pro', false, dirname( __DIR__) . '/languages/' );
 }
-add_action( 'init', 'pmpro_load_textdomain', 1 );
+add_action( 'plugins_loaded', 'pmpro_load_textdomain', 1 );
 
 function pmpro_translate_billing_period($period, $number = 1)
 {	


flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3629 — Fixes the _load_textdomain_just_in_time was called incorrectly error**
*JarrydLong → dev | 1 file, +1 -1 lines*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3629>

**Summary**
Moves textdomain loading from `init` (priority 1) to `plugins_loaded` (priority 1) to prevent the WP 6.7+ notice that fires when something in the plugin triggers a translation call before the `init` hook runs. Clean, targeted fix. Ready to merge.

**Looks Good**
The `pmpro_load_textdomain()` function already calls `unload_textdomain()` before reloading, so it handles being invoked early without double-loading. Moving the hook earlier is safe given the function's design. Loading at `plugins_loaded` priority 1 also ensures the textdomain is available before the deprecated and compatibility shims that fire at priorities 10-20 on the same hook.

**Questions**
- What code was triggering the early translation call before `init`? Worth a quick note in the PR or a code comment, since moving the hook is a workaround — if the root cause is a class instantiation or a `plugins_loaded` callback using `__()` somewhere, that's worth knowing for future reference.

dparker1005:
After review, this change moves `pmpro_load_textdomain()` from `init` to `plugins_loaded`, which runs earlier in the WP load order. The notice itself states translations should be loaded at the init action or later, so this goes in the opposite direction of WP 6.7's guidance — it silences the notice only by pre-registering the textdomain before the offending early `__()` call, rather than fixing the early call itself.

The root cause is somewhere in PMPro, an add on, custom code, etc, a translation function (`__()`, `esc_html__()`, etc.) is being invoked before `init`. The proper fix is to identify that caller and defer it, similar to what #3200 did for the admin reports. Without a reliable local repro or stack trace from the reporter, we don't have enough to pin down the specific caller right now.

Closing for now. If this notice resurfaces with a reproduction case or error trace, we can open a fresh issue and track down the early caller.

PR #3630, opened 2026-04-02
Merge subscription delays and set expiration dates into core

dparker1005:
## Summary

Merges the functionality of the [Subscription Delays](https://www.paidmembershipspro.com/add-ons/subscription-delays/) and [Set Expiration Dates](https://www.paidmembershipspro.com/add-ons/pmpro-expiration-date/) Add Ons into PMPro core.

- **First Recurring Payment** setting on the Edit Level page (recurring levels): default (one billing cycle after checkout), after a number of days, or on a specific date via a date pattern builder (same day each month / same date each year / custom pattern like `Y-12-31`).
- **Expiration Type** setting: the existing duration-based expiration, or a fixed expiration date via the same pattern builder.
- Both settings are also available per level on the Edit Discount Code page.

### Storage and migration

Uses the **same `wp_options` keys as the retired add-ons** (`pmpro_subscription_delay_{level_id}`, `pmprosed_{level_id}`, `pmprosed_{level_id}_{code_id}`, `pmpro_discount_code_subscription_delays`), so existing sites migrate with no data changes. On update:

- Both add-ons are registered in `pmpro_get_deprecated_add_ons()`: auto-deactivated with a "delete these plugins" notice, and reactivation through the Add Ons page is blocked.
- During the one-request window where an add-on is still active, core unhooks its handlers (the legacy save handlers would otherwise wipe settings, and the text/checkout filters would double-apply).
- The add-ons' public API is preserved: their filters live on under their original names (`pmprosd_current_date`, `pmprosed_expiration_date_raw`, `pmprosed_expiration_date`), and their public functions (`pmprosd_convert_date()`, `pmprosd_getDelay()`, `pmpro_getDCSDs()`, `pmpro_saveDCSDs()`, `pmprosed_fixDate()`, `pmpro_getSetExpirationDate()`, `pmpro_saveSetExpirationDate()`) are defined as deprecated wrappers when the add-ons aren't loaded.

### Implementation notes

- Date pattern utilities live in `includes/date-patterns.php` (`pmpro_resolve_date_pattern()`, `pmpro_resolve_expiration_date_pattern()`, getters, the builder renderer, and request assembly). Everything else is integrated inline where core already works:
  - `pmpro_getLevelAtCheckout()` sets `profile_start_date` on the checkout level for a subscription delay (the mechanism gateways already consume), and applies a set expiration date to the level as `expiration_number`/`expiration_period` in days, exactly like the retired add-on did. The level object stays the source of truth for the end date, so `pmpro_checkout_level` callbacks can see and change it and `pmpro_isLevelExpiring()` is accurate at checkout. The PayPal Standard IPN and 2Checkout INS handlers do the same for their own level objects.
  - `pmpro_checkout_level_extend_memberships()` skips levels with a set expiration date, since a set date replaces duration-based expiration (this preserves the old add-on's ordering, where its callback ran after the extension).
  - `pmpro_getLevelCost()` / `pmpro_getLevelExpiration()` include the wording ("Recurring payments will begin on…", "Membership expires on…"). Expiration text is derived from the level object; the configured set expiration date is only consulted for level objects with no expiration of their own (e.g. the levels page), so a plugin that changes the expiration at checkout gets matching text.
  - Level and discount code save handlers assemble the pattern from the builder's submitted fields server-side.
- **No custom JavaScript or CSS**: all show/hide behavior uses the v3.9 declarative `data-pmpro-depends` engine, and the builder submits plain named fields.
- The date engine fixes several boundary bugs the add-ons had: "the 31st of each month" resolving to today (or a year ahead) at short month ends, patterns landing on "today" jumping a full year, and malformed patterns producing invalid dates. Unparseable patterns save as entered and are ignored at read time.

### Behavior notes for release

- Monthly patterns checked out on the pattern day itself now resolve to next month (previously jumped a full year); `Y-M-31` style patterns clamp to each month's last day.
- **Wording change:** day-based expirations longer than 31 days now render as a date ("Membership expires on December 1, 2026.") instead of "Membership expires after 90 Days." Shorter day counts and all other periods are unchanged.
- **Auto-Renewal Checkbox**: no update needed for launch. The expected setup for ARC levels is recurring billing without a set expiration date; opt-out members expire after one billing cycle as before. ARC's `remove_filter` of the old add-on's callback is now a harmless no-op, so only the unsupported combination (ARC enabled *and* a set expiration date on the same level) behaves differently: opt-in members would receive the set date. A small safety-net change for that case is drafted and can ship later if needed.
- Companion updates recommended: **PayFast** and **Paystack** webhook handlers should apply set expiration dates to their level objects like core's IPN handler now does. Two snippets-library recipes that `remove_filter`'d the old add-on hooks need rewrites against `pmpro_checkout_level`.

## Proposed changelog entries

```
= TBD =
* FEATURE: Merged the Subscription Delays Add On into core. Levels and discount codes can delay the first recurring payment by a number of days or until a specific date.
* FEATURE: Merged the Set Expiration Dates Add On into core. Levels and discount codes can expire memberships on a specific date using date patterns.
* ENHANCEMENT: The Subscription Delays and Set Expiration Dates Add Ons are automatically deactivated when active; their settings, filters, and public functions are preserved.
* BUG FIX: Date patterns such as "the 31st of each month" now resolve correctly at the end of shorter months, and patterns matching the current day advance to the next occurrence instead of the next year.
* ENHANCEMENT: Membership expiration text now shows the expiration date instead of a day count when the expiration is more than 31 days away.
```

## Test plan

- Date engine verified with a boundary-case harness (month ends, leap years, today-collisions, token limits, malformed input) plus an exhaustive sweep of every monthly/yearly day value across 2026–2029.
- Browser-verified: Edit Level and Edit Discount Code round-trips for all modes, depends-driven visibility, save-as-typed for unparseable patterns, add-on-era stored values rendering correctly.
- Real checkout verified end to end (check gateway): delayed first payment sets `profile_start_date`; a fixed expiration date lands as the level's end date at `23:59:59`.
- Checkout level verified via wp-cli for the final approach: set date applied as N Days, past fixed dates floor to 1 day, invalid/absent patterns leave the level untouched, renewing members with a set date are not extended while plain duration levels still are, and expiration text matches the level object for changed/zeroed expirations.
- Migration verified live: activating the old add-ons alongside this branch (no fatals, no doubled text, no save wipes), auto-deactivation with notice, legacy function/filter API intact.


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3630 — Merge subscription delays and set expiration dates into core**
*dparker1005 → v3.9 | 10 files, +876 -86*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3630>

---

**Summary**
Solid core merge with a well-built date-pattern engine and a genuinely careful migration/neutralization strategy for the two retired add-ons. Not ready to merge: the expiration-date fix was applied to two of three gateway completion paths and missed the third, and a shared helper wasn't updated so one admin screen still misreports fixed-date expirations. Neither is a fundamental design problem — both are narrow, fixable gaps.

---

**Issues**

- **Major** `services/twocheckout-ins.php:310-318` — `pmpro_insChangeMembershipLevel()` still computes `$enddate` from `expiration_number`/`expiration_period` only (confirmed unchanged from upstream `v3.9`, and not among the 10 files this PR touches). The parallel fix landed in `includes/checkout.php:238` and `services/ipnhandler.php:555` but not here. A level or discount code configured with a fixed "on a specific date" expiration and no duration set will fall through to `$enddate = "NULL"` on a 2Checkout IPN, silently granting indefinite access. 2Checkout is deprecated (`includes/deprecated.php:1029`) but its `wp_ajax_twocheckout-ins` handler is still registered (`includes/services.php:78-79`) for existing sites on that gateway. Mirror the `pmpro_get_set_expiration_date()` / `pmpro_resolve_expiration_date_pattern()` block into this handler.

- **Major** `includes/functions.php:273-283` — `pmpro_isLevelExpiring()` (unchanged from `v3.9`) still only checks `expiration_number`/`enddate` and has no awareness of `pmpro_get_set_expiration_date()`. The PR works around this locally in `edit-level.php` and `discountcodes.php:648` with inline `|| $expiration_date_type === 'date'` checks, but `adminpages/membershiplevels.php:496` (untouched by this PR) calls `pmpro_isLevelExpiring($level)` directly on the level-definition object and will show `—` in the Expiration column for a level that actually has a fixed expiration date configured. Fix the shared function so every caller benefits instead of patching call sites individually.

- **Major** `includes/functions.php:2969` / `adminpages/levels/save-level.php:168` / `adminpages/discountcodes.php:283` — The numeric (days) subscription-delay path has no floor and no server-side minimum. `save-level.php` stores `intval( $_REQUEST['subscription_delay_days'] )` with no `max(1, ...)` (the UI's `min="1"` isn't enforced server-side), and `pmpro_getLevelAtCheckout()`'s numeric branch sets `profile_start_date = date('Y-m-d', strtotime('+ ' . intval($subscription_delay) . ' Days', ...))` with no clamp against today — unlike the date-pattern branch two lines below it, which explicitly does `max( $delay_date, date('Y-m-d', current_time('timestamp')) )`. A negative or zero stored value sends a past `profile_start_date` to the gateway. Apply the same floor to the numeric branch and clamp the days input to `max(1, intval(...))` in both save paths, matching the pattern already used in `pmpro_get_date_pattern_from_request()`.

- **Minor** `includes/deprecated.php` — The retired Set Expiration Dates add-on shipped a dashboard warning (`pmprosed_show_admin_notice_past_dates()`, checked via the add-on source) that flagged any level whose resolved expiration date had already passed. That mechanism isn't carried over anywhere in this PR, so admins lose visibility into misconfigured or migrated past-dated expirations. Worth a follow-up (not a blocker — checkout completing with a past literal date isn't a new regression, the add-on never blocked it either, it just warned).

- **Minor** `includes/date-patterns.php:255` / `:279` — `pmpro_get_subscription_delay()`'s discount-code branch indexes `$all_delays[ $code_id ][ $level_id ]` with raw (uncast) values, while `pmpro_get_set_expiration_date()`'s equivalent branch casts both to `intval()`. Low practical risk since both IDs come from internal DB rows, but worth aligning for consistency.

---

**Looks Good**
`includes/date-patterns.php`'s month/year rollover math checks out by hand — today-collisions advance correctly, day values clamp to month length at boundaries (Feb 31 → Feb 28/29). The two-phase add-on neutralization is well-targeted: `plugins_loaded` for most legacy hooks, `init` priority 11 separately for `pmprosd_pmpro_checkout_level`, which the old add-on itself registered at `init`. `pmpro_define_deprecated_date_pattern_functions()`'s activation guards (`pmpro_activating_plugin()`, the `activate-plugin`/`pmpro_addon_activate` actions, WP-CLI) correctly cover the "cannot redeclare" fatal risk during an add-on activation request. Reusing the add-ons' `wp_options` keys means zero-data-migration for existing sites, and routing the delay through `profile_start_date` in `pmpro_getLevelAtCheckout()` (a mechanism every gateway already consumes) is the right upstream fix rather than patching each gateway separately.

---

**Questions**

1. The PR description says "the membership levels list warns about" unparseable patterns — `adminpages/membershiplevels.php` isn't in this diff and has no such warning currently. Is that a separate follow-up PR, or should it be added here?
2. Any plan to restore the past-expiration-date admin notice the old add-on had, or is that intentionally being dropped?

flintfromthebasement (commented):
**Follow-up review**

**PR: strangerstudios/paid-memberships-pro#3630 — Merge subscription delays and set expiration dates into core**
*dparker1005 → v3.9 | 10 files, +876 -86*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3630>

**Summary**
Both blocking Majors from my last pass are resolved, and the fix is architecturally better than what I suggested: instead of mirroring the expiration-date resolution block into each gateway completion path, `pmpro_getLevelAtCheckout()` and both IPN handlers now convert a resolved fixed date into `expiration_number`/`expiration_period = 'Day'` on the level object itself, so the existing enddate math handles it uniformly. Ready to merge from my side, modulo the two pre-existing minors below.

**Prior Issues — Status**

- **Major** `services/twocheckout-ins.php:310-318` (missing fix on the 2Checkout IPN path) — **Resolved.** Verified against `pr-3630-review` head (`9bb92fe8f`): the handler now computes `$set_expiration_date`/`$resolved_expiration_date` and converts to `expiration_number`/`'Day'` before the existing enddate calculation, mirroring `services/ipnhandler.php`.

- **Major** `includes/functions.php:273-283` (`pmpro_isLevelExpiring()` unaware of set-expiration-dates, causing `adminpages/membershiplevels.php:496` to misreport) — **Resolved (symptom), not addressed (root ask).** `membershiplevels.php:493-503` now resolves the set expiration date inline and prints "On &lt;date&gt;" before falling back to `pmpro_isLevelExpiring()`, matching the same workaround already present in `edit-level.php` and `discountcodes.php`. The reported bug is gone — all four callers of `pmpro_isLevelExpiring()` now have call-site handling. I still think fixing the shared function once would be more durable than four independent inline duplicates, but that's a design preference, not a blocker. Not holding this up further.

- **Major** `includes/functions.php:2969` / `save-level.php:168` / `discountcodes.php:283` (no floor on numeric subscription delay) — **Resolved.** Both admin save paths now clamp with `max(1, intval(...))`, and `pmpro_getLevelAtCheckout()`'s numeric branch now floors `profile_start_date` against today with `max($delay_date, date('Y-m-d', current_time('timestamp')))`, matching the pattern the date-based branch already used.

- **Minor** `includes/deprecated.php` (no admin notice for past set-expiration-dates) — **Not addressed.** Still no equivalent to the old add-on's `pmprosed_show_admin_notice_past_dates()`. Non-blocking, as before — worth a follow-up ticket.

- **Minor** `includes/date-patterns.php:255/279` (uncast array index vs. `intval()` elsewhere) — **Not addressed**, unchanged since last review. Still low-risk, still worth a cleanup pass.

**New Findings (delta since d5e70e6e)**

None. Reviewed the day-count conversion math added to `pmpro_getLevelAtCheckout()`, `ipnhandler.php`, and `twocheckout-ins.php` (`DAY_IN_SECONDS` round-trip, `max(1, ...)` floor on same-day/past dates) and the new `pmpro_getLevelExpiration()` display branch for >31-day counts — traced all four callers of that function (`pages/levels.php`, `pages/checkout.php`, `includes/scripts.php`, `services/applydiscountcode.php`) and confirmed they're all pre-checkout/pricing-page contexts where "today" is the correct anchor, not a stale enddate-display bug.

**Looks Good**
Converting the resolved date to `expiration_number`/`'Day'` once, upstream of all three completion paths, instead of duplicating the resolve-and-format block three times, is a cleaner fix than the one I proposed — single source of truth for "how long until this level expires," reused by every downstream consumer of the level object.

dparker1005:
Hi @flintfromthebasement, thank you for the follow-up and for confirming the three Majors.

Fixed across b6d2271bd, 69020fa1f, 75cbd5548 and 9bb92fe8f: the 2Checkout INS path applies set expiration dates, the levels list shows fixed dates in the Expiration column, the numeric delay is floored on save and clamped at checkout, and a resolved fixed date is now converted to `expiration_number`/`Day` once on the level object so every completion path shares the same enddate math.

Not changing in this PR:
- `pmpro_isLevelExpiring()`: the set date isn't a property of the level object, it lives in an option and whether it applies depends on the discount code in play. The call sites know that context; the shared function doesn't, and teaching it to read options would give wrong answers for code-overridden levels. So the resolution stays at the call sites.
- Past-date admin notice: dropped on purpose to keep this PR at add-on parity and minimum diff. Checkout on a past fixed date just expires the member on the next cron run, which is the same behavior the add-on had. Happy to revisit as a follow-up.
- Uncast array indexes in `pmpro_get_subscription_delay()`: both IDs come from internal rows, so leaving it.

includes/deprecated.php
```diff
@@ -1163,3 +1171,177 @@ function pmpro_was_loading_frontend_css_notice() {
...
+		remove_action( 'pmpro_discount_code_after_level_settings', 'pmprosed_pmpro_discount_code_after_level_settings' );
+		remove_action( 'pmpro_save_discount_code_level', 'pmprosed_pmpro_save_discount_code_level' );
+		remove_filter( 'pmpro_checkout_level', 'pmprosed_pmpro_checkout_level' );
+		remove_filter( 'pmpro_discount_code_level', 'pmprosed_pmpro_checkout_level' );
+		remove_filter( 'pmpro_ipnhandler_level', 'pmprosed_pmpro_ipnhandler_level' );
+		remove_filter( 'pmpro_payfast_itnhandler_level', 'pmprosed_pmpro_ipnhandler_level' );
+		remove_filter( 'pmpro_paystack_webhook_level', 'pmprosed_pmpro_ipnhandler_level' );
+		remove_filter( 'pmpro_checkout_end_date', 'pmprosed_force_set_expiration_enddate' );
+		remove_filter( 'pmpro_level_expiration_text', 'pmprosed_pmpro_level_expiration_text' );
+	}
```
kimcoleman, line 1332: Should we also unhook the column added to the Memberships > Settings > Levels screen here?

```
remove_action( 'pmpro_membership_levels_table_extra_cols_header', 'pmprosed_membership_levels_table_extra_cols_header', 15 );
remove_action( 'pmpro_membership_levels_table_extra_cols_body', 'pmprosed_membership_levels_table_extra_cols_body', 15 );
```

PR #3637, opened 2026-04-13
Fix export download access failures in Action Scheduler export system

dalemugford:
### All Submissions:

Users reported that export CSV downloads return "You do not have permission to access this file" (403) or "File not found" (404) even when the export appears to have completed. Three bugs were identified and fixed:

**Bug 1 — Export record deleted before file delivery is confirmed**

`cleanup_after_download()` was called inside the `pmpro_can_access_restricted_file` filter, which runs *before* `readfile()` in `pmpro_restricted_files_check_request()`. If the file was missing (or any error occurred after the filter), the export record was already gone — leaving the user with a 404 and no way to retry without starting over. Deferred cleanup to a new `pmpro_restricted_file_served` action that fires immediately after `readfile()` succeeds.

**Bug 2 — Token transient expiry breaks download after 24 hours**

The plaintext download token was stored in a transient with a hardcoded 1-day TTL. After expiry, `get_download_url()` silently built a URL with an empty `token=` parameter, causing a 403 for any export older than 24 hours even though the record and file still existed. Extended both the token and owner transient TTLs to 7 days. Also hardened `get_download_url()` to return `null` when no token is available, and updated `format_public_export_response()` to omit `download_url` in that case — so the UI falls back to idle/retry instead of presenting a broken link.

**Bug 3 — Zero-record exports create a ghost "complete" state with no file**

When `total_count = 0`, the export was immediately marked `complete` without ever opening or writing a CSV file. The download attempt would pass access validation, run cleanup (deleting the record), then fail with a 404. Added an early return in `start_export()` when the total count is 0, returning a descriptive error instead.

1. Export members (< 499) — confirm the CSV downloads and the button resets to idle.
2. Export members (> 499, async) — wait for background completion, confirm download works.
3. Filter members to a level with no members and attempt export — confirm an inline error message appears immediately with no ghost export record created.
4. After a successful export, manually delete the token transient (`delete_transient('pmpro_export_token_{id}')`) and poll the status endpoint — confirm `download_url` is absent from the response and the button shows idle/retry state.
5. Simulate a failed file serve (rename the export file after access is granted) — confirm the export record still exists after the 404, allowing a retry.

### Other information:

### Changelog entry

> Fix export CSV download failures caused by premature record cleanup, token transient expiry after 24 hours, and zero-record exports creating an undownloadable "complete" state.

### DIFF
--- a/classes/class-pmpro-exports.php
+++ b/classes/class-pmpro-exports.php
@@ -155,7 +155,7 @@ class PMPro_Exports {
 	 * @return void
 	 */
 	public function handle_restricted_file_served( $file_dir, $file ) {
-		if ( 'exports' !== $file_dir || empty( $this->pending_post_serve ) ) {
+		if ( 'exports' !== $file_dir || empty( $this->pending_post_serve ) || $file !== $this->pending_post_serve['file'] ) {
 			return;
 		}
 		$ctx                    = $this->pending_post_serve;
--- a/classes/class-pmpro-exports.php
+++ b/classes/class-pmpro-exports.php
@@ -40,6 +40,14 @@ class PMPro_Exports {
 	protected $default_export_exp = 6 * HOUR_IN_SECONDS;
 
 	/**
+	 * Pending post-serve cleanup context.
+	 * Populated by export_can_access_restricted_files() and consumed by handle_restricted_file_served().
+	 *
+	 * @var array|null
+	 */
+	protected $pending_post_serve = null;
+
+	/**
 	 * Get singleton instance.
 	 *
 	 * @return PMPro_Exports
@@ -61,6 +69,8 @@ class PMPro_Exports {
 		add_action( 'pmpro_export_delete_file', array( $this, 'delete_file_task' ), 10, 1 );
 		// File access filter.
 		add_filter( 'pmpro_can_access_restricted_file', array( $this, 'export_can_access_restricted_files' ), 10, 2 );
+		// Post-serve cleanup (fires after readfile() confirms the file was delivered).
+		add_action( 'pmpro_restricted_file_served', array( $this, 'handle_restricted_file_served' ), 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3637 — Fix export download access failures in Action Scheduler export system**
*dalemugford → dev | 2 files, +56 -8*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3637>

---

**Summary**

Solid bug fix PR. All three stated problems are real, the root causes are correctly diagnosed, and the fixes are architecturally clean. The deferred-cleanup approach (storing context in `$pending_post_serve` and consuming it in the `pmpro_restricted_file_served` action) is the right call. Ready to merge with two minor caveats noted below.

---

**Issues**

- **Minor** `includes/restricted-files.php:97` — `readfile()` return value is not checked before firing the cleanup action. The PR description says the action "fires immediately after `readfile()` succeeds," but the implementation unconditionally fires it:
  ```php
  readfile( $file_path );
  do_action( 'pmpro_restricted_file_served', $file_dir, $file );
  ```
  If `readfile()` returns `false` (disk error, permission issue after the `file_exists` check passed), the export record is still cleaned up even though nothing was delivered. Fix: `$bytes = readfile( $file_path ); if ( false !== $bytes ) { do_action( ... ); }` — or at minimum, note in a comment that the false-return case is an acceptable tradeoff.

- **Minor** `classes/class-pmpro-exports.php:157` — `handle_restricted_file_served()` checks `$file_dir === 'exports'` but never verifies that `$file` matches `$this->pending_post_serve['file']`. In practice this can't misfire (single request, `exit` after serve), but the guard is incomplete relative to what it's protecting:
  ```php
  if ( 'exports' !== $file_dir || empty( $this->pending_post_serve ) ) {
  ```
  Could add `|| $file !== $this->pending_post_serve['file']` to make the guard airtight.

---

**Looks Good**

- `get_download_url()` null-guard + caller update at `class-pmpro-exports.php:686-691` is clean. The old behavior (silently building a URL with `token=`) was a silent failure; this makes it explicit and lets the UI handle the missing-token case gracefully.
- Zero-record bail at `class-pmpro-exports.php:486-488` correctly exits before `create_export_record()` — no ghost record created. The `=== 0` strict check is appropriate since `count_total_for_type()` always casts to `int`.
- TTL bump from `DAY_IN_SECONDS` to `7 * DAY_IN_SECONDS` for both the token and owner transients is consistent and addresses the stated 24-hour expiry bug without introducing any new state concerns.

dalemugford:
Thanks for the review! Both issues have been addressed and pushed:

- **`readfile()` return value** — now captured and the action only fires when it returns a byte count rather than `false`, so a disk error after the `file_exists` check won't trigger a spurious cleanup.
- **File name guard** — added `$file !== $this->pending_post_serve['file']` to the `handle_restricted_file_served()` condition to make the guard airtight.

PR #3649, opened 2026-04-21
Prevent concurrent Stripe webhook deliveries from racing on the same order

dparker1005:
## Summary

When Stripe delivers multiple webhook events for the same checkout within a short window (normal "at least once" delivery from different Stripe infrastructure nodes), PHP-FPM workers can process them concurrently. Both threads pass the `status IN ('token','pending')` dedup check and enter `pmpro_changeMembershipLevel()`, which has no locking around its cancel -> insert -> dedup-cleanup sequence, producing:

- **One-time payment:** duplicate membership rows, both set to `changed` by each thread's own "remove duplicates" cleanup, leaving zero `active` rows. User paid but has no membership.
- **Recurring subscription:** the second thread cancels the Stripe subscription the first thread just created, triggering `customer.subscription.deleted` and marking the membership `cancelled` seconds after creation.

This PR adds a MySQL advisory lock (`GET_LOCK`/`RELEASE_LOCK`) scoped per-order in `services/class-pmpro-stripe-webhook-handler.php`. Concurrent webhooks that cannot acquire the lock bail out immediately. Advisory locks are truly atomic (no TOCTOU window), auto-release when the MySQL connection closes (intrinsic worker-death recovery, no TTL needed), and are available on virtually all MySQL/MariaDB hosts across both InnoDB and MyISAM since they are a server-level feature independent of storage engine.

On the rare host where `GET_LOCK` returns `NULL` (unavailable or errored), the webhook handler logs a warning and proceeds without concurrency protection rather than breaking webhook handling entirely. Those hosts end up in the same state they were in before this PR.

Also refactors the duplicated processing logic at the two checkout-session completion handler call sites into a single private method, `process_checkout_session_completion` (renamed from `change_membership_level`, which was a thin 2-line wrapper only called from these two sites).

## Test plan

- [ ] Normal one-time Stripe Checkout: single `active` membership row, no `changed` duplicates with `enddate = startdate`
- [ ] Normal recurring Stripe Checkout: single `active` row, subscription stays `active`
- [ ] Simulated concurrent delivery: capture a real webhook body and `Stripe-Signature` header, fire two simultaneous `curl -X POST` requests; only one processes, the second logs "already being processed by a concurrent webhook"
- [ ] Simulated `pmpro_complete_async_checkout` failure: order marked `error`, advisory lock still released via connection close
- [ ] Worker-death recovery: force a fatal error mid-processing; advisory lock auto-releases when the PHP worker's MySQL connection closes, next webhook proceeds normally
- [ ] Fast-path regression: manually set order status to `success`; second webhook exits at status check without attempting the lock
- [ ] Verify on a test host where `GET_LOCK` is unavailable: warning logged, processing continues (back to pre-PR behavior for that minority of hosts)


flintfromthebasement:
**PR: strangerstudios/paid-memberships-pro#3649**
<https://github.com/strangerstudios/paid-memberships-pro/pull/3649>

**Summary**
Not ready as-is. At least one reviewer found a critical/major issue that should be checked before merge.

**Synthesized Issues**

- **[Major]** issue below.
  _Raised by: pi._
- **[Major]** `class-pmpro-stripe-webhook-handler.php` — Lock name `pmpro_stripe_order_` . `$morder->id` is not namespaced to the site. On shared MySQL instances (common in budget hosting), `GET_LOCK` names are global to the MySQL server, not the database. Order ID 123 on site A and order ID 123 on site B will share the same advisory lock. This can cause a webhook for site B to be silently skipped because site A holds the lock on the same name. Fix: include the database name or a site-specific prefix in the lock name, e.g. `$wpdb->dbname . '.pmpro_stripe_order_' . $morder->id`, or hash `DB_NAME` into the lock string. MySQL advisory lock names support up to 64 characters, so there is room.
  _Raised by: pi._
- **[Minor]** issues worth addressing before it ships.
  _Raised by: sonnet._
- **[Minor]** `services/class-pmpro-stripe-webhook-handler.php` (new `try` block, post-lock re-read) — The comment directly above `$current_status = $wpdb->get_var(...)` reads: *"Re-read order status now that we hold the lock."* This comment is inaccurate when `$got_lock !== '1'` (i.e., the NULL fallback path where GET_LOCK errored or was disabled via filter). In that path the code correctly re-reads from DB — which is actually better than trusting the stale in-memory status — but the comment claims the lock is held when it may not be. Change to something like: *"Re-read order status from DB. The in-memory status may be stale if loaded before a concurrent worker persisted a terminal status."* The behavior is fine; the comment misleads.
  _Raised by: sonnet._
- **[Minor]** `services/class-pmpro-stripe-webhook-handler.php` (`try/finally` block) — The advisory lock is held for the full execution of `pmpro_complete_async_checkout()`, which internally calls `do_action('pmpro_after_checkout', ...)` and sends two emails via `PMProEmail::sendCheckoutEmail()` / `sendCheckoutAdminEmail()`. These are entirely outside the critical section — the race is resolved once `pmpro_changeMembershipLevel()` and `$order->saveOrder()` complete. Holding the lock through third-party hook code and email delivery means a slow `pmpro_after_checkout` hook (Sponsored Members, Group Memberships, external API calls) will block any re-delivered webhook for the full duration. Not a correctness issue since Stripe allows 30 seconds, but narrowing the scope post-`saveOrder()` would make the locking intent clearer and reduce unnecessary hold time. Requires either inlining or splitting `pmpro_complete_checkout()` — reasonable to defer to a follow-up if refactoring scope is a concern.
  _Raised by: sonnet._
- **[Minor]** `services/class-pmpro-stripe-webhook-handler.php` (method signature) — `process_checkout_session_completion( $morder, $checkout_session, &$logstr )` receives the full `$checkout_session` object but uses only `$checkout_session->id` for log strings. Accepting `string $checkout_session_id` directly would make the data dependency explicit and make the method more testable in isolation. Low priority, but it's a leaky abstraction.
  _Raised by: sonnet._
- **[Minor]** `class-pmpro-stripe-webhook-handler.php` — The `$got_lock = null` branch (filter disabled or `GET_LOCK` unavailable) proceeds without any concurrency protection but only logs a `WARNING` appended to `$logstr`. This warning string will only be visible if the site has Stripe webhook logging enabled. Consider also triggering a `trigger_error()` or `error_log()` call so the warning surfaces in the PHP error log regardless of Stripe debug settings, making it discoverable during incident investigation.
  _Raised by: pi._
- **[Minor]** `class-pmpro-stripe-webhook-handler.php` — The post-lock status re-check queries `$wpdb->pmpro_membership_orders` directly, but the pre-lock fast-path check reads `$morder->status` from the in-memory object. These use different data sources (raw column vs. `MemberOrder` property which may have been set at object construction time and could be stale by the time the webhook runs). This is fine — the post-lock check is the authoritative one — but worth a brief comment on the pre-lock check noting it's an optimization, not a correctness guard. The PR description mentions this logic but the code comment could be clearer.
  _Raised by: pi._

**Reviewer Notes**

- **Claude Sonnet (senior-review)**: *dparker1005 → dev | 1 file, +98 -41 lines* <https://github.com/strangerstudios/paid-memberships-pro/pull/3649> --- **Summary** This PR is a correct, well-reasoned fix for a real and painful data integrity bug. The two-phase check (pre-lock fast-path + post-lock DB re-read), the `'0'`/`'1'`/`null` lock result handling, and the `finally`-based release are all correctly implemented. It's close to mergeable. Three minor issues worth addressing before it ships.
- **Codex GPT (build-release-verifier)**: (no output)
- **pi GLM (independent-edge-case-review)**: *dparker1005 → dev | 1 file, +98 -41 lines* <https://github.com/strangerstudios/paid-memberships-pro/pull/3649> --- **Summary** Solid, well-scoped fix for a real production bug. The advisory-lock approach is correct — atomic, auto-releases on connection close, graceful degradation on hosts where `GET_LOCK` is unavailable. The refactor consolidating duplicated processing logic into `process_checkout_session_completion` is clean. Ready to merge after addressing the one major issue below.

**Metrics**

- Claude Sonnet: 178s, 559 words, ~$0.3672
- Codex GPT: 0s, 0 words
- pi GLM: 72s, 536 words, ~$0.0706

PR #3656, opened 2026-04-26
Reworking security and email settings admin areas for detecting sending client, Max services active, removing SendWP built-in

kimcoleman:
Reworks the Email Settings and Security Settings admin areas to detect what's actually handling email/security on the site, surface that on the page (and in Site Health), give Max customers context-aware messaging, and remove the legacy SendWP built-in. 

### Email Settings (adminpages/emailsettings.php)

- Replaced the SendWP "Connect / Disconnect" UI with a Sending Method display driven by the new `pmpro_detect_email_method()` helper
- Per-source description copy (plugin / constant / hosting / default) so the user understands what we detected and what it does (or doesn't) tell them
- Hosting-specific intro paragraph when source === 'hosting' (PMPro Max) confirming transactional email is included 

### Sending-method detection (includes/email.php) 

- New `pmpro_detect_email_method()` checking, in order: known SMTP/transactional plugins > SMTP_HOST wp-config constant > PMPro Max (PMPRO_HOSTING_VERSION and environment) > unknown/default fallback
- Plugin signatures: WP Mail SMTP, FluentSMTP, Post SMTP, Easy WP SMTP, Gravity SMTP, WP Offload SES, Gmail SMTP, Postmark, Brevo, Mailgun, SendWP
- Signature schema accepts string-or-array for class / function and adds a constant check, so plugins that span major versions can be matched without duplicate entries 
- Catchall labeled Unknown / WordPress default (PHP mail) rather than confidently claiming PHP mail when we can't actually tell 

### Security Settings (adminpages/securitysettings.php) 

- Section renamed WordPress Security Plugins > WordPress Security 
- PMPro Max is now recognized as a security provider (via `getenv( 'PMPRO_HOSTING' )`) and listed alongside detected security plugins 
- Refactored the hard-coded MalCare / Wordfence / Solid Security blocks into a single loop over a `$security_plugins_to_check` map
- "No security detected" copy now points to PMPro Max documentation instead of pushing a MalCare install
- When Max is active alongside another security plugin, each row gets a per-plugin recommendation to deactivate (since Max already covers it) 
- Style: replaced notice notice-warning notice-large inline with `pmpro_message pmpro_alert` for the "Multiple Security Plugins Active" warning 
- Added UTM params to outbound documentation/marketing links

### Site Health (classes/class-pmpro-site-health.php) 

- New Email Sending Method row inside the existing PMPro Site Health Info section (`get_email_method_info()`), exposing detected method, source, and SMTP relay (when known). Useful for support tickets — it travels in the standard Site Health export.

### SendWP cleanup

- Deleted `includes/lib/SendWP/sendwp.php` and `includes/lib/SendWP/installer.js` 
- Removed the `require_once` of SendWP from `includes/admin.php`
- Removed the connect/disconnect UI from `emailsettings.php`. Anyone currently connected has the SendWP plugin installed and active and can manage it from their settings page.

The Email Deliverability section was previously SendWP-centric and didn't tell users what was actually sending their mail. Likewise the Security section pushed MalCare regardless of what was already in place. Both now reflect reality and surface PMPro Max where it's relevant. The Site Health row makes the email-method diagnostic available in support exports.


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3656 — Reworking security and email settings admin areas**
*kimcoleman → dev | 7 files, +323 -344*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3656>

---

**Summary**

Solid cleanup. SendWP removal is clean (both registration and enqueue hooks are gone), the `pmpro_detect_email_method()` abstraction is well-structured with a filter on every return path, and refactoring the security plugin checks into a data-driven loop is a clear improvement. One inconsistency in how Max is detected across files and a few phpcs hygiene gaps — neither is a blocker, but the env var inconsistency is worth fixing before merge.

---

**Issues**

**Major** `adminpages/securitysettings.php:194` vs `includes/email.php:538` — `PMPRO_HOSTING` is checked two different ways:

```php
// securitysettings.php — loose truthy
$pmpro_max_status = getenv( 'PMPRO_HOSTING' ) ? 'active' : 'not-installed';

// email.php — strict string match
getenv( 'PMPRO_HOSTING' ) === '1'
```

Any truthy non-`'1'` value (e.g. `'true'`, `'yes'`) marks Max as active in security settings but falls through to "default" in email detection. The security page and email section would disagree about whether Max is active. Pick one form and use it everywhere — `=== '1'` is the safer choice.

**Minor** `adminpages/securitysettings.php:317` — loose `!=` for a string comparison:

```php
$plugin[0] != 'pmpro-hosting/pmpro-hosting.php'
```

Use `!==` to match the strict comparison style used throughout the rest of the file.

**Minor** `adminpages/emailsettings.php:97-99` and `~112-115` — two `printf()` calls embedding HTML are missing `// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped`, inconsistent with the identical pattern used at lines 63, 69, and 77 in the same file. The output is safe but PHPCS will flag these inconsistently if the project runs the escape sniff.

**Minor** `adminpages/securitysettings.php:196, 213` — `$plugin[2]` (the status string) is set in `$installed_security_plugins` entries but the render loop now hardcodes `pmpro_tag-active` instead of reading it. Either drop `$plugin[2]` from the array construction or read it in the render loop — dead array slots are confusing.

---

**Looks Good**

`pmpro_detect_email_method()` in `includes/email.php:430-555` — clean layered detection with `apply_filters()` on every return path, good plugin signature schema that supports string-or-array for class/function/constant, and an honest "unknown/default" catchall instead of confidently claiming PHP mail. The docblock accurately describes the best-effort limitations.

Removing the standalone `$active_security_plugins_count` counter and replacing it with `count( $installed_security_plugins )` is a good simplification — the count was always equivalent.

---

**Questions**

1. Is `PMPro_Hosting_Transactional_Emails::get_instance()` safe to call at the point `pmpro_detect_email_method()` runs (i.e., is that class guaranteed to be initialized by then, or could it run before its plugin has set up state)?

kimcoleman:
Added a commit to fix the getenv check for `PMPRO_HOSTING` to always use strict === '1'.

Removed the dead `$plugin[2]` slot in the active plugins check - not used, not needed anymore. That was there from the SendWP check we needed to see if it was "installed" but "not connected". Since we no longer show these connection systems in our screens, that third array value wasn't needed.

Also updated the one check to be a strict conditional for consistency.
                                                                                                                 
As far as the potential question raised about the `PMPro_Hosting_Transactional_Emails::get_instance()` usage, it is safe here. `pmpro_detect_email_method()` is only called from `emailsettings.php` at admin page render time, well after `plugins_loaded`, and is gated by `class_exists()`.

Issue #3659, opened 2026-04-27
Deleting a user with membership levels first causes warning on pmpro function called incorrectly.


LSNicholls:
**Describe the bug**
 If you use the wp-admin users page to delete a user and opt to cascade the deletion to PMPro-specific data, a PHP notice says  pmpro_changeMembershipLevel is being called when pmpro_cancelMembershipLevel should be called.  It still seems to work.

**To Reproduce**
1. Turn on debug log.
2. From wp-admin/users.php), choose a user to delete who has an active membership (it might happen without one, but I didn't test that).
3. Click on Delete
4. You should see a screen similar to the screenshot below.  
5. Click the checkboxes, as shown, and press the button to **Confirm Deletion**
6. You are returned to the Users screen, with your selected individual deleted.  However there is a notice in the log:

```
Function pmpro_changeMembershipLevel was called <strong>incorrectly</strong>.  
The pmpro_cancelMembershipLevel() function should be used to cancel membership levels.  
```
**Screenshots**

Please note that, although the illustrative member in the screenshot was uploaded via csv,  that fact isn't a relevant part of the scenario.
<img width="656" height="419" alt="Image" src="https://github.com/user-attachments/assets/d130862b-2c6c-4245-b0df-6b90f6d93150" />

**Expected behavior**
pmpro_delete_user function needs an update?

**Additional observation**
I'm pretty sure the information was deleted from the pmpro_memberships_users table as expected.  However, queries of the pmpro_membership_orders table indicate orphans.   A note in the code (` we don't remove the orders because it would affect reporting`) indicates that this behavior is deliberate.  However it would be a good idea to make it possible ( have a third checkbox to "also delete orders"), for at least two reasons:  (a) being able to scrub test data thoroughly when needed (b) being able to do an extra-thorough job in a "right to be forgotten" scenario.  I know that the orders table doesn't contain PII, but I'm not sure everybody would see that point as persuasive.

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only Paid Memberships Pro plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Memberlite](https://memberlitetheme.com).
- [x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
This is something I noticed while putting a whole bunch of users in quickly, and then trying to delete them, for #3658, and occurred in the same cleanroom environment.
</details>

dparker1005:
[...]

PR #3660, opened 2026-04-27
Fix deprecated pmpro_changeMembershipLevel call when deleting a WP user    

kimwhite:
Reported here https://github.com/strangerstudios/paid-memberships-pro/issues/3659                                                                               
 
**Summary**                                                                                                                                                                                                                                                                                                                                                              
When an admin deletes a WordPress user with active membership levels, PMPro was calling pmpro_changeMembershipLevel( 0, $user_id ) to cancel their levels. Passing 0 as the level
was deprecated in PMPro 3.0 in favor of pmpro_cancelMembershipLevel(), which generated a PHP Notice on every user deletion:                            

> Function pmpro_changeMembershipLevel was called incorrectly. The pmpro_cancelMembershipLevel() function should be used to cancel membership levels.

**Changes**

  - includes/cleanup.php — Replaced the single pmpro_changeMembershipLevel( 0, $user_id ) call in pmpro_delete_user() with a loop over pmpro_getMembershipLevelsForUser() that
  calls pmpro_cancelMembershipLevel() for each level. This matches the behavior the deprecated code path was already falling back to internally.

  Testing

  1. Create a user with one or more active membership levels.
  2. Delete the user from WP Admin → Users with "Cancel active subscriptions" checked.
  3. Confirm the membership levels are cancelled and no PHP Notice is logged.         

WordPress 6.9.4
PMPro 3.7.1

### DIFF
--- a/includes/cleanup.php
+++ b/includes/cleanup.php
@@ -22,7 +22,9 @@ function pmpro_delete_user( $user_id ) {
 	 * @param int  $user_id                     The WordPress user ID.
 	 */
 	if ( apply_filters( 'pmpro_user_deletion_cancel_active_subscriptions', $cancel_active_subscriptions, $user_id ) ) {
-		pmpro_changeMembershipLevel( 0, $user_id );
+		foreach ( pmpro_getMembershipLevelsForUser( $user_id ) as $level ) {
+			pmpro_cancelMembershipLevel( $level->id, $user_id );
+		}
 	}
 
 	//Remove all membership history for this user from the pmpro_memberships_users table


flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3660 — Fix deprecated pmpro_changeMembershipLevel call when deleting a WP user**
*kimwhite → dev | 1 file, +3 -1*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3660>

**Summary**
Clean, minimal fix. Replaces a deprecated single-call cancel with an explicit loop that calls `pmpro_cancelMembershipLevel()` per level — which is exactly what the deprecated code path was doing internally. Ready to merge.

**Looks Good**
- `$level->id` is the correct property name — confirmed against `pmpro_cancelMembershipLevel`'s own use of `wp_list_pluck( $user_levels, 'id' )` at `functions.php:1056`. No drift here.
- `pmpro_getMembershipLevelsForUser()` returns an empty array (not `false`) for a valid user with no levels, so the foreach is safe given the `empty( $user_id )` guard already in `pmpro_delete_user` at line 9.
- Default `$include_inactive = false` on `pmpro_getMembershipLevelsForUser` is correct — no reason to try cancelling already-inactive levels.

**Minor note** (not a blocker)
Each iteration calls `pmpro_cancelMembershipLevel()`, which itself re-calls `pmpro_getMembershipLevelsForUser()` to verify the level still exists before cancelling (`functions.php:1055`). For a user with N levels, that's N+1 total queries to that function. Negligible for typical memberships (1-3 levels), and fixing it would require changing `pmpro_cancelMembershipLevel` itself — out of scope here. Just something to be aware of if this pattern shows up elsewhere.

PR #3663, opened 2026-04-29
Add TinyMCE WYSIWYG editor with Liquid autocomplete to email templates

dparker1005:
## Summary
- Replaces the plain textarea on the email template edit page with a vendored TinyMCE 8.5.0 instance, with a toolbar trimmed to email-safe formatting
- Adds Liquid-aware autocomplete: `{` shows variables + tags, `{{` narrows to variables, `|` shows filters when inside an unclosed `{{ ... }}`
- Conditionals (`{% if %} ... {% endif %}`) insert as a paired template with the cursor at the condition; the `default` filter inserts `| default: "` so the cursor lands inside the open quote
- Filter list and descriptions are sourced from `PMPro_Liquid_Renderer::get_filters()` so the dropdown stays in sync with what the renderer actually supports

## Test plan
- [ ] Open a class-based email template (e.g. Checkout - Free) — TinyMCE editor loads
- [ ] Type `{` — initial dropdown shows variables + a `Liquid Tags` separator + the four block tags
- [ ] Pick a variable — `{{ name }}` is inserted
- [ ] Inside an existing `{{ }}`, type `|` — filter dropdown appears
- [ ] Pick `default` — inserts ` | default: "` with cursor after the opening quote
- [ ] Pick `{% if … %} … {% endif %}` from the initial dropdown — paired template inserts, cursor lands at the condition
- [ ] Insert image via toolbar — wp.media opens, selecting an image inserts it
- [ ] Save Template, reload — body persists
- [ ] Reset Template — restores defaults
- [ ] Disable email — editor goes readonly
- [ ] Open a non-class template (legacy `!!var!!` style) — autocomplete is silent (variables/tags do not fire)
- [ ] Send a test email — content renders correctly through `pmpro_kses` sanitization

## Notes
- TinyMCE 8.5.0 is vendored at `includes/lib/tinymce/` (GPL-2.0+, sourced from `npm pack tinymce@8.5.0`)
- `fontfamily`, `media`, `codesample`, `anchor`, `fullscreen` are intentionally excluded — see the comment above `tinymce.init` for why each was dropped
- Full-document templates (with `<html>`/`<head>` wrappers in the saved option) will have those wrappers stripped on save through this editor. PMPro's `header`/`footer` templates already auto-wrap the body, so the visible email rendering is unaffected; the only edge case is users who have disabled both header and footer and rely on the body to carry the document wrapper


dparker1005:
Closing this in favor of #3695, which replaces the TinyMCE 8 approach with WordPress' built-in wp_editor() while keeping Liquid autocomplete support.

PR #3665, opened 2026-05-01
Move Member Notes to its own member edit panel

dparker1005:
## Summary

- Replaces the single Member Notes textarea on the User Info panel with a dedicated Member Notes panel that operates like Order Notes — each entry is timestamped with the author's display name, and entries can be trashed individually. No edit-all affordance.
- Notes are still stored as a string in the `user_notes` user meta (same key as before), using the same `<date> by <author>: <body>` shape order notes use, with notes separated by a blank line. Existing free-form `user_notes` content is parsed as one or more "Legacy note" entries that can be trashed like any other.
- Self-contained in a new `PMPro_Member_Edit_Panel_Member_Notes` class — no new public helper functions or DB migrations.

## Test plan

- [ ] On a user with **no** prior member notes: panel shows "No notes for this member." Add a note → it appears with the current time and your display name. Add a second → both visible, newest at top. Trash one → only the other remains.
- [ ] On a user with **existing** `user_notes` content (free-form text from the old textarea): panel shows it as a "Legacy note" entry. Trashing it removes it from `user_notes`. Adding a new note alongside legacy content works.
- [ ] User Info panel no longer shows the Member Notes textarea; saving the User Info panel does not clobber `user_notes`.
- [ ] `get_user_meta( $user_id, 'user_notes', true )` still returns a string after add/delete (verify via WP-CLI or a quick `var_dump`), so any add-on / export reading that key is unaffected.
- [ ] Multi-line note bodies render correctly (line breaks preserved). Notes containing blank lines get their blank lines collapsed on add (intentional, so the `\n\n` separator stays unambiguous).
- [ ] User without `pmpro_get_edit_member_capability()` cannot add or delete notes.

### DIFF
--- a/adminpages/member-edit.php
+++ b/adminpages/member-edit.php
@@ -14,6 +14,7 @@ function pmpro_member_edit_get_panels() {
 	$panels[] = new PMPro_Member_Edit_Panel_Memberships();
 	$panels[] = new PMPro_Member_Edit_Panel_Subscriptions();
 	$panels[] = new PMPro_Member_Edit_Panel_Orders();
+	$panels[] = new PMPro_Member_Edit_Panel_Member_Notes();
 	$panels[] = new PMPro_Member_Edit_Panel_TOS();
 
 	// Add user fields panels.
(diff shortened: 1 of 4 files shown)


dparker1005:
Holding off on this PR  for now.

We should consider if having a combined "notes" table for member notes, order notes, and maybe subscription notes is a better design long-term. Also worth taking a look at a plugin like WP Activity Log to see if we want to integrate with an existing solution.

dparker1005:
Closing for now. May reopen this conversation in the future.

PR #3682, opened 2026-05-15
Allow shortcodes in account page message; no longer show confirmation message if order is not confirmed

kimcoleman:
### All Submissions:
* [x] Have you followed the [Contributing guidelines](https://github.com/strangerstudios/paid-memberships-pro/blob/dev/.github/CONTRIBUTING.md)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/strangerstudios/paid-memberships-pro/pulls/) for the same update/change?

This PR adds shortcode support to the membership level's confirmation message and account page message, and gates the confirmation page level message behind a successful order status. This ensures shortcodes that gate by membership access (e.g., `[pmpro_member]`, `[pmpro_download]`) render correctly and are only displayed once a user's payment has been confirmed.

**Account page level message** (`includes/functions.php`):
- Added shortcode support to the per-level "Membership Account Message" displayed via the `pmpro_membership_account_after_level_card_content` action. Output now runs through `do_shortcode( shortcode_unautop( wpautop( ... ) ) )` so admins can embed shortcodes (e.g., `[pmpro_member]`, `[pmpro_download]`) in the message.

**Confirmation page level message filter** (`includes/filters.php`):
- Updated `pmpro_pmpro_confirmation_message()` to also process shortcodes — `do_shortcode( shortcode_unautop( wpautop( $message ) ) )` instead of just `wpautop( $message )`. This makes shortcodes render in the level confirmation message on the confirmation page.

**Confirmation page level message gating** (`pages/confirmation.php`):
- Wrapped the level confirmation message output in a new `<div class="pmpro_confirmation-confirmation-message">` element for styling.
- Gated the level confirmation message behind `'success' == $pmpro_invoice->status`, so it no longer renders for pending/non-success orders where shortcodes that gate by membership access would fail.
- Applied `wpautop()` to the level message before the surrounding div wrapper to ensure proper paragraph formatting inside the block element.
- Bumped template version from 3.1 → 3.7.3.

**Email confirmation message rendering** (all 6 email templates in `classes/email-templates/`):
- `class-pmpro-email-template-checkout-paid.php`
- `class-pmpro-email-template-checkout-paid-admin.php`
- `class-pmpro-email-template-checkout-free.php`
- `class-pmpro-email-template-checkout-free-admin.php`
- `class-pmpro-email-template-checkout-check.php`
- `class-pmpro-email-template-checkout-check-admin.php`

Updated the `$confirmation_message` assignment from raw `$membership_level->confirmation` to `do_shortcode( shortcode_unautop( wpautop( $membership_level->confirmation ) ) )`, so paragraphs and shortcodes render in the `{{ membership_level_confirmation_message }}` email merge tag.

1. Edit a membership level and add a confirmation message and account message that include a shortcode (e.g., `[pmpro_member fields="..."]` or `[pmpro_download]`) along with some plain text/paragraphs.
2. Check out for that level as a new user with a paid (success) order.
3. On the confirmation page, verify the level confirmation message renders with proper paragraph formatting, shortcodes are processed, and the message is wrapped in the new `pmpro_confirmation-confirmation-message` div.
4. Check the checkout confirmation email — verify the `{{ membership_level_confirmation_message }}` merge tag renders paragraphs and shortcodes correctly.
5. Visit the Membership Account page and verify the per-level account message renders shortcodes and paragraphs correctly.
6. Create a pending order (e.g., via Pay by Check) and visit the confirmation page — verify the level confirmation message is NOT displayed for the pending order.
7. Update the order status to `success` and reload — verify the level confirmation message now appears.

### Other information:

### Changelog entry

* ENHANCEMENT: Added shortcode support to the membership level's confirmation message (on the checkout confirmation page and in checkout confirmation emails) and to the per-level account page message, so shortcodes like `[pmpro_member]` and `[pmpro_download]` now render in these areas.
* BUG FIX/ENHANCEMENT: The membership level's confirmation message on the checkout confirmation page is now only displayed for orders with a `success` status, preventing display issues with membership-gated shortcodes for pending or non-success orders.
* ENHANCEMENT: Wrapped the level confirmation message on the checkout confirmation page in a new `pmpro_confirmation-confirmation-message` div for improved styling control. Template version will need to be bumped (TBD)

### DIFF
--- a/classes/email-templates/class-pmpro-email-template-checkout-check-admin.php
+++ b/classes/email-templates/class-pmpro-email-template-checkout-check-admin.php
@@ -147,7 +147,7 @@ class PMPro_Email_Template_Checkout_Check_Admin extends PMPro_Email_Template {
 
 		$confirmation_in_email = get_pmpro_membership_level_meta( $membership_level->id, 'confirmation_in_email', true );
 		if ( ! empty( $confirmation_in_email ) ) {
-			$confirmation_message = $membership_level->confirmation;
+			$confirmation_message = do_shortcode( shortcode_unautop( wpautop( $membership_level->confirmation ) ) );
 		} else {
 			$confirmation_message = '';
 		}
--- a/classes/email-templates/class-pmpro-email-template-checkout-check-admin.php
+++ b/classes/email-templates/class-pmpro-email-template-checkout-check-admin.php
@@ -147,7 +147,7 @@ class PMPro_Email_Template_Checkout_Check_Admin extends PMPro_Email_Template {
 
 		$confirmation_in_email = get_pmpro_membership_level_meta( $membership_level->id, 'confirmation_in_email', true );
 		if ( ! empty( $confirmation_in_email ) ) {
-			$confirmation_message = do_shortcode( shortcode_unautop( wpautop( $membership_level->confirmation ) ) );
+			$confirmation_message = $membership_level->confirmation;
 		} else {
 			$confirmation_message = '';
 		}
--- a/classes/email-templates/class-pmpro-email-template-checkout-check.php
+++ b/classes/email-templates/class-pmpro-email-template-checkout-check.php
@@ -201,7 +201,7 @@ class PMPro_Email_Template_Checkout_Check extends PMPro_Email_Template {
 		$confirmation_message = '';
 		$confirmation_in_email = get_pmpro_membership_level_meta( $membership_level->id, 'confirmation_in_email', true );
 		if ( ! empty( $confirmation_in_email ) ) {
-			$confirmation_message = do_shortcode( shortcode_unautop( wpautop( $membership_level->confirmation ) ) );
+			$confirmation_message = $membership_level->confirmation;
 		}
 
 		$email_template_variables = array(
--- a/classes/email-templates/class-pmpro-email-template-checkout-check.php
+++ b/classes/email-templates/class-pmpro-email-template-checkout-check.php
@@ -201,7 +201,7 @@ class PMPro_Email_Template_Checkout_Check extends PMPro_Email_Template {
 		$confirmation_message = '';
 		$confirmation_in_email = get_pmpro_membership_level_meta( $membership_level->id, 'confirmation_in_email', true );
 		if ( ! empty( $confirmation_in_email ) ) {
-			$confirmation_message = $membership_level->confirmation;
+			$confirmation_message = do_shortcode( shortcode_unautop( wpautop( $membership_level->confirmation ) ) );
 		}
 
 		$email_template_variables = array(
--- a/classes/email-templates/class-pmpro-email-template-checkout-free-admin.php
+++ b/classes/email-templates/class-pmpro-email-template-checkout-free-admin.php
@@ -168,7 +168,7 @@ class PMPro_Email_Template_Checkout_Free_Admin extends PMPro_Email_Template {
 		$confirmation_message = '';
 		$confirmation_in_email = get_pmpro_membership_level_meta( $membership_level->id, 'confirmation_in_email', true );
 		if ( ! empty( $confirmation_in_email ) ) {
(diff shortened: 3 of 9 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3682 — Allow shortcodes in confirmation message and account page message; no longer show confirmation message if order is not confirmed**
*kimcoleman → dev | 9 files, +18 −14*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3682>

**Summary**

Solid change that wires up the correct WordPress shortcode pipeline (`wpautop` → `shortcode_unautop` → `do_shortcode`) consistently across three surfaces: the confirmation page, 6 email templates, and the account page. The success-status gate in the confirmation template is the right call — access-gated shortcodes would silently fail for pending orders. One unresolved placeholder and one CSS naming issue need attention before merge.

**Issues**

- **Major** `pages/confirmation.php:4,10` — Both `Version:` and `@version` in the file header are set to `"TBD"`. PMPro's template version-checking system compares this string against the bundled version to warn site owners when their customized template is outdated. `"TBD"` won't match any real version string, so the upgrade notice will never fire for anyone who has customized this template. The current plugin version is 3.7.3; set the template version to the actual next release version (or coordinate with whatever cut this is targeting) before merging.

- **Minor** `pages/confirmation.php:53` — The new CSS class `pmpro_confirmation-confirmation-message` has "confirmation" twice. Likely a copy-paste artifact. `pmpro_confirmation-level-message` or `pmpro_level-confirmation-message` would be unambiguous.

**Looks Good**

- The shortcode pipeline pattern (`do_shortcode( shortcode_unautop( wpautop( $message ) ) )`) is the correct WordPress idiom — `shortcode_unautop` undoes auto-paragraph wrapping around shortcodes before processing, preventing block-level shortcodes from being stranded inside `<p>` tags.
- The `'success' == $pmpro_invoice->status` gate is the right boundary. Free level orders land at `success` immediately, so they aren't blocked.
- `esc_attr()` on the new class in the confirmation template and `wp_kses_post()` wrapping the `do_shortcode()` output on the account page (`functions.php:5192`) — both correct.
- All 6 email templates changed consistently; no partial coverage.

**Questions**

- For async gateway flows (PayPal IPN callbacks, webhook-triggered email sends), `wp_get_current_user()` returns 0 — no logged-in user. If `[pmpro_member]` or similar access-gated shortcodes are used in a confirmation message, they'd render the "not a member" version in the email for those gateways. Is that an expected edge case, or does PMPro always send these emails synchronously within the checkout request?

kimcoleman:
@dparker1005 I did the second thing Flint suggested here - a more specific CSS selector - but I did not adjust the TBD since that is a key piece of the release process. This one is ready for you.

dparker1005 (changes requested):
Thanks for the updates here. I think this is close, but I’d like us to think through one remaining edge case before merging.

The checkout email templates now run `do_shortcode()` on the level confirmation message while building `membership_level_confirmation_message`. That seems right for shortcodes that are self-contained, but I’m not sure we have a safe member context for all email-send paths.

For example, `[pmpro_member]` defaults to `$current_user->ID`. Checkout emails can be generated from async/webhook/admin/test contexts where the current WP user may not be the member receiving the email. In those cases, member-context shortcodes could render empty or for the wrong user.

The tricky part is that simply not rendering shortcodes in emails is not a great fallback either, because we also don’t want raw shortcode text showing in customer emails. Stripping shortcodes would avoid the raw text, but then it largely defeats the purpose of enabling shortcodes in the confirmation message email variable.

I think we should decide on the intended behavior explicitly. Maybe email confirmation-message shortcodes need to be rendered under the recipient/member context, via a helper that accepts the user/order context and restores the previous current user afterward. Or, if we don’t want to support all shortcodes in email contexts, we should document/enforce that and strip unsupported shortcodes before sending.

The confirmation page and account page changes look good to me. The CSS selector concern from the earlier review is fixed. I’m not blocking on `Version: TBD` since that’s part of the release process, but it still needs to be replaced before packaging.

PR #3684, opened 2026-05-15
Adds a new `level_group_id` template variable to every email template class that already exposes `membership_id`

kimcoleman:
### All Submissions:
* [x] Have you followed the [Contributing guidelines](https://github.com/strangerstudios/paid-memberships-pro/blob/dev/.github/CONTRIBUTING.md)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/strangerstudios/paid-memberships-pro/pulls/) for the same update/change?

This PR adds a new `level_group_id` template variable to every email template class that already exposes `membership_id`. The real benefit: site admins can now share a single email template across all levels in a group (e.g., all "Courses" levels) using Liquid conditional logic, instead of maintaining a separate email per level.

For example, in an email body:
`{% if level_group_id == 2 %}The Level Group ID is: {{ level_group_id }}{% endif %}`

**What was changed:**
- Added a new `level_group_id` template variable (with description) to every email template class that already exposes `membership_id`. The description is consistent across all templates: "The ID of the level group that the membership level belongs to."
- Uses the existing `pmpro_get_group_id_for_level()` helper from `includes/level-groups.php` to resolve the group ID from the level ID.
- 21 email template classes updated; the 4 templates that don't expose `membership_id` (`admin-change`, `admin-change-admin`, `payment-action`, `payment-action-admin`) were intentionally skipped.

**Cancel templates** (`cancel.php` / `cancel-admin.php`): `$cancelled_level_ids` can be empty, an array of IDs, or a single ID. `level_group_id` mirrors the shape of `membership_id` in each case — empty string when no levels, group of the first cancelled level when an array, or group of the single level when one ID.

**Membership expired / expiring templates:** When the user has no current level, `membership_id` falls back to `0`, and `level_group_id` falls back to `0` to match.

1. Set up at least one level group and assign a membership level to it.
2. Edit one of the affected email templates and add the following to the email body:

`{% if level_group_id == 2 %}The Level Group ID is: {{ level_group_id }}{% endif %}`

(Adjust the group ID to match the group ID assigned to your test level.)
3. Trigger the corresponding email (e.g., complete a checkout) for a user on that level.
4. Verify the conditional output renders correctly with the group ID.
5. Test the cancel email by cancelling a level — verify `level_group_id` resolves to the cancelled level's group ID.
6. Test the membership expired email for a user with no current level — verify `level_group_id` is `0`.

### Other information:

### Changelog entry
* ENHANCEMENT: Added a new `level_group_id` variable to email templates that already expose `membership_id` (checkout, cancel, membership expired/expiring, billing, and related templates).


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3684 — Adds a new `level_group_id` template variable to every email template class that already exposes `membership_id`**
*kimcoleman → dev | 21 files, +46 -0 lines*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3684>


**Summary**
Adds `level_group_id` as a Liquid template variable to the 21 email template classes that already expose `membership_id`. The approach is consistent and the coverage is correct. One issue worth addressing before merge: `pmpro_get_group_id_for_level()` returns `false` for ungrouped levels — not `0` — which creates a behavioral gap the PR description doesn't acknowledge.


**Issues**

- **Major** `includes/level-groups.php:192` — `pmpro_get_group_id_for_level()` returns `false` (not `0`) when a level isn't assigned to any group: `return empty( $group_id ) ? false : $group_id;`. Every template in this PR passes that return value directly to the Liquid context without coalescing. For a level with no group, `{{ level_group_id }}` renders as an empty string, and `{% if level_group_id == 0 %}` won't reliably match — Liquid doesn't equate PHP `false` to `0`. This is the default state for anyone who hasn't set up level groups, which is most sites. The PR description says "falls back to `0`" but the code doesn't implement that. Fix: coalesce every call site — `pmpro_get_group_id_for_level( $level_id ) ?: 0` — or normalize inside the helper itself. This affects all 21 templates.

- **Minor** `class-pmpro-email-template-membership-expired.php:355` / `class-pmpro-email-template-membership-expiring.php:375` — The ternary here guards for a missing `$membership_level` object (correctly falling back to `0`) but still passes the raw `false` return through `pmpro_get_group_id_for_level()` when the level exists but has no group. After applying the `?: 0` fix above, this pattern becomes correct; without it, it's only half-protected.


**Looks Good**

The pattern is applied uniformly across all 21 classes — description, variable list, and runtime assignment all updated together with no missed files. The cancel template handling (`cancel.php`, `cancel-admin.php`) mirrors the shape of `membership_id` correctly across all three branches (empty, array, scalar). The decision to skip the 4 templates that don't expose `membership_id` is the right call.


**Questions**

1. Does `pmpro_get_group_id_for_level()` returning `false` vs `0` matter to Liquid's comparison operators in practice? Worth a quick test with `{% if level_group_id == 0 %}` on an ungrouped level before this ships — if it silently doesn't match, site admins debugging their templates won't have a great time.

PR #3685, opened 2026-05-15
Fix subtotal and tax not being set correctly if Stripe Tax is enabled

dwanjuki:
### All Submissions:

**1. Checkout Session handler**
_Before:_ `subtotal` was set from `checkout_session->amount_subtotal` before tax was calculated. If _Membership price includes tax_ is selected, `amount_subtotal` includes the tax amount.

_After:_ tax is set first, then `subtotal = total - tax`. This ensures subtotal always represents the pre-tax amount regardless of whether Stripe Tax is inclusive or exclusive.

  ---
**2. Recurring invoice handler**
_Before:_ Tax for recurring orders was read from `invoice->tax`, and subtotal from `invoice->subtotal`.
`invoice->tax` is not available in Stripe API version 2025-09-30.clover, so it is always null, meaning recurring orders were saved with `tax = 0` even when Stripe Tax was enabled. Additionally, `invoice->subtotal` does not deduct tax when _Membership price includes tax_ is selected.

_After:_ Tax and subtotal are both derived from `invoice->total_excluding_tax`.
When it is set (Stripe Tax is active), `subtotal = total_excluding_tax` and `tax = total - total_excluding_tax`. When it is null (no Stripe Tax), `subtotal = total` and `tax = 0`. This works correctly for both inclusive and exclusive Stripe Tax modes.

1. Configure Stripe Tax in Stripe test mode.
2. Enable price inlcusive Stripe Tax option in PMPro settings.
3. Create a level with an intial payment and recurring billing amount.
4. Complete checkout.
5. Confirm that tax, order, and subtotal amounts shown in Stripe are set on the PMPro Order and that subtotal + tax = total.
7. Advance clocks past renewal date.
8. Confirm that tax, order, and subtotal amounts shown in Stripe are set on the recurring PMPro Order and that that subtotal + tax = total.
9. Retest with price exclusive Stripe Tax option and with Stripe Tax disabled. Confirm no errors occur.

### Other information:

### Changelog entry

> Fixed the subtotal and tax not being set correctly on orders created via Stripe when Stripe Tax is enabled and tax is applied.


andrewlimaza:
Adding a small note here, that Stripe adds a new taxes property which may be nice to use if we want itemized and multiple tax rates shown (eventually) - https://docs.stripe.com/changelog/basil/2025-03-31/invoice-tax-configurations

dparker1005:
Thanks @dwanjuki — the deprecated `invoice->tax` is a real regression and we're going to ship that fix.

The `subtotal` change is trickier. Two conventions for storing `subtotal` / `tax` / `total`:

- **US sales tax (exclusive)**: `subtotal` is pre-tax, `total = subtotal + tax`. Invoices show all three lines.
- **EU VAT (inclusive)**: `subtotal = total` — the customer-facing price already includes VAT. `tax` is shown as an informational breakout.

When a level is configured with inclusive Stripe Tax, Stripe returns `amount_subtotal == amount_total` and `invoice->subtotal == invoice->total` — matching the EU VAT convention. PMPro has been storing orders that way since we added Stripe Tax support, so EU VAT customers have order data, CSV exports, and downstream integrations built around `subtotal = total`.

Switching to `subtotal = total - tax` for inclusive sites would unify everything on the US-style invariant. It's defensible, but it's a silent BC change for every inclusive Stripe Tax customer.

Going to err on the side of consistency-with-the-past in this PR: keep the existing `subtotal` behavior, only swap the deprecated `invoice->tax` for the derived `total - total_excluding_tax`. The unification question is worth its own discussion if there's appetite to migrate later.

PR #3702, opened 2026-06-11
A11y: Implement live regions to announce when autocomplete liquid listbox opens and when navigating options

RachelRVasquez:
### All Submissions:

The Liquid autocomplete in the email template editor was silent for screen reader users. The listbox lives in the admin document while focus stays inside the TinyMCE iframe, so `aria-activedescendant` cannot resolve across the two documents and the usual combobox pattern never engages. This PR adds ARIA live regions to announce the list instead.

- Announces "Autocomplete list opened" plus the first option when the list appears.
- Announces each option with its position ("3 of 24") as you arrow through the list. Arrow navigation is assertive; opening and filtering are polite so they do not talk over the screen reader's own typing echo.
- Announces "Autocomplete list closed" when the list is dismissed with Escape, by clicking elsewhere in the editor, or by filtering until nothing matches. Choosing an option, or moving focus out of the editor, stays silent because the inserted tag or the new focus target is the feedback.
- Announcements are debounced by 150ms so a fast burst of keystrokes is described once, when the user pauses, rather than once per character. Identical repeats are skipped.
- Adds the `autocompleteOpened`, `autocompleteClosed` and `autocompletePosition` strings to `pmpro_get_liquid_autocomplete_settings()` so the announcements are translatable.
- `shouldIgnoreKeyup()` now also ignores Shift, Ctrl, Alt, Caps Lock and Meta. Modifier keyups were re-running the autocomplete scan and producing duplicate announcements.

Originally opened by @RachelRVasquez. Rebased onto `v3.9`, finished and taken over after she left the team. Follows #3701, which has already merged.

1. Test with VoiceOver + Safari and/or [NVDA](https://www.nvaccess.org/download/) + Firefox/Chrome.
2. In wp-admin go to Memberships > Settings > Email Templates and open a template that uses Liquid syntax, for example **Checkout - Free**. Legacy `!!variable!!` templates do not load the autocomplete at all.
3. Click into the "Body" TinyMCE field.
4. Type `{{` (two braces). The listbox should appear and be announced as "Autocomplete list opened", followed by the first option and its position.
5. Press the down and up arrows. Each option should be read out with its position, for example "3 of 24".
6. Keep typing to filter the list. The list should be re-described once you pause, without interrupting your own typing.
7. Press Escape. "Autocomplete list closed" should be announced. Reopen with `{{` and click elsewhere in the editor body; that should announce it closed too.
8. Reopen with `{{` and type letters until nothing matches. "Autocomplete list closed" should be announced.
9. Reopen with `{{` and press Enter, or click an option with the mouse. The tag should be inserted with no announcement.
10. Reopen with `{{` and Tab out of the editor. The list should close with no announcement.

Verified so far by asserting the text of the live regions in the DOM after each action (Playwright, Chromium). It has **not** yet been verified with a real screen reader, which is the remaining step before this is fully signed off.

### Other information:

```
* ENHANCEMENT: Added screen reader announcements to the Liquid autocomplete in the email template editor.
```


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3702 — A11y: Implement live regions to announce when autocomplete liquid listbox opens and when navigating options**
*RachelRVasquez → v3.9 | 2 files, +191 -15*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3702>

---

**Summary**
Not ready. The open/navigate announcement logic is careful and correct, but the new "closed" announcement is suppressed on the most common dismissal paths (blur, click-away, editor hide) because the TinyMCE event object leaks into the new `silent` parameter. One-line fix, then this is good to merge after #3701.

---

**Issues**

- **Major** `js/pmpro-liquid-autocomplete.js:745` — `editor.on( 'click blur hide', closeMenu );` (pre-existing, unchanged) now hands TinyMCE's event object to the new `closeMenu( silent )` signature at `:160`. The event object is truthy, so `if ( wasOpen && ! silent )` at `:188` is always false on those paths and "Autocomplete list closed" never fires. Only Escape (`:727`, bare `closeMenu()`) announces. Blur and click-away are how most users leave the listbox without choosing, so in practice the close announcement mostly does not happen, contradicting the PR description and test step 6. Also, with `silent` truthy, the `! wasOpen && ! silent` early-return at `:174` is skipped, so a stray click on an already-closed editor runs `clearAnnouncers()` and can cancel a pending "opened" or navigation message. Fix: `editor.on( 'click blur hide', function () { closeMenu(); } );` and keep `closeMenu( true )` for the two intentional silent paths (`chooseActive()` at `:659`, `remove` at `:752`).

- **Minor** `js/pmpro-liquid-autocomplete.js:173-193` — After the early return at `:174`, the `wasOpen &&` in `if ( wasOpen && ! silent )` at `:188` is redundant: any non-silent call that reaches that line already had `wasOpen` true. Not a bug, but the two guards read as if they cover different cases. Collapse to `if ( ! silent )` or restructure as a single `if/else` so the intent is obvious.

- **Minor** `js/pmpro-liquid-autocomplete.js:751-752` — The `remove` handler calls `closeMenu( true )` but leaves both announcer divs appended to `document.body` by `ensureAnnouncer()`. `getElementById` reuse means it doesn't compound across re-inits, but it is untidy on pages that tear down the editor. Remove both `ANNOUNCER_IDS` elements in the `remove` handler.

---

**Looks Good**
- `announce()` (`:128-146`) clears `textContent` before the delayed set, which is the reliable way to force re-reading of identical strings, and cancelling the pending timer coalesces the multi-render-per-keystroke burst.
- `openPrefixPending` (`:313-315`, `:349-351`) correctly handles the `{{` case where the second render would otherwise overwrite the queued "opened" prefix before it ever hit the DOM. The inline comments explain the why succinctly.
- Filter dedupe via `lastAnnounced` (`:337`) plus polite for open/filter and assertive for arrow navigation is the right politeness split.
- All announcer text goes through `textContent`, and the labels originate in hard-coded PHP suggestion arrays, so no injection surface.
- The `%1$d of %2$d` string with a translators comment lets translators reorder position and total. The JS `.replace()` on the literal tokens works for that.

---

**Questions**
1. Was step 6 of the test plan exercised via blur or click-away, or only Escape? That would explain how the Major above slipped through.
2. `shouldIgnoreKeyup` (`:692-706`) now also ignores Shift, Ctrl, Alt, Caps Lock, and Meta. Is that part of the a11y work (modifier keyups causing spurious re-renders and duplicate announcements), or a separate fix? Worth a line in the description either way.

flintfromthebasement (approved):
**Follow-up review**

**PR: strangerstudios/paid-memberships-pro#3702 — A11y: Implement live regions to announce when autocomplete liquid listbox opens and when navigating options**
*RachelRVasquez → v3.9 | 2 files, +193 -15*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3702>

**Summary**
The blocking issue is fixed — ready to merge. The prior Major (blur/click-away silently swallowed by TinyMCE's event object leaking into `silent`) is resolved with the exact fix suggested last time.

**Prior issues — status**

- **Resolved** (was Major) `js/pmpro-liquid-autocomplete.js:742-747` — `editor.on( 'click blur hide', closeMenu )` is now split: `click` calls `closeMenu()` (announces — focus stays in the editor) and `blur hide` call `closeMenu( true )` (silent — focus moves somewhere a screen reader already describes). No event object reaches `silent` anymore. Comment at `:737-741` explains the split correctly.
- **Resolved** (was Minor, redundant `wasOpen &&`) `js/pmpro-liquid-autocomplete.js:181-196` — collapsed to `if ( ! silent )` for the announce branch as suggested; the early-return guard at `:181` is unchanged and still correctly gated on `wasOpen`. Reads clean now.
- **Not addressed** (Minor) `js/pmpro-liquid-autocomplete.js:753-760` — the `remove` handler still doesn't clean up the two `ANNOUNCER_IDS` divs appended to `document.body` by `ensureAnnouncer()`. Not a blocker, still worth a `getElementById(id).remove()` pass alongside the menu node removal at `:757-759`.

**New in this diff**
- `js/pmpro-liquid-autocomplete.js:80-84` — the mousedown handler on menu items dropped the `'pointer'` reason (`setActive( index )` instead of `setActive( index, 'pointer' )`), and `announceActive()`'s `'pointer'` early-return was removed entirely (`:307` docblock now lists only `'open' | 'filter' | 'navigate'`). This looks like dead-code cleanup rather than a behavior change: `chooseActive()` runs synchronously right after `setActive()` at `:84`, and its `closeMenu( true )` → `resetAnnouncements()` cancels the just-scheduled `announceTimer` before the 150ms delay ever fires, so a mouse/touch selection still produces no spoken announcement — same end result as the old explicit early return, just via the existing cancel-in-flight path instead of a special case. No regression found; worth a quick manual click-select check with a screen reader running to be sure, since the reasoning is timing-dependent rather than structurally guaranteed.
- `js/pmpro-liquid-autocomplete.js:112-128` — `clearAnnouncers()` renamed to `resetAnnouncements()` and now also resets `lastAnnounced`/`openPrefixPending` internally, removing the duplicated reset calls that used to sit at the `closeMenu()` call site. Good consolidation.

**Looks Good**
- The `click` vs `blur hide` split matches the reasoning from the original review precisely, including keeping the two genuinely-silent paths (`chooseActive()`, `remove`) untouched.
- Comments were updated in the same diff to match the new behavior instead of going stale.

**Questions**
None new — my prior question about the modifier-key changes to `shouldIgnoreKeyup()` wasn't touched by this diff, so it stands as a documentation nit only, not a merge blocker.

dparker1005:
Hi @flintfromthebasement, thank you for catching the blur/click-away suppression, that would not have surfaced in any of the typing-path tests.

Fixed in 2a8e2b6b2 and 901aa872f: `click` now announces the close while `blur`/`hide` stay silent, the redundant `wasOpen` guard is collapsed, and the dead `pointer` announcement case is removed.

Leaving the announcer divs in place on `remove`: they're looked up by ID and shared across editor instances on the page, so tearing them down when one editor is removed would silence any other instance still mounted. Two empty divs on `document.body` is the cheaper trade.

The `shouldIgnoreKeyup()` modifier change is part of this work (modifier keyups re-ran the scan and produced duplicate announcements) and is now called out in the description. The remaining step before merge is a real VoiceOver/NVDA pass.

kimcoleman:
Closing in favor of #3819, which supersedes this PR. It keeps Rachel's approach and strings and David's follow-up fixes, with the announcement mechanism trimmed to use core's wp.a11y.speak() after testing with NVDA. Both are credited as co-authors on the new commit. Thank you both for the groundwork.

PR #3704, opened 2026-06-12
Restrictable taxonomies: protect custom taxonomy terms by level, stop level-save wiping term restrictions

kimcoleman:
## What

Generalizes PMPro's term-level protection (previously hardcoded to `category`/`post_tag`) behind a new `pmpro_restrictable_taxonomies` filter, mirroring the existing `pmpro_restrictable_post_types` pattern. Add Ons can now protect their own taxonomy terms by membership level with a one-line filter — the term "Require Membership" UI, access checks, search/archive filtering, edit-level checklists, and term-delete cleanup all follow automatically.

Also fixes a shipped data-loss bug: **saving any membership level wiped that level's `post_tag` restriction rows** (protectable via the term edit screen since 2.10, but the level-save path deleted all of the level's rows in `pmpro_memberships_categories` and re-inserted only the Categories checklist).

## Changes

- **`includes/terms.php`** — New `pmpro_get_restrictable_taxonomies()` + `pmpro_restrictable_taxonomies` filter (default `category`, `post_tag`). New `pmpro_init_term_restrictions()` (on `init` 99) wires the term add/edit form fields, save handler, and delete cleanup for every restrictable taxonomy, replacing the hardcoded `add_action` calls.
- **`includes/content.php`** — `pmpro_has_membership_access()` gathers terms from restrictable taxonomies registered for the post's post type instead of special-casing `post_type == "post"`. Term-derived levels now also populate the returned `$level_ids`/`$level_names` for CPTs (templates can render "Requires X membership" + direct checkout links). Also fixes a latent bug in `pmpro_search_filter()`: the category subqueries compared `tr.term_taxonomy_id` against stored **term IDs** — now joined through `term_taxonomy`, so split-term sites filter correctly.
- **`includes/functions.php`** — New `pmpro_list_restrictable_terms()` (hierarchical + flat checklists, new `pmpro_list_restrictable_terms_args` filter); `pmpro_listCategories()` kept as a back-compat wrapper. `pmpro_updateMembershipCategories()` now scopes its DELETE to terms in currently restrictable taxonomies (plus orphaned terms) so restrictions saved by an inactive Add On survive level saves.
- **`adminpages/levels/edit-level.php`** — The Content Settings section renders one checklist row per restrictable taxonomy (Categories, Tags, and any Add On taxonomy). The per-row advanced-settings messaging is consolidated into a single section-level note ("Based on your advanced settings… Display can vary by content type and theme.") since we can't promise rendering behavior for arbitrary CPTs/themes.
- **`classes/class-pmpro-levels.php`** — `save()` reuses `pmpro_updateMembershipCategories()` instead of its own DELETE-all + category-only re-insert.
- **`includes/cleanup.php`** — `pmpro_delete_category()` registration moved into the restrictable-taxonomies loop (`delete_{$taxonomy}`).

## Add On usage

```php
function my_addon_restrictable_taxonomies( $taxonomies ) {
	$taxonomies[] = 'my_custom_taxonomy';
	return array_unique( $taxonomies );
}
add_filter( 'pmpro_restrictable_taxonomies', 'my_addon_restrictable_taxonomies' );
```

## Behavior changes to note in changelog

1. **BUG FIX**: Saving a level no longer wipes `post_tag` (or Add On taxonomy) restriction rows.
2. **BEHAVIOR**: Sites that attach the built-in `category`/`post_tag` taxonomies to a CPT will now have term protection *enforced* for that CPT at view time (previously the checkboxes silently did nothing for non-`post` types). This is the checkbox doing what admins already expected, but it can lock previously visible content after updating — worth a release-post callout.
3. **UI**: The edit level screen now shows a Tags checklist (and one per Add On taxonomy).
4. Term restriction hooks are now registered on `init` 99 instead of at file load; customizations that `remove_action()` them must run later than that.
5. `pmpro_updateMembershipCategories()` no longer deletes rows it cannot re-create (documented in the docblock).
6. Edit-level copy changed: per-row "Non-members will not see posts in these categories" messages replaced by a single hedged section note; translators get new strings.

## Testing done

Tested end-to-end on a local site via wp-cli with the Downloads Add On registering `pmpro_download_category` as restrictable:

- Logged-out user denied on a download restricted only via a protected term; member of the level allowed.
- `pmpro_has_membership_access( $id, null, true )` returns the term-derived level IDs/names for a CPT (locked templates render level name + single-level checkout URL).
- Regression: posts in protected categories and tags behave identically to before.
- Level save with term checked keeps the row; level save while the Add On taxonomy is *not* restrictable (deactivated Add On) preserves its rows; orphaned rows for deleted terms are cleaned up.
- Term delete removes its restriction rows for custom taxonomies.


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3704 — Restrictable taxonomies: protect custom taxonomy terms by level, stop level-save wiping term restrictions**
*kimcoleman → dev | 6 files, +173 -84 lines*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3704>

---

**Summary**

This is solid work. The data-loss fix in `pmpro_updateMembershipCategories()` is correct — the scoped DELETE preserves rows from inactive add-ons and orphans out deleted terms cleanly. The SQL search filter fix (joining through `term_taxonomy` to compare term_taxonomy_ids properly) is the right solution for split-term sites. Two issues to address before merge: a missing static cache in a hot-path function, and a behavior-change that will lock CPT content on upgrade with no changelog entry to warn admins.

---

**Issues**

**Major** `includes/terms.php:18` — `pmpro_get_restrictable_taxonomies()` re-fires `apply_filters()` on every call with no caching. It's called from `pmpro_has_membership_access()`, which fires via `the_content` filter on every post rendered in a loop — on a 10-post archive, that's 10 full filter-chain executions per page load. Any add-on that registers a `pmpro_restrictable_taxonomies` callback with any I/O (option fetch, DB check) will pay that cost repeatedly. Fix:

```php
function pmpro_get_restrictable_taxonomies() {
    static $taxonomies = null;
    if ( $taxonomies === null ) {
        $taxonomies = apply_filters( 'pmpro_restrictable_taxonomies', array( 'category', 'post_tag' ) );
    }
    return $taxonomies;
}
```

Taxonomies are registered at `init` and are static per request, so this is safe.

**Major** No changelog or upgrade notice for the CPT behavior change. From the PR description: "Sites that attach the built-in `category`/`post_tag` taxonomies to a CPT will now have term protection enforced for that CPT at view time (previously the checkboxes silently did nothing for non-`post` types)." This is the correct behavior, but it will silently lock previously-visible content on upgrade. A clear upgrade notice (not just a changelog line) is needed — admins won't expect content to become inaccessible after a minor update with no warning.

**Minor** `classes/class-pmpro-levels.php:180` — `array_filter( array_map( 'intval', $this->categories ), 'term_exists' )` relies on `term_exists()` returning a truthy array when found and null when not. It works, but `term_exists($int)` with no taxonomy argument matches any term with that ID across all taxonomies — not just restrictable ones. In practice this is harmless since `$this->categories` is sourced from the admin form which only presents restrictable taxonomy checkboxes, but an explicit lambda makes the intent clear: `array_filter( $cats, fn( $id ) => (bool) term_exists( $id ) )`.

**Minor** `includes/functions.php:1562` — `pmpro_updateMembershipCategories()` runs a DELETE followed by a loop of INSERTs with no transaction wrapper. If an INSERT fails mid-loop, the level is left with partial category associations. This is a pre-existing pattern, and the PR already touches this function — wrapping the DELETE + INSERT sequence in `START TRANSACTION` / `COMMIT` with a `ROLLBACK` on error would close the gap while you're here.

---

**Looks Good**

The DELETE query in `pmpro_updateMembershipCategories()` — the LEFT JOIN scoping (`tt.term_taxonomy_id IS NULL OR tt.taxonomy IN (...)`) correctly handles both orphaned terms and inactive add-on taxonomy rows without clobbering them. That's the right logic for a non-trivial constraint.

The search filter join fix in `pmpro_search_filter()` (queries B and D) is correct — the original comparison of `tr.term_taxonomy_id` against raw `category_id` values was the source of the split-term bug, and routing through `term_taxonomy` on `term_id` fixes it cleanly.

`pmpro_init_term_restrictions()` on `init:99` is the right registration pattern — late enough for add-on taxonomies to register, early enough for all admin term actions to fire after it.

---

**Questions**

1. Does the `pmpro_list_restrictable_terms_args` filter interact safely with the existing `pmpro_list_categories_args` filter for the `category` taxonomy? Both fire for categories in sequence — if a site filter modifies `taxonomy` in `pmpro_list_categories_args`, it could break the general filter's taxonomy assumption. Worth a note in the docblock.

2. The `pmpro_term_add_form_fields` and `pmpro_term_edit_form_fields` hooks are now wired for every restrictable taxonomy, but the form fields render a static `pmpro-memberships-checklist` ID regardless of which taxonomy context they're in. On a term-edit screen for a custom taxonomy with multiple restrictable taxonomies registered simultaneously, is there a double-render risk?

dparker1005:
I definitely think that `pmpro_restrictable_taxonomies` is the right pattern, just like we have `pmpro_restrictable_post_types`. The only thing that I think is worth a conversation is the settings that we should expose on the Edit Level page

For all post types, the restriction flow is that you go to the piece of content that you want to restrict, then choose the levels to restrict to. This is also how “tags” worked previously.

“Categories” was special though and could be restricted from the Edit Level page itself. I’m sure this made sense at the time, but I believe it’s the only restriction flow that currently works this way across our entire codebase.

So my question is if we want to lean into the legacy “categories” restriction flow that we have and make ALL restrictable taxonomies editable on the Edit Level page, or keep things simple and lean into only allowing setting restriction settings when editing the content being restricted.

I’d lean towards reducing the number of settings on the Edit Level page and maybe only keeping the categories option on the Edit Level page as a “special case” for people who are used to that option there, but happy to discuss further.

PR #3706, opened 2026-06-21
Prevent fatal on billing page when subscription gateway object is null

kimcoleman:
## Problem

Loading the billing page for a subscription whose gateway can't be resolved throws a fatal:

```
Uncaught Error: Call to a member function supports() on null
in pages/billing.php on line 66
```

## Cause

`PMPro_Subscription::get_gateway_object()` returns `null` when the subscription's `gateway` is empty or its gateway class isn't loaded. `pages/billing.php` then called `$subscription_gateway_obj->supports( 'payment_method_updates' )` in three branches without a null check, white-screening the page.

## Fix

Compute the support value once with an `is_object()` guard and reuse it in all three branches. When the gateway object is `null`, `$payment_method_updates` is `false`, so the page falls through to the existing "update your billing information via your payment gateway dashboard" message instead of fataling.

## To reproduce / test

- Visit the billing page for a subscription with a blank/unknown `gateway` value (e.g. `/membership-billing/?pmpro_subscription_id=N`). Before: fatal. After: the off-site update message renders.
- Regression: a normal subscription on the site's default gateway still shows the on-site billing form as before.

### DIFF
--- a/pages/billing.php
+++ b/pages/billing.php
@@ -62,8 +62,11 @@
 						// Check if the gateway for this subscription updates a single subscription at once or all subscriptions at once.
 						$subscription_gateway_obj = $pmpro_billing_subscription->get_gateway_object();
 
+						// Gateway object is null when the subscription's gateway is unset or unavailable.
+						$payment_method_updates = is_object( $subscription_gateway_obj ) ? $subscription_gateway_obj->supports( 'payment_method_updates' ) : false;
+
 						// If it's not an 'all' update method, we can show specific level information.
-						if ( 'all' !== $subscription_gateway_obj->supports( 'payment_method_updates' ) ) {
+						if ( 'all' !== $payment_method_updates ) {
 							// Show the cost text for the subscription.
 							?>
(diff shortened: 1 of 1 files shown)


flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3706 — Prevent fatal on billing page when subscription gateway object is null**
*kimcoleman → dev | 1 file, +6 -3*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3706>

**Summary**
Clean, surgical null-guard fix for a genuine fatal. Pre-computing `$payment_method_updates` once and reusing it across all three branches is strictly better than the original three redundant `->supports()` calls. Ready to merge.

**Looks Good**
- All three vulnerable call sites are patched in a single consistent move — not just the one that triggered the report. That's the right scope.
- Fallback behavior with a null gateway (`$payment_method_updates = false`) is correct: `'all' !== false` is `true`, so the subscription info block renders, then `! false` at line 149 routes to the "update via your gateway dashboard" message. Graceful degradation without hiding the subscription from the user.
- The `is_object()` guard is the right check here — handles both `null` and any non-object value `get_gateway_object()` might return.

PR #3707, opened 2026-06-21
Improve HTML validity of frontend forms and pages

kimcoleman:
Cleans up a batch of HTML validity errors across the frontend templates, found by running the rendered pages through the W3C Nu validator. No visual or behavioral changes intended.

## Changes

**Form-group structure**
- Checkout account/billing/payment sections keep `<fieldset>` for grouping but use `aria-labelledby` to a card-title `<h2>` instead of a `<legend>` buried inside the card. The same pattern is applied to field groups (`PMPro_Field_Group::display()`) and the Stripe payment fields. The Check gateway section (no form controls, just instructions) becomes a `<div>` + `<h2>`.
- Grouped field types (`radio`, `checkbox_grouped`) use `role="group"` + `aria-labelledby` + a `<span>` label instead of a `<label for>` that can't target a single control.

**Attributes / markup**
- `select2` fields use `data-placeholder` instead of the invalid `placeholder` attribute on `<select>`.
- Removed the JS that injected an `href` onto the discount-code toggle `<button>`.
- Dropped empty `action=""` on the checkout, profile, change-password, and avatar forms (omitted = submit to current URL).
- Split billing address inputs use `autocomplete="billing address-line1"` / `address-line2` instead of the single-field `street-address` token.
- Member action links (Renew / Update Billing / Change / Cancel) include the level/membership id in their `id` so they're unique per card instead of duplicating across multiple levels.

## Validation

Ran the W3C Nu validator against: Checkout (multiple gateways, logged in/out, custom fields, TOS), Profile Edit (all three sections), Update Billing, Account (multi-level), Orders/Invoice, Avatar, Change Password, Cancel. Login/Lost Password and the Levels page were out of scope (core WP / not reviewed here).


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3707 — Improve HTML validity of frontend forms and pages**
*kimcoleman → dev | 12 files, +53 -53*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3707>


**Summary**

Solid HTML validity cleanup — the `<legend>` inside nested divs fix alone justifies this PR since that's been invalid since forever. The ARIA pattern choices are correct and the scope is well-contained. One real issue to address before merging: orphaned CSS in the variation stylesheets that this PR should clean up.


**Issues**

- **Major** `css/frontend/variation_1.css:256-269`, `css/frontend/variation_high_contrast.css:256-268` — `.pmpro_form_legend`, `.pmpro_form_heading`, and `.pmpro_card_content .pmpro_form_heading` rules are now dead code. The elements they target have been removed from every template in this PR. The PR adds styles that activate `.pmpro_card_title` (the `pmpro_card_title + pmpro_card_content` adjacent sibling selector now fires), so that side is correct — but the old rules should be deleted here. Leaving them is harmless functionally but misleading and inconsistent with the intent.

- **Minor** `includes/terms-of-service.php:201` — Converting `<fieldset id="pmpro_tos_fields">` to a `<div>` when the section contains a live form control (the TOS checkbox) is different from the Check gateway conversion (which has zero form controls). The Nu validator flags `<fieldset>` without `<legend>` as a warning, but the fix here changes semantics rather than adding a legend or `aria-labelledby`. If the validator warning was the driver, adding `aria-labelledby` to the fieldset and a visually-hidden heading (or matching the pattern used elsewhere) would preserve the grouping semantics. Worth confirming this was intentional.

- **Minor** `shortcodes/pmpro_account.php:354` — The update-billing link ID is `pmpro_actionlink-update-billing-{level_id}`, but the link targets `pmpro_subscription_id`. If a member somehow has multiple subscriptions for the same level, these IDs would collide. Subscription ID would be more semantically accurate and future-proof here: `pmpro_actionlink-update-billing-{subscription_id}`.

- **Minor** (for documentation) — The rename from `pmpro_form_heading` / `pmpro_form_legend` to `pmpro_card_title` is a breaking change for any site with custom CSS targeting those classes. Worth a changelog note so upgraders can update their custom styles.


**Looks Good**

- `aria-labelledby` on `<fieldset>` pointing to an `<h2>` sibling of `pmpro_card_content` is exactly right — the `<legend>` was buried inside `.pmpro_card_content > .pmpro_card` which is invalid; the new structure activates the pre-existing `pmpro_card_title + pmpro_card_content` adjacent sibling selector in the variation CSS correctly.
- `role="group"` + `aria-labelledby` + `<span>` for radio/checkbox_grouped fields is the correct ARIA pattern — a `<label for>` can't target a set of inputs.
- `data-placeholder` on `<select>` for select2 is correct; `placeholder` isn't a valid attribute on `<select>`.
- `autocomplete="billing address-line1/address-line2"` is spec-correct for split fields; `street-address` is for a single multiline input.
- Dropping `jQuery('#other_discount_code_toggle').attr('href', 'javascript:void(0)')` is the right call — buttons don't have `href`.
- Action link ID uniqueness fix (`-{level_id}` suffix) is a real HTML validity bug fixed correctly.
- `<fieldset>` → `<div>` for the Check gateway section (instructions only, zero form controls) is semantically correct.

PR #3708, opened 2026-06-21
Style the checkout button shortcode and add a style attribute

kimcoleman:
## Summary

The `[pmpro_checkout_button]` shortcode was missed during the 3.1 styling pass. Its output (`<span><a class="pmpro_btn">`) had no `.pmpro` ancestor, but every button rule in the variation CSS is nested inside `.pmpro { … }` — so the link rendered completely unstyled, unlike the equivalent Gutenberg block (which uses core `wp-block-button` classes).

## Changes

- **Scope the output under `.pmpro`** (added to the wrapping `<span>`) so the link picks up our button/link styles and respects the active style variation. A `<span>` is used instead of a `<div>` so the shortcode stays inline-friendly inside paragraphs.
- **New `style` attribute** (`link` | `button`), defaulting to `link`. This keeps existing/new uses rendering as a plain link rather than unexpectedly becoming a full button on update. Opt into the button with `style="button"`.
- **`.pmpro_link` rule** added to `variation_1.css` and `variation_high_contrast.css`, modeled on `.pmpro_btn-print`. It uses the accent color custom properties, so each variation renders it in its own palette.
- **Backward compatibility preserved**: a custom `class` attribute still *fully overrides* the base class (pre-3.x behavior), so `[pmpro_button class="my-btn"]` produces byte-identical output to before.

## Rendering

| Usage | Class on `<a>` | Result |
|---|---|---|
| `[pmpro_checkout_button level="4"]` | `pmpro_link` | accent-colored text link (default) |
| `[pmpro_checkout_button level="4" style="button"]` | `pmpro_btn` | full button, matches the block |
| `[pmpro_checkout_button level="4" class="my-btn"]` | `my-btn` | full override, `style` ignored |

## Test plan

- [ ] Add `[pmpro_checkout_button level="X"]` to a page → renders as a styled text link
- [ ] Add `style="button"` → renders as a full button matching the block
- [ ] Add `class="..."` → only that class is output (no `pmpro_btn`/`pmpro_link`)
- [ ] Switch style variation under Memberships → Settings → Design → link/button colors follow the variation

### DIFF
--- a/css/frontend/variation_1.css
+++ b/css/frontend/variation_1.css
@@ -233,6 +233,19 @@
 	}
 
 	/**
+	 * Text Links
+	 */
+	.pmpro_link {
+		color: var(--pmpro--color--accent);
+		text-decoration: underline;
+	}
+
+	.pmpro_link:hover,
+	.pmpro_link:focus {
+		color: var(--pmpro--color--accent--variation);
+	}
+
+	/**
 	 * Form Styles
 	 */
 	.pmpro_form {
--- a/css/frontend/variation_high_contrast.css
+++ b/css/frontend/variation_high_contrast.css
@@ -233,6 +233,19 @@
 	}
 
 	/**
+	 * Text Links
+	 */
+	.pmpro_link {
+		color: var(--pmpro--color--accent);
+		text-decoration: underline;
+	}
+
+	.pmpro_link:hover,
+	.pmpro_link:focus {
+		color: var(--pmpro--color--accent--variation);
+	}
+
+	/**
 	 * Form Styles
 	 */
 	.pmpro_form {
--- a/shortcodes/checkout_button.php
+++ b/shortcodes/checkout_button.php
@@ -8,20 +8,29 @@ function pmpro_checkout_button_shortcode($atts, $content=null, $code="")
 	// $content ::= text within enclosing form of shortcode element
 	// $code    ::= the shortcode found, when == callback name
 	// examples: [pmpro_checkout_button level="3"]
+	//           [pmpro_checkout_button level="3" style="button"]
 
 	extract(shortcode_atts(array(
 		'level' => NULL,
 		'text' => NULL,
-		'class' => NULL
+		'class' => NULL,
+		'style' => 'link', // 'link' (default) or 'button'.
 	), $atts));
 
+	// A custom "class" overrides the base class. Otherwise use the "style" attribute.
+	if ( ! empty( $class ) ) {
+		$classes = $class;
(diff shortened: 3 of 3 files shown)


flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3708 — Style the checkout button shortcode and add a style attribute**
*kimcoleman → dev | 3 files, +40 -5*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3708>


**Summary**

This is clean and ready to merge. The `.pmpro` wrapper approach is the right fix for the scoping gap, the `pmpro_element_class` filter is correctly preserved (element identifier stays `span_pmpro_checkout_button`), and the `$class` override path is unchanged from pre-3.1 behavior. Security is fine — `pmpro_getCheckoutButton` runs `esc_attr($classes)` regardless of source.


**Issues**

- **Minor** `shortcodes/checkout_button.php:14` — `style` as a shortcode attribute name shadows the HTML `style` attribute in authors' mental models. A user who writes `[pmpro_checkout_button level="4" style="color: red;"]` intending inline CSS will silently get `pmpro_link` applied instead (since `'color: red;' !== 'button'`). Not a bug, but consider sanitizing to `in_array($style, ['link', 'button'], true) ? $style : 'link'` so unrecognized values fall back predictably, and document the accepted values in the comment block. The example comment on line 11 only shows `style="button"` — add `// Accepted values: 'link', 'button'.` inline near the default.

- **Minor** `css/frontend/variation_1.css` and `variation_high_contrast.css` — `.pmpro_link` has no rule in `base.css`, so the styled link only appears when a variation is active. This matches the `pmpro_btn-print` model (same files, same scoping), so it's internally consistent. But the PR description doesn't mention it, and a future maintainer adding a new variation will need to remember to include this class. A comment like `/* Also defined in variation_high_contrast.css */` near the rule, or a note in the PR description, would help.


**Looks Good**

The `pmpro_get_element_class( 'pmpro span_pmpro_checkout_button', 'span_pmpro_checkout_button' )` call is exactly right — space-separated first arg gets exploded correctly by the function, `pmpro` ends up in the class list, and the `$element` identifier is preserved so existing `pmpro_element_class` filter hooks on `span_pmpro_checkout_button` still fire without change.

The three-branch logic in `$classes` assignment (custom class → full override, `style="button"` → `pmpro_btn`, default → `pmpro_link`) correctly keeps `pmpro_getCheckoutButton`'s internal default (`pmpro_btn`) out of the shortcode path without touching the function itself.

PR #3711, opened 2026-06-25
Improve member export CSV for filtering requests.

### LINKED ISSUE #3710: Members List When Exporting Searched Results Adds current user as first record
**Describe the bug**
When exporting the members list, while searching for members - the first record is always the current user (admin) and replaces the actual data. If there is only one result, it exports it incorrectly.

**To Reproduce**
1. Go to Members List
2. Search for a specific member email or username to filter the list. Make sure the search isn't that of your own username.
3. Export and download the CSV
4. See that the first record is that of your own and the actual first record is replaced.

andrewlimaza:
* BUG FIX: Fixes an issue where filtering the members list first record would be incorrect.

Resolves: https://github.com/strangerstudios/paid-memberships-pro/issues/3710

### DIFF
--- a/classes/class-pmpro-exports.php
+++ b/classes/class-pmpro-exports.php
@@ -1045,10 +1045,10 @@ class PMPro_Exports {
 		$query               = "SELECT COUNT(DISTINCT u.ID) FROM {$wpdb->users} u ";
 		$needs_usermeta_join = false;
 		$search              = $this->build_members_search_sql_fragment( $filters, $needs_usermeta_join );
+		$query              .= " INNER JOIN {$wpdb->pmpro_memberships_users} mu ON u.ID = mu.user_id ";
 		if ( $needs_usermeta_join ) {
 			$query .= " LEFT JOIN {$wpdb->usermeta} um ON u.ID = um.user_id ";
 		}
-		$query .= " LEFT JOIN {$wpdb->pmpro_memberships_users} mu ON u.ID = mu.user_id ";
 		$query .= ' WHERE mu.membership_id > 0 ';
 
 		$filter = $this->build_members_filter_sql_fragment( $filters );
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3711 — Improve member export CSV for filtering requests.**
*andrewlimaza → dev | 1 file, +6 -3 lines*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3711>


**Summary**

Fixes a real bug: when searching members for export, an admin's own `pmpro_export_*` usermeta (which stores the export job JSON, including the search term) was causing the admin to match the search and appear in results. The fix is correct and targeted. One minor issue with comment verbosity; otherwise good to merge.


**Issues**

- **Minor** `classes/class-pmpro-exports.php:1344–1347` — The 3-line comment violates the project's one-line comment rule. Design rationale belongs in the commit message, not source. Condense to one line: `// Exclude pmpro_export_* keys so the exporting admin's job record doesn't match the search term.`


**Looks Good**

- The `pmpro\\_export%` escape is correct: `\\` in a PHP double-quoted string produces a literal `\`, which MySQL LIKE interprets as escaping the `_` wildcard — so it matches `pmpro_export*` exactly as intended.
- The INNER JOIN change in `members_count_total` and `members_fetch_ids_chunk` is semantically equivalent to the old LEFT JOIN (the `WHERE mu.membership_id > 0` and status filter from `build_members_filter_sql_fragment` would exclude non-members either way), but is cleaner and filters earlier.
- Moving the `mu` INNER JOIN before the conditional `um` LEFT JOIN is the right ordering — INNER JOINs before outer JOINs improves clarity and lets MySQL prune earlier.
- The search fragment fix propagates automatically to the row data query in `write_csv_rows` (line 1178 calls the same function), so no parallel fix needed there.
- `COUNT(DISTINCT u.ID)` and `SELECT DISTINCT u.ID` correctly handle users with multiple membership rows under the INNER JOIN.

PR #3712, opened 2026-06-25
Fix email log details modal error on Plain permalinks

dwanjuki:
### All Submissions:

The email log modal hard-coded the REST query string with "?log_id=", which produced a double "?" on sites using Plain permalinks (where pmpro.rest_url ends in "?rest_route=/"). WordPress then folded log_id into the rest_route value, matching no route and returning a 404.

Pass log_id via jQuery's data object instead so the separator is chosen correctly under both pretty and plain permalink structures, matching the existing quick_search handler.

1. Set **Settings → Permalinks** to **Plain** and save.
2. Ensure email logging is enabled and at least one email has been logged.
3. Go to the email log report.
4. Click the **View Content** link on a log entry.
5. **Before the fix:** the modal shows "Error loading email log details." and the `email_log_popup` REST request returns a 404.
    **After the fix:** the modal opens and displays the email details, and the REST request returns a 200.
6. **Regression check:** switch **Settings → Permalinks** to **Post name**, save, and repeat steps 3–4. The modal should still open correctly (request returns 200).

### Other information:

### Changelog entry

> BUG FIX: Fixed the View Email Log details modal returning an error on sites using Plain permalinks.

### DIFF
--- a/js/pmpro-admin.js
+++ b/js/pmpro-admin.js
@@ -1743,8 +1743,11 @@ jQuery(document).ready(function($) {
 
 		// Use WordPress REST API
 		$.ajax({
-			url: pmpro.rest_url + 'pmpro/v1/email_log_popup?log_id=' + logId,
+			url: pmpro.rest_url + 'pmpro/v1/email_log_popup',
 			type: 'GET',
+			data: {
+				log_id: logId
+			},
 			beforeSend: function(xhr) {
 				xhr.setRequestHeader('X-WP-Nonce', pmpro.nonce);
 			},


flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3712 — Fix email log details modal error on Plain permalinks**
*dwanjuki → dev | 1 file, +4 -1*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3712>

**Summary**
Correct fix, ready to merge. The root cause (double `?` when `pmpro.rest_url` already ends with `?rest_route=/` on Plain permalink sites) is real and the solution is the right one — moving `log_id` into jQuery's `data` object so jQuery owns the separator logic.

**Looks Good**
- `js/pmpro-admin.js:1884-1889` — the `quick_search` handler already uses exactly this `data:` pattern, so the fix brings `email_log_popup` in line with the existing codebase convention rather than introducing something new.
- `includes/rest-api.php:390` registers `log_id` as the expected query param and `intval`s it server-side (`rest-api.php:1322`), so the key name in the JS `data` object is correct.
- The `manage_options` auth check on the endpoint (`rest-api.php:1824`) and the nonce header are both untouched by this change.

PR #3714, opened 2026-06-29
Show pending memberships more clearly on the account page

kimcoleman:
## What

Surfaces memberships that are waiting on payment on the **Membership Account** page so members can see the status of an order that hasn't completed yet, such as a Pay by Check order or a failed recurring payment.

- **Held level with a past-due payment:** if the latest order on one of the level's active subscriptions is `pending` and dated in the past, the level card shows "Your latest payment for this membership is past due. We are waiting for your payment to be completed." The member can use the existing **Update Billing Info** link, or a **View Order** link when billing updates aren't available for the gateway.
- **Pending checkout for a level the member doesn't hold yet:** a new `pmpro_card pmpro_card-pending` card is shown for that level with "We are waiting for your payment to be completed. Your membership will be activated once the payment has been confirmed." and a **View Order** link.

The change is confined to `shortcodes/pmpro_account.php`. No new helper functions.

## When a pending order counts as live

Held levels: the level's active subscriptions are checked for a latest order that is `pending`, belongs to the current user, and is not dated in the future. Anything else is ignored.

Levels the member doesn't hold: every `pending` order for the user is scanned once per level. A level gets a card only when all of the following are true.

1. The newest `pending`, `success`, or `refunded` order for the level is the pending one. A newer completed order, such as paying again by a different method, means the pending order is stale.
2. The order is not dated in the future. Renewal invoices created ahead of their due date are not yet live.
3. The order has `checkout_request_vars` order meta, meaning it came from a real delayed checkout and will activate the membership when paid. Failed renewal payments for a membership that has since ended do not qualify.
4. In a single-selection level group, the member hasn't started a different level in that group *after* the order was created. Holding another plan in a pick-one group means they chose it instead.

## New hook

`pmpro_membership_account_after_pending_level_card_content` fires inside the pending card with the level object and the pending `MemberOrder`. Add Ons such as Pay by Check can use it to show payment instructions. Note the level object comes from `pmpro_getLevel()`, so it has no membership properties like `startdate`.

## Known limitation

A delayed checkout that is simply abandoned will keep showing its card until the order changes status or the member buys something else. There is no data signal that distinguishes "abandoned" from "check is in the mail," so any age-out is a policy decision.

## Related

Pay by Check currently prints its own "pending order" bullet on the account page with no staleness logic. A companion PBC update should ship with this so the two messages don't contradict each other.

## Proposed changelog entries

- ENHANCEMENT: Membership Account page now shows when a membership's latest payment is past due and shows a card for memberships with a pending checkout that is waiting on payment.
- ENHANCEMENT: Added the `pmpro_membership_account_after_pending_level_card_content` action for adding content to a pending membership card on the Membership Account page.


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3714 — Show pending memberships more clearly on the account page**
*kimcoleman → dev | 2 files, +212 -3 lines*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3714>

**Summary**

Clean, well-reasoned feature. The staleness logic in `pmpro_get_pending_order_for_user_level()` is solid — "most recent order for this level must be pending, and no sibling group level was started after it" — and the static cache is correctly keyed by `user_id:level_id`. The shortcode restructuring from `if/else` to two independent blocks is the right move. Two minor concerns below; nothing blocking.

**Issues**

**Minor** `includes/functions.php` ~line 5506 — `$held_level->startdate >= $recent_order->timestamp` compares two values with different timezone provenance. `startdate` comes from `UNIX_TIMESTAMP(CONVERT_TZ(startdate, '+00:00', @@global.time_zone))` — the MySQL global timezone determines what UNIX_TIMESTAMP sees. `$recent_order->timestamp` is `strtotime( $dbobj->timestamp )` in the MemberOrder constructor — PHP interprets the raw MySQL DATETIME using the PHP default timezone. If `@@global.time_zone` and `date_default_timezone_set()` differ (not uncommon on shared hosts), the comparison can be off by hours. A level started 30 seconds after a pending order was created could be incorrectly treated as not superseding it. Low probability on a well-configured install, but worth a docblock note: `// Both timestamps are Unix; assumes MySQL @@global.time_zone matches the PHP default timezone.`

**Minor** `pmpro_membership_account_after_level_card_content` and `pmpro_member_action_links` now also fire for pending-only level cards. Core handled this correctly — `pmpro_display_member_account_level_message()` got the `pmpro_hasMembershipLevel()` guard. But any third-party code hooked to these actions that assumes the user holds that level will silently output wrong content on pending-only cards. Call this out explicitly in the changelog as a hook behavior change so Add On / custom-code authors know to add their own guard.

**Questions**

If PMPro deactivates a membership after a failed recurring payment (level moves to `cancelled`, user drops out of `$mylevels`), the user gets a pending-only card via `$pending_order_levels`. `pmpro_hasMembershipLevel()` returns false, so the message reads "We are waiting for your payment to be completed. Your membership will be activated once the payment has been confirmed." That reads like a first-time purchase, not a lapsed renewal. Is this scenario reachable in your supported gateway configurations (i.e., does any gateway immediately cancel rather than holding the level active while retrying)? If yes, the copy distinction matters.

**Looks Good**

The two-condition staleness check is exactly right — stale-if-newer-order-exists catches the "paid again by another method" case, stale-if-group-sibling-started-after catches the "switched plans" case. The invoice link being suppressed when `update-billing` is already present avoids a confusing duplicate path. All output is properly escaped (`esc_html`, `esc_url`, `esc_attr`, `wp_kses`). Property access on MemberOrder objects via `__get` correctly surfaces `status`, `timestamp`, and `membership_id` despite their private declarations.

flintfromthebasement (commented):
**Follow-up review**

**PR: strangerstudios/paid-memberships-pro#3714 — Show pending memberships more clearly on the account page**
*kimcoleman → v3.9 | 2 files, +189 -3 lines (10 commits)*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3714>

**Summary**

The "diff since last review" text supplied to this review was mostly base-branch drift (dev's changelog catch-up through 3.8.5, the new `SECURITY.md`, and unrelated settings-page refactors) — I confirmed via `gh pr diff 3714` that the PR itself still only touches `includes/functions.php` and `shortcodes/pmpro_account.php`. The feature did meaningfully change since my last pass: it's grown a new capability — a "past due" banner + View Order link for levels the user *already holds* when their latest subscription payment is pending, distinct from the pending-only card for levels not yet held. One of my two prior Minor issues is resolved (and resolved well); the other still stands. No new blocking issues. Ready to merge as-is if the team is fine leaving the timezone edge case as a follow-up doc note.

**Prior issues — status**

- **Resolved** — hook-behavior concern (`includes/functions.php` / `shortcodes/pmpro_account.php`, pending-only cards firing member-level hooks meant for full members). The author didn't just guard the existing hooks — they split the pending-only card onto its own new action, `pmpro_membership_account_after_pending_level_card_content`, with a docblock that explicitly warns the level object here comes from `pmpro_getLevel()` and has no `startdate`/`enddate`. That's a cleaner fix than what I asked for; no third-party code can mistake this hook for the full-membership one.
- **Not addressed** `includes/functions.php:5599` (new `pmpro_get_pending_order_for_user_level`) — the staleness comparison is now `$held_level->startdate >= $recent_order->getTimestamp()`. I checked `MemberOrder::getTimestamp()` (`classes/class.memberorder.php:1361`): with `$gmt=false` it takes the PHP-interpreted raw timestamp, reformats it as a local datetime string, then re-interprets that string as GMT and converts using the **WP site timezone offset**. Meanwhile `$held_level->startdate` is still produced via MySQL's `CONVERT_TZ(..., '+00:00', @@global.time_zone)`. That's still two independently-configurable timezone settings (WP site timezone vs. MySQL `@@global.time_zone`) feeding one comparison — the specific pipeline changed, but the underlying mismatch risk didn't. Since `getTimestamp()` is pre-existing PMPro code reused here (not something this PR should refactor), a one-line comment on the comparison noting the assumption (`// Assumes WP's configured timezone matches MySQL's @@global.time_zone`) is still the right-sized fix, not a blocker.

**New since last review**

The "user already has the level, but the latest subscription payment is pending and past due" branch (`includes/functions.php:5540-5556`) directly answers the Question I raised last time about lapsed-renewal copy: it now shows "Your latest payment for this membership is past due. We are waiting for your payment to be completed." — distinct wording from the pending-only card's first-purchase-sounding message, plus its own View Order link (suppressed when an `update-billing` link is already present, consistent with the existing pattern). Good handling of the case I flagged.

The pending-only level object is now `clone`d before being handed to hooks (`shortcodes/pmpro_account.php`, `$pending_order_levels[...] = clone $pending_order_level`), preventing hooked code from mutating the cached `$pmpro_levels` global — a nice defensive addition I hadn't asked for.

The card-rendering restructure (`empty($mylevels) && empty($pending_order_levels)` for the empty state, then independent `$mylevels` and `$pending_order_levels` blocks) correctly avoids ever double-rendering a level the user already holds as a second "pending" card — verified the dedupe guard (`in_array($pending_order->membership_id, $mylevel_ids)`).

**Looks Good**

Everything from the prior review still holds (escaping, invoice-link suppression, `__get` property access). The new `pmpro_membership_account_after_pending_level_card_content` hook is genuinely better API design than a guard on the existing one would have been.

dparker1005:
I've pushed a number of changes to this PR since the original review. Summary of where it landed:

**Detection logic was rebuilt.** The original "newest order for the level is pending" check surfaced messaging in cases where no payment was actually expected — orphaned pending orders from failed renewal dunning (permanent "will be activated" cards for churned members), Pay by Check renewal invoices that are future-dated by design ("past due" a week early), and pending orders superseded by paying again through another method. Now:

- **Held levels:** "past due" only shows when the *latest* order on an *active* subscription for the level is pending and past its date.
- **Unheld levels:** the pending card only shows for a real delayed checkout — the newest pending/completed order for the level must be pending, past-dated, and have `checkout_request_vars` order meta — so completing the payment would actually activate the membership. Newer completed payments and single-selection-group switches supersede.
- Future-dated orders are always ignored.

**Pending cards got their own hook.** They no longer fire `pmpro_membership_account_after_level_card_content` / `pmpro_member_action_links`, since existing callbacks can assume the user holds the level. New action: `pmpro_membership_account_after_pending_level_card_content( $level, $pending_order )`. The View Order links are built in the template, link IDs are level-suffixed to match the recent duplicate-ID fix, and pending cards have a `pmpro_card-pending` class for styling.

**The PR now touches only `shortcodes/pmpro_account.php`** (+154/−3). The helper function from earlier revisions was dissolved into the template since each path had exactly one consumer — this also let the past-due check reuse the card's existing subscriptions query.

Also rebased scope: base is now `v3.9` (merged cleanly). Tested end-to-end in a browser across pending/past-due/superseded/future-dated/logged-out scenarios plus several static review passes.

**Still open:** Pay by Check shows its own pending-payment bullet on this page (no staleness logic), so this should ship alongside a PBC release that removes it — I'll open that PR separately. Copy on the two messages is open to review.

ideadude:
Just chiming in to say the updated copy is great.

andrewlimaza (approved):
This is a small but great improvement.

PR #3715, opened 2026-06-30
Resolve orphaned incomplete orders superseded by a newer checkout

kimcoleman:
## Problem

A member can be left with a stale `pending` order that never resolves:

1. Member checks out with **Pay by Check** on a site that filters `pmpro_check_status_after_checkout` to `pending` (withholding access until the check arrives). This creates a `pending` order.
2. Member changes their mind and checks out again with **Stripe**, which succeeds and creates a newer `success` order.
3. The original check order **stays `pending` forever** — the account page shows "Your latest payment for this membership is past due..." even though they have paid.

## Root cause

PMPro already cleans up incomplete orders: `PMPro_Subscription::save()` marks a subscription's `token`/`pending`/`review` orders as `error` when that subscription is cancelled, and a new checkout for the same level cancels the prior subscription.

**But the cleanup keys off a subscription, and the orphaned pending order has none.** A `pending` order receives a `subscription_transaction_id` at checkout, but the `PMPro_Subscription` row is only persisted when the membership is activated. While the order is `pending`, no subscription record exists — so there is nothing to cancel, and the existing cleanup never reaches the orphaned order.

## Fix

On `pmpro_after_checkout`, mark the member's older incomplete (`token`/`pending`/`review`) orders as `error` when a newer order supersedes them.

- **Same level:** re-checking out for the same level resolves the prior incomplete order.
- **Level groups:** switching to another level in a group that only allows one level at a time deactivates sibling levels, so their incomplete orders are superseded too. The fix expands the level set to the whole group in that case, mirroring the group-deactivation logic in `pmpro_changeMembershipLevel()`. Multi-selection groups and ungrouped levels stay same-level-only, so a member who can legitimately hold sibling levels keeps those pending orders.

Guards:
- Only orders **older** than the newly completed order are touched (`id` comparison).
- Orders tied to a **still-active subscription** are left alone, so legitimate past-due renewals are untouched.

An order note is added explaining the status change, mirroring the existing cleanup.

## Testing notes

- Reproduce: create a `pending` check order for a level, then complete a successful Stripe checkout for the same level (or another level in a one-at-a-time group). The older `pending` order should flip to `error` with a note.
- Confirm legitimate past-due renewals (older `success` + newer `pending`) are untouched.

## Not covered here

This resolves the orphan on the member's *next* checkout. A `pending` order that is simply abandoned (never superseded) is out of scope — a cron sweep could be added separately to catch those.

### DIFF
--- a/includes/checkout.php
+++ b/includes/checkout.php
@@ -318,6 +318,54 @@ function pmpro_pull_checkout_data_from_order( $order ) {
 }
 
 /**
+ * Mark a member's older incomplete orders as error when a newer order supersedes them.
+ *
+ * Covers re-checking out for the same level, or switching to another level in a group that only
+ * allows one level at a time. Incomplete orders tied to a still-active subscription are left alone.
+ *
+ * @since TBD
+ *
+ * @param int         $user_id The ID of the user who completed checkout.
+ * @param MemberOrder $order   The newly completed order.
+ */
+function pmpro_error_orphaned_incomplete_orders( $user_id, $order ) {
+	if ( empty( $order->id ) || 'success' !== $order->status || empty( $order->membership_id ) ) {
+		return;
+	}
+
+	// Switching to a level in a one-at-a-time group deactivates sibling levels, so their orders are superseded too.
+	$superseded_level_ids = array( $order->membership_id );
+	$level_group_id       = pmpro_get_group_id_for_level( $order->membership_id );
+	$level_group          = empty( $level_group_id ) ? false : pmpro_get_level_group( $level_group_id );
+	if ( ! empty( $level_group ) && empty( $level_group->allow_multiple_selections ) ) {
+		$superseded_level_ids = pmpro_get_level_ids_for_group( $level_group->id );
+	}
+
+	$incomplete_orders = MemberOrder::get_orders( array(
+		'user_id'             => $user_id,
+		'membership_level_id' => $superseded_level_ids,
+		'status'              => array( 'token', 'pending', 'review' ),
+	) );
+
+	foreach ( $incomplete_orders as $incomplete_order ) {
+		// Only resolve orders older than the one we just completed.
+		if ( (int) $incomplete_order->id >= (int) $order->id ) {
+			continue;
+		}
+
+		$subscription = $incomplete_order->get_subscription();
+		if ( ! empty( $subscription ) && 'active' === $subscription->get_status() ) {
+			continue;
+		}
+
+		$incomplete_order->add_order_note( __( 'Order marked as error because it was superseded by a newer successful order.', 'paid-memberships-pro' ) );
+		$incomplete_order->status = 'error';
+		$incomplete_order->saveOrder();
+	}
+}
+add_action( 'pmpro_after_checkout', 'pmpro_error_orphaned_incomplete_orders', 10, 2 );
+
+/**
  * Legacy function.
  *
  * @since 2.12.3


flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3715 — Resolve orphaned incomplete orders superseded by a newer checkout**
*kimcoleman → dev | 1 file, +48 -0 lines*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3715>

---

**Summary**

Solid fix. The root cause analysis is correct and the implementation is tight. `pmpro_error_orphaned_incomplete_orders()` correctly fires on `pmpro_after_checkout`, scopes cleanup to the right level set (same level or the whole group for one-at-a-time groups), and guards against touching renewal orders that legitimately have an active subscription. Ready to merge.

---

**Issues**

- **Minor** `includes/checkout.php` — `MemberOrder::get_orders()` defaults to a 100-row limit when no `limit` arg is passed. The call at the new function doesn't override it. For the described scenario (one or a few stale pending orders) this is a non-issue in practice, but passing `'limit' => 0` would remove the implicit cap and be more correct. From `class.memberorder.php:577`: `$limit = isset( $args['limit'] ) ? (int) $args['limit'] : 100;` — and a zero value skips the LIMIT clause entirely.

---

**Looks Good**

- The `'success' !== $order->status` early return is the right entry guard; the hook fires from both `checkout.php` and the IPN handlers, so this prevents the cleanup from running on non-final orders regardless of call site.
- The subscription check on the incomplete order is correct. For a `pending` Pay-by-Check order with no subscription record, `get_subscription()` returns null (the order isn't success, so no subscription is materialized). The guard only preserves orders that have a *real, active* subscription row — exactly what "legitimate past-due renewal" means.
- `add_order_note()` mutates `$this->notes` in memory; `saveOrder()` persists both the status change and the updated notes in the same UPDATE query. Sequencing is correct.
- Group logic mirrors `pmpro_changeMembershipLevel()` faithfully: only expands `$superseded_level_ids` to the full group when `allow_multiple_selections` is falsy, leaving multi-selection groups and ungrouped levels as same-level-only.
- The `id` comparison (`>= $order->id`) is the right anchor. An orphaned order is definitionally older; this guards against accidentally touching the newly completed order or any future orders from a race.

dparker1005:
Holding this back from 3.8.3. The intention behind this PR makes sense, but a core-wide sweep changing `pending` orders to `error` also catches Stripe delayed-notification payments (ACH, SEPA, Bacs), which sit at `pending` while the bank settles. Changing these orders to `error` could then silently hide a real payment that was collected if the checkout eventually completes.

I agree that we need a solution around checkouts that are initiated while other checkouts are already in progress, but we may need to consider solutions warning users before they start another checkout or solutions per-gateway since only each gateway knows if it can stop an in-progress `pending` payment.

dparker1005:
I think we should consider closing this PR. The intent is right, but after digging in I don't think a core-wide sweep is the right tool, and the problem it was opened for is now solved elsewhere.

**Why the sweep is unsafe**

Stripe's completion paths only finish an order if its status is still `token` or `pending`. Both `process_checkout_session_completion()` in the webhook handler and `check_token_order()` in the gateway bail out otherwise, on purpose, to avoid double-processing.

That means erroring an in-flight Stripe order loses a real payment. Concretely: a member starts an ACH, SEPA, or Bacs checkout, and the order sits at `pending` for a few days while the bank settles. They get impatient and pay again by card. This PR would flip the bank order to `error`. When `checkout.session.async_payment_succeeded` arrives, the webhook sees a non-pending order, logs "already processed," and exits. The member has been charged twice and PMPro has no record of the second charge. The same thing can happen with a plain card if the member has Stripe's hosted page open in one tab and completes a different checkout in another.

Only the gateway knows whether a `pending` order can still complete, so any cleanup has to be per-gateway rather than in core.

**Why we don't need it anymore**

- #3714 changed how the account page decides whether a pending order is live. A newer `success` order for the level now means the older pending order is never shown, which was the symptom this PR was opened to fix.
- Pay by Check already runs `pmpropbc_cancel_previous_pending_orders` on `pmpro_after_checkout`. It errors older same-level check orders and is safe because it filters on `gateway = 'check'`. That covers the one gateway where we know the pending payment can't still land.

The one piece from here that might still be worth having is extending PBC's cleanup to sibling levels in a one-at-a-time group. If we want that, it belongs in the PBC Add On alongside the account-page changes we already need there for 3.9.

ideadude:
This all makes sense to me @dparker1005. Might be worth asking Kim quick or at least giving her a note in Slack if you haven't yet.

kimcoleman:
I’m fine if we close this and focus on moving forward with the improved view of pending orders work. I still owe David review on that.

This is primarily the case of someone swapping between a check payment and a credit card and vice versa. If you think pay by check can handle that, maybe there’s a tiny bit of work in there to do. No rush.

PR #3719, opened 2026-07-01
Skip archived/inactive Stripe Prices when matching in get_price_for_product()

dparker1005:
## Problem

A customer hit the following error at checkout when using a discount code:

> Could not create checkout session. The price specified is inactive. This field only accepts active prices.

In `get_price_for_product()` (`classes/gateways/class.pmprogateway_stripe.php`), we look up existing Stripe Prices for a product/amount/currency and reuse a match. If the matching Price had been **archived** in the Stripe dashboard, that inactive Price could be handed to Stripe Checkout, which rejects it because Checkout only accepts active prices.

This came up because some Prices on the product had been archived (some dating back to 2022–2024, from when the level had a different name).

## Fix

Per the discussion, the preferred behavior is to **create a new Price** rather than unarchive the old one — the person who archived it likely had a reason, and reporting on the new price won't be muddied by old data.

Two changes to `get_price_for_product()`:

1. Add `'active' => true` to the Price search args. Stripe's List prices endpoint already defaults to active-only, so this is explicit/self-documenting.
2. Add an in-loop guard that skips any Price where `active` is falsy, so an archived Price is never reused. When no active match is found we fall through to the existing "create a new Price" block, which creates a fresh active Price. The archived Price stays archived.

## Testing notes

- `php -l` passes.
- Suggested manual test: archive a Price on a Stripe product that matches a level/discount-code amount, then check out for that amount. Before this change checkout fails with the "inactive price" error; after, a new active Price is created and checkout completes.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -3514,10 +3514,14 @@ class PMProGateway_stripe extends PMProGateway {
 			$tax_behavior = 'no';
 		}
 
+		// Only match active prices. Without this filter, Stripe's list endpoint also returns
+		// archived/inactive prices, which would then be reused and rejected at checkout with
+		// "The price specified is inactive." Excluding them here lets us create a new active Price instead.
 		$price_search_args = array(
 			'product'  => $product_id,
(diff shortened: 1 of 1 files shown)


dparker1005:
Closing in favor of #3717

PR #3721, opened 2026-07-02
Add declarative settings field helpers for admin settings screens

dparker1005:
## What

Adds `includes/settings-fields.php`: render-only helpers (`pmpro_build_settings_section()`, `pmpro_build_settings_fields()`, `pmpro_build_settings_field()`, `pmpro_build_settings_input()`) that generate the repeated section wrapper, form-table, row, and control markup used across PMPro admin screens, and converts most core settings screens to use them.

## Why

Every settings screen currently hand-rolls the same `pmpro_section` / `form-table` / `<tr>` markup, inline show-hide `<script>` blocks, and per-checklist Select All/None scripts. Declaring fields as arrays makes the screens shorter, keeps markup/escaping consistent in one place, and gives us a shared conditional-visibility mechanism.

## How it works

- Fields are declared as arrays (text, number, email, url, tel, password, currency, color, select, radio, textarea, checkbox, checklist, editor, composite, html, callback).
- Rows can declare `depends` conditions. PHP derives the initial visibility server-side (no flash), and a shared `data-pmpro-depends` handler in `pmpro-admin.js` toggles rows live. The attribute also works hand-placed on arbitrary elements.
- Checklist Select All/None is handled by delegated JS instead of per-checklist inline scripts.
- The helpers only render — each screen still loads, sanitizes, and saves its own values, so option storage is unchanged.

## Converted in this PR

- Admin pages: advanced, security, email, design, page, payment settings, license, discount codes, edit level.
- Gateways: Stripe, PayPal Express, Check (other gateways still use the legacy `pmpro_payment_option_fields` compat path and render unchanged).
- Settings fields registered by avatars, reCAPTCHA, Cloudflare Turnstile, and terms of service.
- Bespoke UIs (Stripe Connect, license page, design style variations, discount code per-level pricing) intentionally keep custom markup, using the section open/close helpers or mixing helper-built rows into hand-rolled tables.

## Notes for review

- Input names/ids and save handlers are unchanged; a small number of intentional cleanups are included (fixed malformed tables, dangling `label for`s, duplicated id attribute, a `</textarea>` escaping hole in the Check gateway).
- Some legacy DOM hooks changed (e.g. `#pmpro-hideads` → `#hideads`, checkbox submit values `yes` → `1` on edit level, class-broadcast row toggles replaced by `data-pmpro-depends`) — flagged for discussion.
- The Stripe payment settings field order was adjusted so the flow toggle is followed by the general billing address option, then the on-site group, then the Stripe Checkout group.


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3721 — Add declarative settings field helpers for admin settings screens**
*dparker1005 → dev | 20 files, +2665 -2754*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3721>

**Summary**
This is a large, careful refactor of admin settings markup into a declarative field-array system (`includes/settings-fields.php`). The escaping model is sound (per-context escaping in one place, explicit escape-boundary contract for callables), and I traced the checkbox value change (`yes`→`1`), the ID rename (`#pmpro-hideads`→`#hideads`), and the dropped `gateway_*` row classes on the PayPal/Check gateway rows against the actual save handlers and other markup — none of those break anything (save handlers check truthiness, not exact string equality; the old gateway-class rows are dead markup left over from a single-page gateway UI that no longer exists). One real regression found in Security Settings; otherwise solid.

**Issues**

- **Major** `adminpages/securitysettings.php:250,278,293,408` — Four sections (Restricted Files, HTTPS Settings, DNS Firewall, WordPress Security) are now hardcoded `'open' => true`. On `dev` these were all `data-visibility="hidden" data-activated="false"` (collapsed by default) — confirmed via `grep 'data-visibility="hidden"' adminpages/securitysettings.php` on the pre-PR branch. This is the only settings screen in the diff where this happens (I checked every other converted page/gateway against its pre-PR collapsed/open state; all others already defaulted to open). Every admin who visits Security Settings will now see all four sections expanded instead of the three secondary ones being collapsed. Fix: pass `'open' => false` for these four `pmpro_build_settings_section()` calls to restore the original collapsed state.

- **Minor** `adminpages/advancedsettings.php` (custom advanced settings loop, `pmpro_custom_advanced_settings` filter) — previously, a `$field['field_type']` value outside `select`/`text`/`textarea`/`callback` rendered an empty cell (`default: break;` in the old switch). The new code falls through to `$custom_entry['type'] = 'text';`, so any Add-On registering an unrecognized/typo'd `field_type` now silently renders a bound text input instead of nothing. Low-risk (no Add-On appears to rely on the old silent-no-render behavior), but worth a conscious call — either preserve the old no-render fallback or note the behavior change in the changelog.

**Looks Good**
- `includes/settings-fields.php` — the escaping contract is well thought out (helper-assembled markup escaped centrally, callables are explicit escape boundaries, `phpcs:ignore` comments only where the concatenated string is provably pre-escaped).
- The `check` gateway's `</textarea>` fix (`classes/gateways/class.pmprogateway_check.php`) is correct: swapping `wp_kses_post()` inside a `<textarea>` for the new `textarea` field type's `esc_textarea()` closes a real (if low-severity, admin-only) breakout hole.
- The `depends`/`current` mechanism (PHP-derived initial visibility + shared delegated JS) is a clean replacement for the per-page inline show/hide scripts, and the AND/OR semantics were verified to correctly reproduce the old multi-selector jQuery logic in `edit-level.php` and `discountcodes.php` (including the old class-based `recurring_info`/`trial_info` coupling).

**Questions**
1. Was the Security Settings collapse-state change (see Major above) intentional, or dropped during the conversion?
2. For gateways not converted in this PR (still on `pmpro_payment_option_fields` compat), is there a follow-up PR planned, or is that path staying as-is long-term?

dalemugford:
**Suggestion: derive `depends` initial state from rendered values instead of the `current` key**

Really like where this landed overall. One contract stood out as fragile while reading `settings-fields.php`: the `current` key in `depends`. When a condition targets a checkbox by `value`, the caller has to hand-mirror the JS semantics (input's value while checked, `''` while unchecked). It's documented well, but every future caller has to get that right or the row flashes or starts in the wrong state, and nothing fails loudly when they get it wrong.

The thing is, the render layer already knows every value it needs. `pmpro_build_settings_input()` resolves the final value (explicit `value` or the `get_option( 'pmpro_' . $name )` fallback), and for checkboxes it computes `$is_checked` and `$checkbox_value` right before echoing. `current` makes callers re-derive by hand something the helper computed moments earlier.

Proposal: record rendered state, and let `depends` look it up.

```php
function pmpro_settings_rendered_state( $id = null, $state = null ) {
	static $rendered = array();
	if ( null !== $state ) {
		$rendered[ $id ] = $state;
		return;
	}
	return isset( $rendered[ $id ] ) ? $rendered[ $id ] : null;
}
```

`pmpro_build_settings_input()` records each named control as it renders. The checkbox case is the main win, because the helper stores the JS-mirroring semantics once, in the one place that actually knows them:

```php
// Checkbox case.
pmpro_settings_rendered_state( $name, array(
	'checked' => $is_checked,
	'value'   => $is_checked ? $checkbox_value : '',
) );

// Text/select/textarea case.
pmpro_settings_rendered_state( $name, array( 'value' => (string) $value ) );
```

Then in `pmpro_build_settings_field()`'s depends loop, a condition without an explicit `current` tries the registry before giving up:

```php
if ( ! array_key_exists( 'current', $condition ) ) {
	$state = pmpro_settings_rendered_state( $condition['id'] );
	if ( null !== $state ) {
		$condition['current'] = isset( $condition['checked'] ) ? $state['checked'] : $state['value'];
	}
}
```

The effect on call sites is that `'current' => $payment_flow` and friends just get deleted. As far as I can tell, every `depends` in this PR references a control rendered earlier in the same fields list by these same helpers (Stripe's flow select before its dependent rows, the reCAPTCHA version select before the key fields, `hideads` before its dependent row), so the automatic path would cover all current usage.

Two design notes on why this shape rather than alternatives:

- Resolving from the option (`get_option` at depends time) would break on screens that render an explicit `value` differing from the stored option, e.g. request-precedence values around save handling. The registry captures what was actually rendered, which is what the JS will see.
- `current` stays as the documented escape hatch for targets the registry can't see: hand-rolled inputs like `wp_dropdown_pages()`, radios given manual ids, or a condition referencing something rendered later on the page. The contract flips from "every caller must get this right" to "you only touch `current` when the helpers didn't render the target."

If you'd rather not add the registry, a minimal version that still removes the sharpest edge: a tiny helper so callers write `'current' => pmpro_settings_checkbox_current( $is_checked, $checkbox_value )` instead of hand-encoding the value-while-checked, empty-while-unchecked convention.

dparker1005:
Follow-up on the review feedback, addressed in ff2f82a80 and fa9329e6e:

**@flintfromthebasement — sections defaulting open:** After digging in, those four Security Settings sections were only ever collapsed *on paper*. Nothing reads `data-visibility` on load — collapse is driven by the arrow icon class and an inline `display: none`, and the old markup here had arrow-up icons, `aria-expanded="true"`, and no inline style. So the attributes were stale and admins have always seen all five sections expanded. `'open' => true` preserves actual behavior and fixes the inconsistent attributes as a side effect.

**@flintfromthebasement — unknown `field_type` → text input:** Keeping this deliberately — the old no-render path still ran `pmpro_setOption()` on save with no posted value, blanking the stored option every time. The bound text input preserves it. Now documented with a comment in the adapter loop.

On the unconverted gateways: no changes planned in this PR — they stay on the `pmpro_payment_option_fields` compat path and render unchanged; any conversion would be a separate follow-up.

**@dalemugford — rendered-state registry:** Implemented — thanks, this was the right call. Two tweaks to the sketch: selects record the option the browser will actually show (falling back to the first option when the stored value matches nothing, since that's what the JS reads), and radios are never recorded (they render without ids, so a PHP-derived state would disagree with the JS, which can't target them either). Explicit `current` still wins and stays as the escape hatch for hand-rolled inputs or forward references. All 13 `current` values at call sites are gone.

PR #3724, opened 2026-07-06
Require Authorize.net credentials before processing Silent Post requests

dparker1005:
## Summary

The Authorize.net Silent Post handler (`services/authnet-silent-post.php`) is exposed publicly via `admin-ajax.php?action=authnet_silent_post` (`wp_ajax_nopriv_`). Currently, if the optional Silent Post token is not configured, the handler will accept and process incoming requests even on sites that are not using Authorize.net at all.

This adds a basic guard: if the site has no Authorize.net credentials stored (`pmpro_loginname` / `pmpro_transactionkey`), we bail with a `403` before accepting or processing any Silent Post data.

## Change

```php
// Bail if Authorize.net credentials are not configured. If this site isn't set up to use
// Authorize.net, we should not be accepting or processing Silent Post requests at all.
if ( empty( get_option( 'pmpro_loginname' ) ) || empty( get_option( 'pmpro_transactionkey' ) ) ) {
	status_header( 403 );
	exit;
}
```

## Notes / scope

- This is a credential *presence* check only. It closes the case where a site that isn't configured for Authorize.net still exposes a live, unauthenticated Silent Post endpoint.
- It does **not** authenticate the request itself. Sites that *are* using Authorize.net should still configure the Silent Post token for true request authentication.

### DIFF
--- a/services/authnet-silent-post.php
+++ b/services/authnet-silent-post.php
@@ -14,6 +14,13 @@
 	// Flag if this is an ARB transaction. Set to false by default.
 	$arb = false;
 
+	// Bail if Authorize.net credentials are not configured. If this site isn't set up to use
+	// Authorize.net, we should not be accepting or processing Silent Post requests at all.
+	if ( empty( get_option( 'pmpro_loginname' ) ) || empty( get_option( 'pmpro_transactionkey' ) ) ) {
+		status_header( 403 );
+		exit;
+	}
+
 	// Validate the shared-secret token if one is configured.
 	// Authorize.net sends exactly the Silent Post URL you provide, so a token appended to that URL
 	// will be present on every legitimate request and absent on forgeries.


flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3724 — Require Authorize.net credentials before processing Silent Post requests**
*dparker1005 → dev | 1 file, +7 -0 lines*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3724>

**Summary**
Ready to merge. Adds a guard to `services/authnet-silent-post.php` that bails with a 403 before doing any work if the site has no Authorize.net credentials configured, closing an unauthenticated-endpoint exposure on sites that don't even use Authorize.net. Verified `pmpro_loginname`/`pmpro_transactionkey` are the actual option names the gateway class (`classes/gateways/class.pmprogateway_authorizenet.php`) uses for these credentials, and the guard is correctly placed before the existing token-validation and `$_REQUEST` processing logic.

**Looks Good**
- Placement: the check runs before any request data is read or processed, and before the optional shared-secret token check — correct fail-fast ordering (`services/authnet-silent-post.php:17-22`).
- Scope is honest: PR description explicitly notes this is presence-only, not authentication, and that sites actively using Authorize.net still need the Silent Post token for real request auth. No overclaiming.
- Minimal, single-purpose diff — no unrelated changes.

**Minor**
- `services/authnet-silent-post.php:18` — unlike the token-validation failure path just below it, this new 403 branch doesn't log anything even when `PMPRO_AUTHNET_SILENT_POST_DEBUG` is enabled. Not blocking, but if a legitimate site somehow hits this (e.g. credentials wiped/misconfigured mid-migration), there's no debug trail to explain the sudden 403s. Optional: add the same log-on-`PMPRO_AUTHNET_SILENT_POST_DEBUG` treatment for consistency.

PR #3727, opened 2026-07-10
Add percentage and fixed amount discount code types

dparker1005:
## What this does

Extends discount codes with formula-based discounts alongside the existing per-level price overrides. Each code now has a `discount_type`:

- **`set_price`** — existing behavior, unchanged; remains the default. Existing codes need no migration.
- **`percentage`** — e.g. 20% off the level's regular pricing.
- **`fixed`** — e.g. $10 off the level's regular pricing.

Formula codes store one rule on `pmpro_discount_codes` (`discount_value`, `apply_to_initial`, `apply_to_recurring`) shared across all eligible levels, and inherit everything else — billing cycle, billing limit, trial, expiration — from the level. Sites that need custom structure plus a discount keep using `set_price`.

## How pricing is resolved

New `pmpro_get_discounted_level_for_code( $level, $code )` in `includes/discount-codes.php` is the single source of truth: it confirms the level is eligible via `pmpro_discount_codes_levels`, applies the rule to the level's *current* pricing, rounds with `pmpro_round_price()`, and clamps at $0. All four runtime paths that previously read `pmpro_discount_codes_levels` as final pricing now use it:

- `pmpro_getLevelAtCheckout()`
- the AJAX `applydiscountcode` service (resolved per level *before* the multi-level combined sum)
- `MemberOrder::getMembershipLevel()` (the order fallback used by IPN handlers)
- the 2Checkout INS handler

The `pmpro_discount_code_level` filter still runs at its existing call sites *after* calculation, so documented filter-based extensions keep receiving final prices. A new `pmpro_get_discounted_level_for_code` filter runs inside the resolver.

## Backwards compatibility for direct table readers

`pmpro_discount_codes_levels` still holds one row per eligible level for every code type. For formula codes, its pricing columns hold a **snapshot** of the calculated prices — refreshed on code save and on level save (`pmpro_save_membership_level`) — so add-ons reading the table directly see plausible, usually-exact prices instead of zeros. Checkout never trusts the snapshot; it always recalculates. This also means the existing gateway-compatibility checks evaluate correct cycle/trial values for formula codes with no changes.

## Admin / API

- Edit screen: Discount Type, Discount Amount, and Applies To (initial/recurring) fields; for formula codes the per-level pricing tables collapse to an explanatory note. New Select All / Deselect All buttons for the levels list.
- Codes list table shows the rule (e.g. "20% off initial payment, recurring payments") under the code.
- `PMPro_Discount_Code` and `POST /pmpro/v1/discount_code` accept the new fields.
- Schema via dbDelta, db version bumped to 3.9.

## Also included

- Fixed `pmpro_get_discount_code()` to match code strings before IDs so fully numeric codes (e.g. a code literally named `2025`) resolve correctly at checkout.
- Replaced duplicate raw code-lookup queries in `pmpro_complete_checkout()`, `pmpro_getLevelAtCheckout()`, and the AJAX service with the helper; the resolver now receives the already-fetched code row instead of refetching (once per level in multi-level checkout).

## Testing done (local site, Stripe-less flows)

- Percentage and fixed math on one-time, recurring, and trial-bearing levels; fixed discount clamping ($10 off a $1 level → $0); ineligible levels return null; legacy `set_price` codes resolve identically (including the `level_id` property `MemberOrder` depends on).
- Live checkout page renders discounted price text; AJAX apply-code works single-level and sums correctly for multi-level (`2+3` → $88 for 20% off $10 + $100).
- Admin form round-trips all three types, JS toggles verified, snapshots written correctly and refreshed after a level price change.
- Numeric code `2025` applies its set_price override at checkout after the lookup-order fix.

## Notes for reviewers

- The `merge-subscription-delays-set-expiration-dates` branch touches the same per-level UI region of `adminpages/discountcodes.php`; whichever lands second has a small mechanical conflict.
- Pre-existing (untouched here): `pmpro_check_discount_code_level_for_gateway_compatibility()` queries `pmpro_discount_codes_levels` by an `id` column that doesn't exist, so its check-all branch silently passes — worth a separate issue.
- `@since TBD` placeholders to be filled at release.


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3727 — Add percentage and fixed amount discount code types**
*dparker1005 → v3.9 | 13 files, +581 -38*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3727>

---

**Summary**
Well-designed feature: one resolver, `pmpro_get_discounted_level_for_code()`, now owns pricing for all four runtime paths, and the snapshot-for-back-compat approach is pragmatic. One regression in the 2Checkout INS handler needs fixing before merge. Everything else is minor or a question.

---

**Issues**

- **[Major]** `services/twocheckout-ins.php:295` — `pmpro_get_discounted_level_for_code( $morder->membership_level->level_id, $discount_code_id )` passes a bare integer ID (`$discount_code_id = $morder->discount_code->id`, line 286). The non-object branch of `pmpro_get_discount_code()` (`includes/discount-codes.php:47-56`) matches `WHERE code = %s` first and only falls back to `WHERE id = %d`. This PR itself establishes that codes can be fully numeric (the `2025` fix), and IDs are small sequential integers, so a code literally named `"12"` coexisting with a different code whose `id = 12` resolves the wrong code's pricing for a live 2Checkout order, while the use-count row later in the same handler is still written against the original ID. The pre-PR query here was `WHERE dc.id = ...`, unambiguous, so this is a regression. Every other caller passes a code string or the row object; this is the only raw-ID call site. Fix: pass `$morder->discount_code` (the object already loaded at line 283; the object branch returns it directly). Note also that `getMembershipLevel(true)` on line 285 already resolved the discounted level via the resolver, so this second call may be redundant entirely.

- **[Minor]** `includes/rest-api.php:908-919` — The route accepts `PUT,PATCH` (`rest-api.php:166`) but always builds a fresh `PMPro_Discount_Code` and only sets the new fields when present. Defaults are `set_price`/`0`/`1`/`1` (`class-pmpro-discount-codes.php:45-66`) and `save()` writes all four unconditionally (`class-pmpro-discount-codes.php:303-306`). A PATCH of `{ "code": "SAVE20", "uses": 50 }` converts a percentage code back to `set_price`. The endpoint already clobbers `starts`/`expires`/`uses` the same way, so this is a pre-existing shape rather than new breakage, but the new fields make the clobber turn a working discount into a no-op silently. Cheapest fix: for PUT/PATCH, load the existing code first and only overwrite supplied params.

- **[Minor]** `includes/rest-api.php:909` — `discount_type` is not validated against `pmpro_get_discount_code_types()`; `save()` silently coerces unknown values to `set_price`. Returning a 400 for an unknown type would save API consumers a confusing debug session.

- **[Minor]** `adminpages/discountcodes.php:85-86` / `class-pmpro-discount-codes.php:268-272` — A formula code with both `apply_to_initial = 0` and `apply_to_recurring = 0` saves successfully and discounts nothing. Worth rejecting server-side (and disabling Save in the form) since it is almost certainly user error.

---

**Looks Good**

- The `set_price` branch of the resolver sets `$discounted_level->id = $level->id` explicitly. The legacy `SELECT l.id, cl.*` join let `cl.id` overwrite `l.id` in the row object, which is what the `sigh, I really need to standardize the column names` comment in `class.memberorder.php` was working around. This quietly fixes that wart.
- Input handling is tight: `discount_type` whitelisted in both admin and `save()`, `discount_value` float-cast and clamped, percentage capped at 100, all new queries use `$wpdb->prepare()`. Net security improvement over the `esc_sql()` concatenation it replaces.
- Legacy `pmpro_discount_code_level` filter still runs at its existing call sites after resolution (verified `includes/functions.php:2921`, `services/applydiscountcode.php:82-85`), and the new filter runs inside the resolver first. Correct ordering for existing extensions.
- Snapshot refresh hook signature matches `pmpro_save_membership_level` (`adminpages/levels/save-level.php`). dbDelta defaults leave existing codes untouched.

---

**Questions**

1. `includes/discount-codes.php:143-144` — `trial_amount` is never discounted, even with `apply_to_recurring` set. The description says trial structure is inherited, so this reads as intentional. Confirm that "recurring" is meant to exclude trial payments; the admin label may want to say so.
2. Snapshots refresh only via `pmpro_save_membership_level`, which fires from the admin level form. Programmatic or REST level price changes leave stale snapshots. Given checkout never trusts them, is that acceptable, or a follow-up?

dparker1005:
Hi @flintfromthebasement, thank you for the thorough pass, the 2Checkout catch in particular.

Fixed in 70ad8679a: the INS handler now passes the loaded `$morder->discount_code` object to the resolver instead of the bare ID, so a numeric code can no longer collide with another code's ID. The ID variable stays because the use-count row further down still needs it.

Leaving the rest as-is for this PR:
- REST PUT/PATCH clobbering unspecified fields is the endpoint's existing shape for `starts`/`expires`/`uses` too. Worth fixing, but as its own change to the endpoint rather than folded in here.
- Unknown `discount_type` coercing to `set_price` matches how `save()` handles every other field. I'd rather not have one field 400 while the rest silently coerce.
- A formula code with both apply flags off is harmless (it just discounts nothing), so not adding validation for it.

On the questions: yes, "recurring" intentionally excludes the trial amount, since the trial structure is inherited from the level. Snapshots only refreshing on the admin level save is acceptable because checkout never reads them; they exist only for back compat with code that reads the codes_levels rows directly.

andrewlimaza (approved):

kimcoleman (approved):
Talked myself in and out of concerns around how people think of a "percentage" - whether to enter a whole number like "20" and that translates to 20% off vs. if people think to enter a decimal value like ".2" and expect that to be 20% off. For now, we are going to leave this as David has it coded. If people are properly reading the list table view or testing codes at checkout they will see how the number they entered is being handled.

Merge conflict resolved, tested, and approved.

PR #3728, opened 2026-07-14
Rename Members List to Memberships and add Orders-style filter panel

dparker1005:
### Terminology: Members List → Memberships
Now that a single user can hold multiple levels, each row in this list is a **membership** (user + level), not a member — the query has grouped by `u.ID, mu.membership_id` for a while, so the labels are catching up to the code.

- Admin submenu, admin bar node, header nav tab, and page H1 renamed to **Memberships**.
- List table singular/plural, "Memberships per page" screen option, and "No memberships found." empty state updated.
- Person-oriented language is intentionally kept: **Add New Member**, **Search Members**, **Edit Member**.
- No changes to slugs (`pmpro-memberslist`), capabilities, or hook names — existing links and add-ons are unaffected.

### Filtering: same filter panel as Orders and Subscriptions
Replaces the single overloaded dropdown (which mixed levels with `Cancelled/Expired/Old Members` and could not combine them) with the `pmpro-filter-panel` sidebar used on the Orders and Subscriptions pages:

- **Level** — any level, now combinable with status (e.g. "expired Gold memberships").
- **Status** — Active (default) / Cancelled / Expired / All Ended. "All Ended" uses an explicit status list from the new filterable `pmpro_get_ended_membership_statuses()` (`expired`, `cancelled`, `admin_cancelled`, `changed`, `admin_changed`, `inactive`) rather than `<> 'active'`, so incomplete-checkout statuses (`token`, `pending`, `review`, `error`) can't leak in.
- **Ended Memberships** (shown only for ended statuses) — "Hide if the same level is now active" (default) or "Hide if any level is now active".

### Ended-view semantics
Each row represents the current state of a user's relationship with a level, so ended views **always** hide a row when that same level has been re-activated (a user who expired out of Gold and re-purchased Gold is not an "expired membership"). The old behavior — only show users with **no** active membership at all — is the "any level" option. This fixes the long-standing quirk where a user who downgraded from Gold to Silver never appeared anywhere as an ended Gold membership.

### Shared query semantics for list + export
New `pmpro_sanitize_memberships_list_filters()` and `pmpro_memberships_list_filter_sql()` helpers are used by both the list table and the members CSV export, so exported rows always match the rows on screen.

### Backwards compatibility
- Legacy `l=cancelled`, `l=expired`, and `l=oldmembers` URLs (dashboard, reports, admin activity email, add-ons) are mapped to the new `status`/`excludeactive` params, keeping the original "no active membership at all" gate. One deliberate change: `l=oldmembers` previously used `status <> 'active'`, which also returned incomplete-checkout rows (`token`, `pending`, `review`, `error`). It now maps to `status=ended`, so those rows are no longer included.
- Numeric `l` values behave exactly as before (level + active).
- `pmpro_members_list_sql` and all column/CSV hooks fire unchanged.

### Bug fix along the way
`column_enddate()` keyed off the legacy `l` values, so ended views reached via the new `status` param rendered blank end dates; it now uses the normalized status filter.

## Testing

1. Go to Memberships > Memberships. Confirm the default view shows active memberships and the **Filter Results** button opens the sidebar panel.
2. Create a user with an expired level A and an active level B. Filter Status = Expired: the expired A row **shows** with "Hide if the same level is now active" and **hides** with "Hide if any level is now active".
3. Create a user with an expired level A who re-purchased level A. The expired row is hidden under every ended view.
4. Visit a legacy URL like `admin.php?page=pmpro-memberslist&l=expired` and confirm it matches the pre-change results (only users with no active membership), with the panel reflecting Expired + "any level".
5. Combine a Level with Status = Expired/Cancelled/All Ended and confirm results, sorting by end date, and the End Date column values.
6. Export to CSV from a filtered view and confirm the rows match the screen.


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3728 — Rename Members List to Memberships and add Orders-style filter panel**
*dparker1005 → v3.9 | 6 files, +284 -103*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3728>

---

**Summary**
Good change, ready to merge after one description/behavior reconciliation. The filter-panel scaffolding mirrors the Orders table (reuses `pmproInitFilterSidebar` in `js/pmpro-admin.js` and existing `admin.css` classes, no new assets), the shared `pmpro_sanitize_memberships_list_filters()` / `pmpro_memberships_list_filter_sql()` pair keeps list and CSV export in lockstep, and the SQL is safe (level ID is `(int)`-cast, statuses come from fixed literals or are `esc_sql`'d after the filter). The one real gap is that `l=oldmembers` is not fully behavior-compatible as the description claims.

---

**Issues**

- **Minor** `includes/functions.php:5625-5628` and `:5695-5701` — Legacy `l=oldmembers` maps to `status=ended`, which queries the explicit list from `pmpro_get_ended_membership_statuses()`. The removed code (and the shipped `3.8.6` tag, verified via `git grep` on `classes/class-pmpro-members-list-table.php:446` and `class-pmpro-exports.php:1289`) used `mu.status <> 'active'`, so old links and exports also returned `token`/`pending`/`review`/`error` rows for users with no active membership. Dropping those is an intentional improvement per the description, but the "original behavior preserved" claim for `oldmembers` is not accurate. Either note the deliberate change in the PR description (my preference; abandoned-checkout rows were never "old members"), or add a legacy path that keeps `<> 'active'` for `l=oldmembers` only.

- **Minor** `classes/class-pmpro-members-list-table.php:285` — The "Search all levels" link in `no_items()` now only appears when `$filters['l']` is set. Status-only empty states (e.g. Expired with no level) lose the "try without filters" affordance the old `if ( $l )` provided. Consider `if ( $filters['l'] || 'active' !== $filters['status'] )` with a "Show all memberships" link, or accept the panel's "Clear All" as sufficient.

- **Minor** `includes/functions.php:5691` — `pmpro_get_ended_membership_statuses()` output is `esc_sql`'d but never type-checked before `implode()`. A filter callback returning a non-string element throws a `TypeError` and fatals the list page and export. Cheap hardening: `array_filter( $statuses, 'is_string' )` before the `esc_sql` map.

- **Minor** `classes/class-pmpro-members-list-table.php:147,278,310,742,759` — `pmpro_sanitize_memberships_list_filters( $_REQUEST )` is recomputed in five methods, including `column_enddate()` once per row. Not a correctness issue, but worth caching in a property set in `prepare_items()` if you touch this file again.

---

**Looks Good**
- `pmpro_get_ended_membership_statuses()` with a filter is a real improvement over `<> 'active'`; `changed`/`admin_changed`/`inactive` are legitimate terminal per-level statuses (`includes/functions.php:1200,1222,1387`, `includes/profile.php:609,615`).
- The `samelevel` `NOT EXISTS` (`mu2.membership_id = mu.membership_id`) fixes the downgrade-invisibility quirk cleanly, and `anylevel` reproduces the old any-active exclusion exactly. Legacy mapping forces `excludeactive = 'anylevel'` for `cancelled`/`expired`/`oldmembers`, which is correct for old bookmarks.
- `column_enddate()` keyed off the normalized status fixes the blank end date on `status`-param views.
- `pmpro_members_list_sql` still fires on all three query paths; add-ons are unaffected. Slugs, capabilities, and hook names untouched.

---

**Questions**
1. The `excludeactive` selector is shown for Cancelled and Expired too, not only "All Ended" (`extra_tablenav()`, `'active' === $filters['status'] ? 'style="display:none;"' : ''`). Intentional? The description reads as if it's scoped to ended statuses generally, which matches the code, but confirm the wording.
2. `class-pmpro-exports.php:1283` re-sanitizes filters already normalized by `sanitize_members_filters()`. Is there a call path that bypasses it? If not, the comment is misleading and the call is dead defense.

dparker1005:
Hi @flintfromthebasement, thank you for checking the legacy paths against the shipped tag.

You're right about `l=oldmembers`. The old `<> 'active'` query also swept in token/pending/review/error rows, and the new mapping to `status=ended` intentionally doesn't. I've updated the description to say that plainly instead of claiming the behavior is preserved. Abandoned-checkout rows were never old members, so a legacy-only `<> 'active'` path isn't worth carrying.

Leaving the rest as-is: Clear All in the panel covers the status-only empty state, a filter returning non-strings is a programmer error we don't guard against elsewhere, the repeated filter parsing is cheap, and the export re-sanitize is harmless defense.

The `excludeactive` selector showing for Cancelled and Expired is intentional. Same-level hiding is always on for any ended view, and "any level" is the opt-in for all of them.

ideadude:
Looks good to go!

andrewlimaza (approved):

dparker1005:
**Status: on hold.**

Kim and Jason debriefed this PR on 2026-09-23. Their feedback is here: https://docs.google.com/document/d/1z_Cj5GagmWV3iB9tazESZ5a8VOHJVZd2DCcQVpIZ9Jc/edit?tab=t.0

The direction needs more thought before this merges, in particular whether ended memberships belong in this table or in a dedicated report. I've converted the PR to draft and pulled it from the 3.9 milestone. Nothing from this branch ships in 3.9, since the filter parameters it adds would become a contract we'd have to map again if the design changes.

**Menu rename to PMPro: consider for 4.0.** WordPress derives every submenu screen ID from the top-level menu title, so renaming it changes `memberships_page_pmpro-*` to `pmpro_page_pmpro-*`. Core hardcodes the old IDs in the Reports and User Fields meta boxes, several Add Ons check them, and admins' saved column settings are keyed on them. It can be pinned with one line after `add_menu_page()`, but a change that touches every screen ID is better suited to a major version.

PR #3736, opened 2026-07-21
Add PMPro abilities for WordPress and MCP integrations

mircobabini:
### All Submissions:

- [x] Followed the contributing guidelines.
- [x] Code follows WordPress coding standards.
- [x] Checked for overlapping open pull requests.

Adds a core-native WordPress Abilities API surface based on the production feature implementation from [mircobabini/pmpro-abilities](https://github.com/mircobabini/pmpro-abilities).

- Registers a PMPro ability category and nine REST-exposed abilities for membership levels, member memberships, orders, subscriptions, and scoped PMPro search.
- Includes capability-aware permission callbacks, structured schemas, privacy-filtered outputs, pagination, and MCP metadata.
- Supports administrative membership changes and cancellations through existing PMPro APIs so normal hooks and integrations continue to run.
- Adds a reusable structured-results mode to PMPro quick search while preserving its existing HTML/JSON response behavior.
- Remains inert on WordPress versions that do not provide the Abilities API and adds the `pmpro_enable_abilities_api` opt-out filter.

1. Run WordPress 6.9 or newer with Paid Memberships Pro active.
2. Confirm all nine `pmpro/*` abilities appear in the WordPress ability registry.
3. As an administrator, execute the read abilities and confirm their results pass the declared output schemas.
4. Validate the two membership write abilities with valid user and level IDs; execute them only against disposable member data.
5. Confirm anonymous users cannot execute any PMPro ability.
6. Confirm the existing PMPro Quick Search UI still renders its HTML results.

### Other information:

- [x] Added an explanation of the implementation and rationale.
- [x] Successfully ran local validation.

Local validation covered PHP syntax, staged diff whitespace, live registration of all nine abilities on WordPress 7.0, runtime schema validation for every read ability, input and permission validation for both write abilities without executing mutations, and anonymous-denial checks for all abilities.

### Changelog entry

> FEATURE: Added WordPress Abilities API support for querying and managing PMPro membership, order, subscription, and search data through REST and MCP integrations.

### DIFF
--- a/includes/abilities.php
+++ b/includes/abilities.php
@@ -3,7 +3,7 @@
  * Register Paid Memberships Pro abilities for the WordPress Abilities API.
  *
  * @package PaidMembershipsPro
- * @since TBD
+ * @since 3.8.3
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
@@ -16,7 +16,7 @@ add_action( 'wp_abilities_api_init', 'pmpro_abilities_register_abilities' );
 /**
  * Check whether the WordPress Abilities API is available.
  *
- * @since TBD
+ * @since 3.8.3
  *
  * @return bool
  */
@@ -29,7 +29,7 @@ function pmpro_abilities_is_abilities_api_available() {
 /**
  * Check whether PMPro can register abilities.
  *
- * @since TBD
+ * @since 3.8.3
  *
  * @return bool
  */
@@ -37,7 +37,7 @@ function pmpro_abilities_can_boot() {
 	/**
 	 * Filter whether PMPro should register Abilities API categories and abilities.
 	 *
-	 * @since TBD
+	 * @since 3.8.3
 	 *
 	 * @param bool $enabled Whether ability registration is enabled.
 	 */
@@ -79,7 +79,7 @@ function pmpro_abilities_register_abilities() {
 	}
 
 	$abilities = array(
-		'pmpro/memberships-query'         => pmpro_abilities_get_memberships_query_definition(),
+		'pmpro/levels-query'              => pmpro_abilities_get_levels_query_definition(),
 		'pmpro/member-memberships-get'    => pmpro_abilities_get_member_memberships_definition(),
 		'pmpro/member-membership-change'  => pmpro_abilities_get_member_membership_change_definition(),
 		'pmpro/member-membership-cancel'  => pmpro_abilities_get_member_membership_cancel_definition(),
@@ -185,7 +185,6 @@ function pmpro_abilities_get_member_level_schema() {
 			'name'              => array( 'type' => 'string' ),
 			'description'       => array( 'type' => 'string' ),
 			'status'            => array( 'type' => 'string' ),
-			'subscription_id'   => array( 'type' => array( 'integer', 'null' ) ),
 			'startdate'         => array( 'type' => array( 'string', 'null' ) ),
 			'enddate'           => array( 'type' => array( 'string', 'null' ) ),
 			'initial_payment'   => array( 'type' => 'number' ),
@@ -257,13 +256,13 @@ function pmpro_abilities_get_subscription_schema() {
  *
  * @return array
  */
-function pmpro_abilities_get_memberships_query_definition() {
+function pmpro_abilities_get_levels_query_definition() {
 	return array(
 		'label'               => __( 'Query Membership Levels', 'paid-memberships-pro' ),
 		'description'         => __( 'Query Paid Memberships Pro membership levels with pagination and optional text filtering.', 'paid-memberships-pro' ),
 		'category'            => 'pmpro',
-		'execute_callback'    => 'pmpro_abilities_execute_memberships_query',
-		'permission_callback' => 'pmpro_abilities_can_query_memberships',
+		'execute_callback'    => 'pmpro_abilities_execute_levels_query',
+		'permission_callback' => 'pmpro_abilities_can_query_levels',
 		'input_schema'        => array(
 			'type'       => 'object',
 			'default'    => array(
@@ -360,7 +359,6 @@ function pmpro_abilities_get_member_membership_change_definition() {
 			'properties' => array(
 				'user_id'          => array( 'type' => 'integer' ),
 				'level_id'         => array( 'type' => 'integer' ),
-				'reason'           => array( 'type' => 'string' ),
 				'old_level_status' => array( 'type' => 'string' ),
 			),
 			'required'   => array( 'user_id', 'level_id' ),
@@ -410,8 +408,6 @@ function pmpro_abilities_get_member_membership_cancel_definition() {
 			'properties' => array(
 				'user_id' => array( 'type' => 'integer' ),
 				'level_id' => array( 'type' => 'integer' ),
-				'status'  => array( 'type' => 'string' ),
-				'reason'  => array( 'type' => 'string' ),
 			),
 			'required'   => array( 'user_id', 'level_id' ),
 		),
@@ -683,7 +679,7 @@ function pmpro_abilities_can_manage_members() {
  *
  * @return bool
  */
-function pmpro_abilities_can_query_memberships() {
+function pmpro_abilities_can_query_levels() {
 	if ( ! defined( 'PMPRO_VERSION' ) ) {
 		return false;
 	}
@@ -811,7 +807,6 @@ function pmpro_abilities_normalize_member_level( $membership ) {
 		'name'              => isset( $membership->name ) ? (string) $membership->name : '',
 		'description'       => isset( $membership->description ) ? wp_strip_all_tags( $membership->description ) : '',
 		'status'            => isset( $membership->status ) ? (string) $membership->status : 'active',
-		'subscription_id'   => isset( $membership->subscription_id ) ? (int) $membership->subscription_id : null,
 		'startdate'         => isset( $membership->startdate ) ? pmpro_abilities_normalize_unix_timestamp( $membership->startdate ) : null,
 		'enddate'           => isset( $membership->enddate ) ? pmpro_abilities_normalize_unix_timestamp( $membership->enddate ) : null,
 		'initial_payment'   => isset( $membership->initial_payment ) ? (float) $membership->initial_payment : 0.0,
@@ -899,7 +894,7 @@ function pmpro_abilities_get_user_or_error( $user_id ) {
  * @param array $input Ability input.
  * @return array
  */
-function pmpro_abilities_execute_memberships_query( $input ) {
+function pmpro_abilities_execute_levels_query( $input ) {
 	if ( ! defined( 'PMPRO_VERSION' ) ) {
 		return new WP_Error( 'pmpro_abilities_missing_pmpro', __( 'Paid Memberships Pro is not active.', 'paid-memberships-pro' ), array( 'status' => 503 ) );
 	}
@@ -1058,8 +1053,7 @@ function pmpro_abilities_execute_member_membership_cancel( $input ) {
 		$previous_memberships = array();
 	}
 
-	$status = ! empty( $input['status'] ) ? (string) $input['status'] : 'admin_cancelled';
-	$result = pmpro_cancelMembershipLevel( (int) $level->id, (int) $user->ID, $status );
+	$result = pmpro_cancelMembershipLevel( (int) $level->id, (int) $user->ID, 'admin_cancelled' );
 
 	$current_memberships = pmpro_getMembershipLevelsForUser( $user->ID, true );
 	if ( ! is_array( $current_memberships ) ) {
@@ -1103,8 +1097,9 @@ function pmpro_abilities_execute_orders_query( $input ) {
 			'page'  => 1,
 		)
 	);
-	$page  = max( 1, (int) $input['page'] );
-	$limit = max( 1, min( 50, (int) $input['limit'] ) );
+	$page   = max( 1, (int) $input['page'] );
+	$limit  = max( 1, min( 50, (int) $input['limit'] ) );
+	$offset = ( $page - 1 ) * $limit;
 
 	$query_args = array();
 	foreach ( array( 'user_id', 'membership_level_id', 'status', 'gateway' ) as $key ) {
@@ -1126,13 +1121,14 @@ function pmpro_abilities_execute_orders_query( $input ) {
 		array_merge(
 			$query_args,
 			array(
-				'limit' => $page * $limit,
+				'limit'  => $limit,
+				'offset' => $offset,
 			)
 		)
 	);
 
 	return array(
-		'items' => array_map( 'pmpro_abilities_normalize_order', pmpro_abilities_slice_items( $orders, $page, $limit ) ),
+		'items' => array_map( 'pmpro_abilities_normalize_order', $orders ),
 		'total' => $total,
 		'page'  => $page,
 		'limit' => $limit,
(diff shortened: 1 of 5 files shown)


flintfromthebasement (changes requested):
**PR: strangerstudios/paid-memberships-pro#3736 — Add PMPro abilities for WordPress and MCP integrations**
*mircobabini → dev | 3 files, +1330 -3*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3736>

---

**Summary**
Clean architecture — capability-scoped permission callbacks, privacy-filtered schemas, inert on WP < 6.9 — but not ready to merge. The cancellation ability lets a caller pass a `status` value that leaves a "cancelled" membership looking active and skips PMPro's own hooks, the subscription-query ability doesn't scale, and a returned field will silently mislead any MCP client that acts on it.

---

**Issues**

- **Critical** `includes/abilities.php` (`pmpro_abilities_execute_member_membership_cancel`) — `status` is an unconstrained string (`'status' => array( 'type' => 'string' )`) passed straight to `pmpro_cancelMembershipLevel()`. Verified against `includes/functions.php:1187,1199,1216`: the DB write sets `status` to whatever was passed, and `if ( ! in_array( $status, array( 'changed', 'admin_changed' ) ) )` gates both `pmpro_before/after_change_membership_level` and `pmpro_clear_level_cache_for_user()`. Pass `status: "admin_changed"` (a real PMPro string, not garbage) and the cancel writes an `enddate` but skips clearing the level cache and skips the hooks other integrations rely on — the ability's own response, which re-reads `pmpro_getMembershipLevelsForUser()`, can then report the level as still present. Whitelist `status` to cancellation values only (`cancelled`, `admin_cancelled`, `expired`, etc.) via an `enum` in the input schema, or hardcode it and drop the input entirely.

- **Major** `includes/abilities.php` (`pmpro_abilities_execute_subscriptions_query`) — the count path calls `PMPro_Subscription::get_subscriptions( array_merge( $query_args, array( 'limit' => 0 ) ) )`. Confirmed in `classes/class-pmpro-subscription.php:493-520`: `if ( $limit )` treats `0` as falsy, so **no `LIMIT` clause is applied at all** — every matching row is selected, and then the method loops and constructs a full `PMPro_Subscription` object per row (`new PMPro_Subscription( $subscription_id )`, each its own `SELECT *`, `class-pmpro-subscription.php:207-216`), purely so the ability can call `count()` on the array. `MemberOrder::get_orders()` has a working `return_count` mode used correctly in the orders ability — `PMPro_Subscription` has no equivalent. This is one ability call turning into N+1 full-row queries on any site with meaningful subscription history. Add a `COUNT(*)`-based count path (new method or direct `$wpdb->get_var()` with the same WHERE logic) instead of hydrating every row.

- **Major** `includes/abilities.php` (`pmpro_abilities_get_member_level_schema`, `pmpro_abilities_normalize_member_level`) — the `subscription_id` field returned to MCP callers is not a real subscription ID. `pmpro_getMembershipLevelsForUser()` aliases `mu.id AS subscription_id` (`functions.php:2441`), where `mu.id` is the `pmpro_memberships_users` join-table row ID. `pmpro/subscription-get` expects an actual `pmpro_subscriptions.id` and does `new PMPro_Subscription( (int) $input['subscription_id'] )`. A caller that reasonably chains "get member memberships" → "get subscription" using this field will get the wrong row or a 404. Rename to something like `membership_record_id`, or resolve and return the real subscription ID.

- **Minor** `includes/abilities.php` (`pmpro_abilities_execute_member_memberships_get`, `_change`, `_cancel`) — all three call `pmpro_getMembershipLevelsForUser( $user->ID, true )` with `$include_inactive = true`, so `memberships`/`membership_count`/`current_memberships` include every historical cancelled/expired row, not just current ones — despite the ability being documented as "current membership assignments." A consumer checking `membership_count > 0` as an "is this user a member" proxy gets a false positive for any past member. If full history is intended, say so in the description; otherwise pass `false`.

- **Minor** `includes/abilities.php` (`pmpro_abilities_get_member_membership_change_definition`, `_cancel_definition`) — both schemas declare a `reason` input that is never read in the corresponding execute callbacks. Either wire it into the response/log or drop it — right now it's a documented no-op parameter.

- **Minor** `includes/abilities.php` (`pmpro_abilities_get_order_get_definition`) — neither `order_id` nor `order_code` is marked `required`. Omitting both returns a 404 "Order not found," which misattributes a missing-input error as a missing-resource error.

---

**Looks Good**
- `pmpro_abilities_execute_search_query` relies on `pmpro_rest_api_quick_search()`'s existing per-section capability checks (confirmed at `rest-api.php:1379,1460,1501,1546,1567,1592`) even when scope is `all` — the ability's own permission callback is a coarse pre-filter, but real per-domain authorization still happens inside the shared function, so there's no cross-domain leak.
- The `$return_raw`/`$limit_override` addition to `pmpro_rest_api_quick_search()` is minimal and backward-compatible — default parameters preserve existing HTML output exactly.
- The opt-out pattern (`pmpro_enable_abilities_api` filter + `pmpro_abilities_is_abilities_api_available()` function-existence guard) makes the whole feature genuinely inert on older WP with no load-time cost.
- Output normalization consistently strips tags and casts scalars; gateway transaction IDs, customer IDs, and payment tokens are deliberately excluded from the order/subscription schemas.

---

**Questions**
1. `pmpro/member-membership-change` calls `pmpro_changeMembershipLevel()`, which replaces all existing levels. With Multiple Memberships Per User enabled, should this add a level instead? "Reassign" in the description implies replacement, but MCP consumers may expect additive behavior.
2. `pmpro_abilities_execute_search_query` builds a `WP_REST_Request` and calls the controller method directly rather than through `rest_do_request()`. Any reason to bypass standard REST dispatch (and its `rest_pre_dispatch` filters) here?

flintfromthebasement (commented):
Reviewed as part of the "PMPro CLI and WP MCP/Abilities" work (companion PRs: #3739 REST bulk endpoints, #3740 pmpro-cli). Overall this is in good shape and close to shippable — nice work. 👏

**What's solid**
- Inert-by-default on older WP: `pmpro_abilities_is_abilities_api_available()` gates on `wp_register_ability` / `wp_register_ability_category` / `WP_Ability`, plus the `pmpro_enable_abilities_api` opt-out filter. Category registration is dedup-guarded.
- Permission callbacks are capability-gated with a `manage_options` fallback and a `PMPRO_VERSION` guard.
- The `scope=all` search does **not** leak across capabilities: `pmpro_rest_api_quick_search()` enforces per-type caps inside the method body, so a user with only `pmpro_orders` gets only the orders group even though `can_run_search_query()` passes coarsely. Confirmed by reading the quick_search handler.
- Write abilities go through `pmpro_changeMembershipLevel()` / `pmpro_cancelMembershipLevel()`, so hooks/integrations fire, and they handle the null (unchanged) vs false (failed) return distinction with proper `WP_Error` statuses.
- MCP annotations are correct: reads are `readonly:true / destructive:false`, writes are `readonly:false / destructive:true`.

**Suggestions**
1. **Rename `pmpro/memberships-query` → `pmpro/levels-query` (or `membership-levels-query`).** It queries membership *level definitions* (`pmpro_getAllLevels`), not members. An agent reading "memberships-query" will expect the member roster, which this isn't — the naming collides conceptually with `member-memberships-get`. Clearer names reduce agent misfires.
2. **Fast-follow: add a `pmpro/members-query` ability** for the actual member roster (users holding memberships) with filter + pagination. #3739 adds a reusable `pmpro_get_members()` helper built for exactly this; once it lands, backing an ability with it makes the abilities surface match the REST (`/members`) and CLI (`wp pmpro member list`) surfaces. Not blocking this PR.
3. `@since TBD` tags need the target milestone at merge (same on my two PRs).

Recommendation: ship it (optionally fold in #1's rename first). #2 can follow once #3739 merges. The three PRs together give agents the same capability set through three front doors — abilities, REST, and CLI.

flintfromthebasement (commented):
**Follow-up review**

**PR: strangerstudios/paid-memberships-pro#3736 — Add PMPro abilities for WordPress and MCP integrations**
*mircobabini → dev | +90 -40 since last review (commit a29b9e3a → aaa5f4fb)*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3736>

**Summary**
All three blocking issues from my first review are resolved, and both suggestions from my second pass landed too. This is mergeable — the remaining two items are pre-existing minors I'd note but wouldn't hold the PR for.

**Status of prior findings**

- **Resolved** `includes/abilities.php` (cancel `status` input) — the `status` field is gone from the input schema entirely and `pmpro_cancelMembershipLevel()` is now called with a hardcoded `'admin_cancelled'`. No caller-controlled bypass of the hooks/cache-clear gate anymore.
- **Resolved** `classes/class-pmpro-subscription.php` (`get_subscriptions()` N+1) — added a real `return_count` mode that issues `SELECT COUNT(*)` against the same WHERE clause instead of hydrating every row. `pmpro_abilities_execute_subscriptions_query()` now calls it with `'return_count' => true'` for `$total`, and switched the actual query to `limit`/`offset` instead of `page * $limit`. As a bonus, `pmpro_abilities_execute_orders_query()` picked up the same `limit`/`offset` treatment (dropping the old `pmpro_abilities_slice_items()` post-fetch slice), which wasn't something I flagged but is a clean improvement in the same spirit.
- **Resolved** `includes/abilities.php` (misleading `subscription_id` field) — removed from both the schema and `pmpro_abilities_normalize_member_level()` rather than renamed/fixed. Simpler than what I suggested, and it fully removes the false-lead for MCP clients.
- **Resolved** `includes/abilities.php` (unused `reason` input) — dropped from both the change and cancel input schemas; no more documented no-op parameter.
- **Not addressed** `includes/abilities.php` (`pmpro_getMembershipLevelsForUser( $user->ID, true )` still passed `true` in `member_memberships_get`/`_change`/`_cancel`) — still returns full history including cancelled/expired rows under fields documented as "current." Low urgency, but still worth either documenting or fixing.
- **Not addressed** `includes/abilities.php` (`pmpro_abilities_get_order_get_definition`) — `order_id`/`order_code` are still both optional with no `required`/`oneOf`; omitting both still surfaces as a 404 "Order not found" rather than a validation error.
- **Resolved** (from second review) rename `pmpro/memberships-query` → `pmpro/levels-query` — done, including the definition/execute/permission function names (`pmpro_abilities_get_levels_query_definition`, `pmpro_abilities_execute_levels_query`, `pmpro_abilities_can_query_levels`).
- **Resolved** (from second review) `@since TBD` tags — all replaced with `@since 3.8.3` across `abilities.php` and the two touched core classes.
- Members-query fast-follow (#2 from second review) — correctly deferred to a follow-up PR, not blocking.

**New findings from this diff** — none. The offset/limit changes are correctly scoped (offset is only appended when a limit is present, and the count query skips ORDER/LIMIT/OFFSET entirely), and the same pattern was applied consistently to both `MemberOrder::get_orders()` and `PMPro_Subscription::get_subscriptions()`.

**Looks Good**
- The count-query refactor shares the exact same WHERE-clause construction as the row query, so total counts can't drift from what's actually returned.
- Consistent, mirrored change across both the order and subscription pagination paths rather than a one-off fix.

Recommendation: approve.

ideadude:
This is ready to merge. We need to change the since here:
>@since 3.8.3 Added support for the `offset` and `return_count` arguments.

Should be @since TBD or @since 3.9 if this merges into 3.9

I usually don't want these since updates on methods UNLESS the params or return value is changed or the function does something drastically different. In this case the return is changing.

includes/abilities.php
```diff
@@ -0,0 +1,1315 @@
...
+	if ( ! defined( 'PMPRO_VERSION' ) ) {
+		return new WP_Error( 'pmpro_abilities_missing_pmpro', __( 'Paid Memberships Pro is not active.', 'paid-memberships-pro' ), array( 'status' => 503 ) );
+	}
+
+	$user = pmpro_abilities_get_user_or_error( (int) $input['user_id'] );
+	if ( is_wp_error( $user ) ) {
+		return $user;
+	}
+
+	$memberships = pmpro_getMembershipLevelsForUser( $user->ID, true );
```
dparker1005, line 956: Question to discuss: passing `true` here is `$include_inactive`, so this returns the user's **entire membership history** — and in that branch the query has no `GROUP BY`, so you get one row per historical level change, including repeats of the same level. Meanwhile the ability's description says "current membership assignments" and `membership_count` will count historical rows. The change/cancel abilities also embed two full copies of this list (`previous_memberships` / `current_memberships`) in every response.

On one hand, since the consumer is an AI, oversharing is arguably fine — the model can filter. On the other, a long-standing member could have a very large history feeding straight into an LLM context window with no pagination, and the "current" label would actively mislead the model.

Options I see:
1. Default to active-only and add an `include_inactive` boolean to the input schema for callers that want history.
2. Keep returning everything but fix the description/`membership_count` semantics (and maybe dedupe).

I lean toward option 1 for `member-memberships-get`, and active-only for the `previous_memberships`/`current_memberships` snapshots in the write abilities. What was the thinking on your end?

includes/abilities.php
```diff
@@ -0,0 +1,1315 @@
...
+	if ( empty( $level ) ) {
+		return new WP_Error( 'pmpro_abilities_level_not_found', __( 'Membership level not found.', 'paid-memberships-pro' ), array( 'status' => 404 ) );
+	}
+
+	$previous_memberships = pmpro_getMembershipLevelsForUser( $user->ID, true );
+	if ( ! is_array( $previous_memberships ) ) {
+		$previous_memberships = array();
+	}
+
+	$old_level_status = ! empty( $input['old_level_status'] ) ? (string) $input['old_level_status'] : 'admin_changed';
```
dparker1005, line 996: `old_level_status` gets written to the user's old membership rows as-is, and PMPro has status-sensitive logic downstream (e.g., `pmpro_cancelMembershipLevel()` branches on `'changed'`/`'admin_changed'`). An MCP client passing something creative like `'active'` would leave the old level rows active alongside the new one. Since AI clients are exactly the callers that will improvise values, I'd like to restrict this to the statuses core itself uses:

```suggestion
	$old_level_status = ! empty( $input['old_level_status'] ) ? pmpro_sanitize_with_safelist( (string) $input['old_level_status'], array( 'inactive', 'changed', 'admin_changed', 'cancelled', 'admin_cancelled', 'expired' ) ) : 'admin_changed';
	if ( empty( $old_level_status ) ) {
		return new WP_Error( 'pmpro_abilities_invalid_old_level_status', __( 'Invalid old level status.', 'paid-memberships-pro' ), array( 'status' => 400 ) );
	}
```

dparker1005 (commented):
Thanks @mircobabini — this is in really good shape. The registration guards, per-bucket capability re-checks in quick search, and routing writes through `pmpro_changeMembershipLevel()`/`pmpro_cancelMembershipLevel()` so hooks and gateway cancellation fire all look right to me.

Leaving two inline comments to discuss before we merge — one about what data the memberships abilities should return, and one about validating `old_level_status`.

includes/abilities.php
ideadude: I meant to go with 1 here. I think #3740 expects that then. Active makes more sense for what people want and matches the comment/docs. Param for history if needed. We should double check, but I don't think this change impacts things we did. Small chance someone was relying on this returning everything in their API use. Easy fix if needed. We can call it out as clearly as possible in the changelog.

includes/abilities.php
ideadude: I like it.

I wonder if we should have a helper function to get valid membership statuses with a filter in case people want to add them? Maybe wait for someone to ask. If we have similar "here's a list of valid statuses" arrays in the codebase, would be good just to abstract it to one place if we want to edit in the future.

But okay to ship as is too.

PR #3737, opened 2026-07-22
Add automatic cleanup for abandoned checkout users

dparker1005:
## Summary

- add an opt-in Advanced Setting to automatically delete abandoned checkout users
- delete users still carrying the existing `pmpro_abandoned_signup` term after 30 days
- allow the retention period to be customized with `pmpro_abandoned_signup_deletion_days`
- process one bounded batch during the existing `pmpro_schedule_daily` Action Scheduler task
- use `wp_delete_user()` and load its admin dependency in the scheduled callback

## Testing

- `git diff --check`
- PHP syntax check for `includes/abandoned-signups.php`
- stubbed assertions for the Advanced Settings field, UTC cutoff, taxonomy query, and batch size

## Scope

This uses the existing abandoned-signup taxonomy as the source of truth. It does not add new spam-detection heuristics, historical rescanning, or deletion logging.

### DIFF
--- a/includes/abandoned-signups.php
+++ b/includes/abandoned-signups.php
@@ -134,6 +134,10 @@ function pmpro_add_abandoned_signup_advanced_setting( $settings ) {
 		),
 	);
 
+	if ( is_multisite() ) {
+		$settings['auto_delete_abandoned_signups']['description'] .= ' ' . __( 'On multisite, users are removed from this site only.', 'paid-memberships-pro' );
+	}
+
 	return $settings;
 }
 add_filter( 'pmpro_custom_advanced_settings', 'pmpro_add_abandoned_signup_advanced_setting' );
@@ -191,11 +195,8 @@ function pmpro_delete_abandoned_signups() {
 		return;
 	}
 
-	// wp_delete_user() and wpmu_delete_user() are not loaded automatically during scheduled actions.
+	// wp_delete_user() is not loaded automatically during scheduled actions.
 	require_once ABSPATH . 'wp-admin/includes/user.php';
-	if ( is_multisite() ) {
-		require_once ABSPATH . 'wp-admin/includes/ms.php';
-	}
 
 	foreach ( $user_ids as $user_id ) {
 		$user_id = (int) $user_id;
@@ -205,13 +206,9 @@ function pmpro_delete_abandoned_signups() {
 			continue;
 		}
 
-		// On multisite, wp_delete_user() only removes the user from the current site. Delete the account
-		// from the network instead, unless the user also belongs to another site.
-		if ( is_multisite() && count( get_blogs_of_user( $user_id, true ) ) <= 1 ) {
-			wpmu_delete_user( $user_id );
-		} else {
-			wp_delete_user( $user_id, null );
-		}
+		// On multisite, this only removes the user from the current site. The user may be active on
+		// another site in the network, so network-wide deletion is left to the network admin.
+		wp_delete_user( $user_id, null );
 	}
 }
 add_action( 'pmpro_schedule_daily', 'pmpro_delete_abandoned_signups' );
--- a/includes/abandoned-signups.php
+++ b/includes/abandoned-signups.php
@@ -90,6 +90,115 @@ function pmpro_remove_abandoned_signup_taxonomy_on_page_load() {
 add_action( 'wp', 'pmpro_remove_abandoned_signup_taxonomy_on_page_load' );
 
 /**
+ * Get the number of days that an abandoned signup should be kept before deletion.
+ *
+ * @since 3.9
+ *
+ * @return int The number of days to keep abandoned signups.
+ */
+function pmpro_get_abandoned_signup_deletion_days() {
+	/**
+	 * Filter the number of days that an abandoned signup should be kept before deletion.
+	 *
+	 * @since 3.9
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3737 — Add automatic cleanup for abandoned checkout users**
*dparker1005 → dev | 1 file, +109 -0*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3737>

---

**Summary**
Well-scoped opt-in feature: off by default, bounded batch on the existing daily action, fully prepared SQL, and it reuses the existing `pmpro_abandoned_signup` taxonomy as the source of truth. One correctness bug on multisite blocks merge as-is; the rest is minor.

---

**Issues**

- **Major** `includes/abandoned-signups.php:196` — `wp_delete_user( (int) $user_id, null )` does not delete the account on multisite. In WP core (`wp-admin/includes/user.php:441-450`) the `is_multisite()` branch calls `remove_user_from_blog()` and never touches `wp_users`/`wp_usermeta`. It still fires `deleted_user`, which this file already hooks at line 65 (`pmpro_remove_abandoned_signup_taxonomy`), so on a network the batch strips the abandoned term from accounts that persist forever. They then vanish from future queries, and the setting's copy ("Permanently delete user accounts") is false. Note the `null` reassign branch still deletes authored posts first, so it's partially destructive without actually removing the user. PMPro already handles this split in `includes/cleanup.php:37-38` (hooks both `delete_user` and `wpmu_delete_user`). Fix: branch on `is_multisite()` and call `wpmu_delete_user()` (needs `require_once ABSPATH . 'wp-admin/includes/ms.php'`), or hide/disable the setting on multisite until that's wired.

- **Minor** `includes/abandoned-signups.php:190-198` — No re-check between the SELECT and the delete. A user could complete checkout (clearing the term via `pmpro_after_checkout`) between the query and the loop. The window is tiny, but `is_object_in_term( $user_id, 'pmpro_abandoned_signup', 'abandoned-signup' )` before each `wp_delete_user()` is cheap and closes it. Consider also skipping any user with a row in `pmpro_memberships_users` as a belt-and-braces guard, since this path is irreversible.

- **Minor** `includes/abandoned-signups.php:93,109,151` — docblocks use `@since 3.9`. Everything else in flight in this repo uses `@since TBD` (95 occurrences vs. one hardcoded value), and the conventions say TBD is swapped at release. Switch to `TBD` so the release pass catches these.

---

**Looks Good**
- Query is a proper prepared statement with taxonomy, term ID, cutoff, and limit all bound. `gmdate()` for the cutoff is correct: `wp_insert_user()` stores `user_registered` in UTC (`wp-includes/user.php:2421`), so there's no timezone skew.
- Batch bound via `PMPRO_CRON_LIMIT` with a fallback of 250 and an `absint()` zero-guard matches the existing cron convention.
- Option round-trip is right: `pmpro_custom_advanced_settings` saves through `pmpro_setOption()` with the `pmpro_` prefix, and the read at line 152 uses the prefixed name. The `is_associative` / `options` shape matches the Advanced Settings field renderer.
- The abandoned term is only ever applied to brand-new accounts created inside checkout (`preheaders/checkout.php`, new-user branch), so existing members can't be tagged by a third party.

---

**Questions**
1. Was this tested on a multisite install, or single-site only?
2. Can a tagged user become a paying member without clearing the tag, e.g. a pending/eCheck order activated later by webhook, or an admin assigning a level via Add Member? Neither path fires `pmpro_after_checkout` as far as I can see, and such a user who never loads a front-end page would be deleted at day 30. If that's a real path, the membership-row exclusion above becomes a requirement rather than a nicety.
3. Deleting with `reassign = null` also deletes any posts the user authored. Intended, or should the setting reassign to an admin?

dparker1005:
Thanks for the thorough pass. Here's where things landed across the two follow-up commits.

**Multisite (Major).** You were right that `wp_delete_user()` only removes the user from the current site there. I first switched to `wpmu_delete_user()`, then traced the path further and backed it out in 40e5583. The problem: the abandoned tag and the memberships table are both per-site. A user who abandons checkout on subsite A, then completes a paid checkout on subsite B while still logged in via the network cookie, keeps A's tag forever. Nothing on B clears it, and PMPro never calls `add_user_to_blog()`, so a blog-membership check doesn't see B either. Site A's cleanup would have deleted a paying member on B network-wide. Removing the user from the current site is the safe behavior, and it's exactly what a subsite admin can already do manually. Network-wide cleanup stays with the network admin. The setting description now says so on multisite.

**Race re-check (Minor).** Added. Each user is re-checked with `is_object_in_term()` right before deletion.

**Membership guard (Minor) and your question 2.** This turned out to be required, not belt-and-braces. `pmpro_after_checkout` only fires from `pmpro_complete_checkout()`. The IPN and webhook handlers never fire it, so a PayPal Standard buyer who pays and closes the tab before the confirmation page keeps the tag while holding a membership. Same for an admin-assigned level. The query now excludes anyone with a row in the memberships table.

**`@since` (Minor).** Switched to TBD.

**Question 1.** Single-site only so far.

**Question 3.** Keeping `reassign = null`. A tagged user has never loaded a page, so any authored posts came from a plugin auto-creating content on registration, and those should go with the account.

andrewlimaza (approved):

PR #3739, opened 2026-07-24
Add REST API bulk collection endpoints for members, orders, and subscriptions

flintfromthebasement:

Adds REST API **bulk collection endpoints** so AI agents and integrations can query PMPro data at scale instead of making N single-record calls. Part of the "PMPro CLI and WP MCP/Abilities" pitch (companion PRs: pmpro-cli, and @mircobabini's abilities PR #3736).

New read-only endpoints under `pmpro/v1`:

- `GET /pmpro/v1/members` — filter by `level`, `status`, `search`; paginate; select fields.
- `GET /pmpro/v1/orders` — filter by `user_id`, `level`, `status`, `gateway`, `start_date`, `end_date`.
- `GET /pmpro/v1/subscriptions` — filter by `user_id`, `level`, `status`, `gateway`.

All three support:
- **Pagination** — `page` + `per_page`, with `X-WP-Total` and `X-WP-TotalPages` response headers. Hard-capped (default max 100) via the new `pmpro_rest_api_max_per_page` filter, so no unbounded queries.
- **Field selection** — `fields=id,status,total` to trim the payload.
- The **existing permission model** (`pmpro_rest_api_get_permissions_check`) — `pmpro_edit_members` for members/subscriptions, `pmpro_orders` for orders. No new auth scheme.

Supporting changes:
- New reusable `pmpro_get_members( $args )` query helper (with a `pmpro_get_members_sql` filter) in `includes/functions.php`, shared by this endpoint and the forthcoming `pmpro-cli`. There was no core function to query members with filters + pagination; this fills that gap.
- Backward-compatible `offset` support added to `MemberOrder::get_orders()` and `PMPro_Subscription::get_subscriptions()` (both already supported `limit`) to enable pagination.

1. With PMPro active and some members/orders/subscriptions present, authenticate as a user with `pmpro_edit_members` / `pmpro_orders`.
2. `GET /wp-json/pmpro/v1/members?status=active&per_page=5&page=1` — confirm 5 members, correct `X-WP-Total` header, and page 2 returns the next set.
3. `GET /wp-json/pmpro/v1/orders?status=success&fields=id,total,status` — confirm only those fields return.
4. `GET /wp-json/pmpro/v1/subscriptions?status=active` — confirm the collection and pagination headers.
5. Confirm a user without the relevant capability gets a 401/403.

### Changelog entry

> FEATURE: Added REST API bulk collection endpoints (`/members`, `/orders`, `/subscriptions`) with filtering, pagination, and field selection.

---
*`@since TBD` tags are placeholders — set to the target milestone at merge.*

### DIFF
--- a/classes/class-pmpro-subscription.php
+++ b/classes/class-pmpro-subscription.php
@@ -494,6 +494,13 @@ class PMPro_Subscription {
 		if ( $limit ) {
 			$sql_query .= ' LIMIT %d';
 			$prepared[] = $limit;
+
+			// Maybe offset the data for pagination.
+			$offset = isset( $args['offset'] ) ? (int) $args['offset'] : 0;
+			if ( $offset > 0 ) {
+				$sql_query .= ' OFFSET %d';
+				$prepared[] = $offset;
+			}
 		}
 
 		// Maybe prepare the query.
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -725,6 +725,13 @@
 				if ( $limit ) {
 					$sql_query .= ' LIMIT %d';
 					$prepared[] = $limit;
+
+					// Maybe offset the data for pagination.
+					$offset = isset( $args['offset'] ) ? (int) $args['offset'] : 0;
+					if ( $offset > 0 ) {
+						$sql_query .= ' OFFSET %d';
+						$prepared[] = $offset;
+					}
 				}
 			}
 
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -5520,3 +5520,135 @@ function pmpro_update_post_level_restrictions( $post_id, $level_ids ) {
 		do_action( 'pmpro_after_updating_post_level_restrictions', $post_id );
 	}
 }
+
+/**
+ * Query PMPro members (WordPress users who hold a membership) with filtering and pagination.
+ *
+ * Provides a single reusable member query for the REST API collection endpoint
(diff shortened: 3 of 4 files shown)


flintfromthebasement:
Pushed one follow-up commit after testing these endpoints against a real site (WP 7.0.2, ~1,400 members, live orders):

**`fields=` now validates.** Previously an unknown field name was silently dropped, so `fields=id,user_login,membership_level_id` on `/members` returned only `user_login` with no indication the other two do not exist. It now returns a 400 that names the invalid fields *and* lists the valid ones:

```
Unknown field(s): id, membership_level_id. Valid fields are: user_id, user_login,
user_email, display_name, membership_id, membership_name, status, joindate,
startdate, enddate.
```

That matters more than usual here: the main consumers of these endpoints are agents, and a silent partial response reads as "this field is empty" rather than "you asked for the wrong thing."

**Order amounts are numeric.** `/orders` was returning `subtotal`/`tax`/`total` as raw DB strings (`"total":"10"`) while the abilities API declares them `type: number` and casts to float. Same data, two types, one plugin. Now cast to float, with the ID/code fields cast to string to match how `/subscriptions` already builds its items.

Verified on the live site: valid field lists unchanged, unknown names 400, `"total":10` instead of `"total":"10"`.

ideadude:
This stuff is included in this PR now: https://github.com/strangerstudios/paid-memberships-pro/pull/3740

PR #3740, opened 2026-07-24
Add pmpro-cli: WP-CLI commands in core for members, levels, orders, subscriptions

flintfromthebasement:

Adds an official **`pmpro-cli` to PMPro core** — `wp pmpro <noun> <verb>` — so any install is scriptable by default and agents pointed at a site (or the repo) have a first-class CLI. Part of the "PMPro CLI and WP MCP/Abilities" pitch.

> **Depends on #3739** (REST bulk endpoints). This branch is stacked on it because the CLI reuses `pmpro_get_members()` and the query `offset` support added there. Until #3739 merges, this PR's diff includes those commits; I'll rebase once it lands.

Commands (thin wrappers over the same PMPro functions the REST API uses — no duplicated logic):

- `wp pmpro member list|get|change-level|cancel`
- `wp pmpro level list|get`
- `wp pmpro order list|get`
- `wp pmpro subscription list|get|sync`

Details:
- Standard WP-CLI output — `--format` (`table`/`csv`/`json`/`yaml`/`ids`/`count`) and `--fields`, via `WP_CLI\Formatter`.
- List commands support `--number`/`--page` pagination and the same filters as the REST endpoints.
- **Write commands** (`member change-level`, `member cancel`, `subscription sync`) support `--dry-run` (preview, no change) and prompt for confirmation unless `--yes` is passed — an over-eager script/agent can't silently mass-cancel memberships.
- `subscription sync` re-syncs one or more subscriptions with their gateway via `PMPro_Subscription::update()`. Accepts subscription IDs **or** gateway transaction IDs (e.g. `sub_XXXX`), and surfaces the `sync_error` subscription meta so a gateway-side failure shows as a warning instead of a silent success. (Ported from the POC's `sync-subscriptions-to-gateway`; originally opened as #3748, folded in here.)
- Loaded via `includes/cli.php`, which returns early unless `WP_CLI` is defined (zero overhead on web requests).
- `pmpro_cli_init` action lets Add Ons register their own commands.

Design note: this harvests the command shape from the `strangerstudios/pmpro-cli` POC and the dry-run/confirm patterns from PMPro Toolkit's CLI, but ships the general-purpose member/level/order/subscription commands in core rather than as a separate plugin. Ops/dev-only commands (bulk generate/delete test users, scrub data) intentionally stay out of core.

1. With PMPro active, run `wp pmpro member list --status=active --number=5`.
2. `wp pmpro order list --format=json` and `wp pmpro subscription list`.
3. `wp pmpro level list` and `wp pmpro level get <id>`.
4. `wp pmpro member change-level --user=<id> --level=<id> --dry-run` — confirm it previews without changing, then run without `--dry-run` and confirm the prompt.
5. `wp pmpro member get <id>` — confirm the member's levels display.
6. `wp pmpro subscription sync <id> --dry-run` — previews; then `wp pmpro subscription sync <gateway txn id>` — resolves by `subscription_transaction_id`, prompts, syncs. Bogus IDs error out before any sync runs.

### Changelog entry

> FEATURE: Added a `wp pmpro` WP-CLI command set (members, levels, orders, subscriptions) to core.

---
*`@since TBD` tags are placeholders — set to the target milestone at merge.*


flintfromthebasement:
Collapsed the CLI into one class in `includes/cli.php` (Toolkit shape). Deleted the abstract base and the per-noun subclasses.

Command names are unchanged (`wp pmpro member list`, etc.). Writes still wrap `pmpro_changeMembershipLevel()` / `pmpro_cancelMembershipLevel()`; cancel-all no longer goes through the deprecated `changeMembershipLevel(0)` path.

`pmpro_get_members()` stays in `includes/functions.php` — that's shared with the REST `/members` endpoint.

includes/rest-api.php
```diff
@@ -1253,6 +1345,382 @@ public function pmpro_rest_api_recent_orders( $request ) {
...
+				$query_args['membership_level_id'] = array_map( 'intval', explode( ',', (string) $params['level'] ) );
+			}
+			if ( isset( $params['status'] ) && '' !== $params['status'] ) {
+				$query_args['status'] = array_map( 'sanitize_text_field', explode( ',', (string) $params['status'] ) );
+			}
+			if ( ! empty( $params['gateway'] ) ) {
+				$query_args['gateway'] = sanitize_text_field( $params['gateway'] );
+			}
+
+			$total         = (int) PMPro_Subscription::get_subscriptions( array_merge( $query_args, array( 'return_count' => true ) ) );
```
dparker1005, line 1498: This is the line that depends on #3736 — `get_subscriptions()` on dev has no `return_count` support, so this returns an array of subscription objects and the cast produces 0/1 (plus it runs the full row query twice). Works once #3736 lands; flagging so we remember the merge order isn't optional.

includes/cli.php
```diff
@@ -0,0 +1,1039 @@
...
+	 * @param array $assoc_args Command flags, including --format and --fields.
+	 */
+	private function output_items( $items, $fields, $assoc_args ) {
+		if ( isset( $assoc_args['format'] ) && 'count' === $assoc_args['format'] ) {
+			WP_CLI::line( (string) count( $items ) );
+			return;
+		}
+
+		$formatter = new \WP_CLI\Formatter( $assoc_args, $fields );
+		$formatter->display_items( $items );
```
dparker1005, line 792: `--format=ids` is advertised on `level list`, `order list`, and `subscription list`, but the Formatter implodes the raw associative rows for `ids` — locally this outputs a stream of "Array to string conversion" warnings and no IDs. `member list` already blocks `ids` with a helpful error; for the other three, plucking the first field (always `id` for these commands) fixes it:

```suggestion
		if ( isset( $assoc_args['format'] ) && 'ids' === $assoc_args['format'] ) {
			$items = wp_list_pluck( $items, $fields[0] );
		}

		$formatter = new \WP_CLI\Formatter( $assoc_args, $fields );
		$formatter->display_items( $items );
```

includes/functions.php
```diff
@@ -5568,3 +5568,135 @@ function pmpro_update_post_level_restrictions( $post_id, $level_ids ) {
...
+
+	if ( $return_count ) {
+		if ( $prepared ) {
+			$sql = $wpdb->prepare( $sql, $prepared );
+		}
+		return (int) $wpdb->get_var( $sql );
+	}
+
+	// One row per user/level pair.
+	$sql .= ' GROUP BY u.ID, mu.membership_id';
```
dparker1005, line 5661: With `status=all`, a user who re-subscribed to the same level has multiple `pmpro_memberships_users` rows for one user/level pair, and this `GROUP BY` collapses them — MySQL then picks an arbitrary row's `status`/`startdate`/`enddate` (no SQL error since WP strips `ONLY_FULL_GROUP_BY`, just nondeterministic data). A re-subscribed member can show their old cancelled row's dates. Worth making the pick deterministic — e.g., join/aggregate on the latest `mu.id` per pair — or at minimum documenting that `status=all` output is per-pair, not per-row.

includes/functions.php
```diff
@@ -5568,3 +5568,135 @@ function pmpro_update_post_level_restrictions( $post_id, $level_ids ) {
...
+
+	/**
+	 * Filter the SQL used by pmpro_get_members().
+	 *
+	 * @since TBD
+	 *
+	 * @param string $sql  The prepared SQL query.
+	 * @param array  $args The parsed query arguments.
+	 */
+	$sql = apply_filters( 'pmpro_get_members_sql', $sql, $args );
```
dparker1005, line 5699: Let's drop this filter until there's a concrete use case — it has no callers anywhere (checked the repo and a GitHub-wide code search), and filtering the fully-prepared SQL string is a fragile contract: any future change to this query silently breaks anyone doing string surgery on it. If a customization need shows up later, filtering the `$args` array on the way in would be the safer shape.

```suggestion
```

includes/rest-api.php
```diff
@@ -256,6 +256,98 @@ public function pmpro_rest_api_register_routes() {
...
+					'args'     => array(
+						'page'       => array( 'sanitize_callback' => 'absint' ),
+						'per_page'   => array( 'sanitize_callback' => 'absint' ),
+						'user_id'    => array( 'sanitize_callback' => 'absint' ),
+						'level'      => array(),
+						'status'     => array(),
+						'gateway'    => array( 'sanitize_callback' => 'sanitize_text_field' ),
+						'start_date' => array( 'sanitize_callback' => 'sanitize_text_field' ),
+						'end_date'   => array( 'sanitize_callback' => 'sanitize_text_field' ),
+						'orderby'    => array( 'sanitize_callback' => 'sanitize_text_field' ),
```
dparker1005, line 308: This `orderby` arg (and the same one on `/subscriptions` below) is registered and sanitized but never read — the callbacks don't forward it into `$query_args`, so `?orderby=total` is silently ignored and results always come back in the default order. `/members` wires it through correctly. Either forward it here too — ideally through an allowlist map like `pmpro_get_members()` uses, since `get_orders()` would otherwise accept arbitrary column names and 500 on typos — or remove the arg from these two registrations so the API doesn't advertise a knob that does nothing.

includes/cli.php
```diff
@@ -0,0 +1,1039 @@
...
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     wp pmpro level list
+	 *
+	 * @when after_wp_load
+	 */
+	public function level_list( $args, $assoc_args ) {
+		$items = array();
+		foreach ( (array) pmpro_getAllLevels( ! empty( $assoc_args['include-hidden'] ), true ) as $level ) {
```
dparker1005, line 422: The second parameter here is `$deprecated_argument` (ignored since 3.0) — `$force` is the third. The static cache is fine for a single CLI invocation, so simplest is to drop it:

```suggestion
		foreach ( (array) pmpro_getAllLevels( ! empty( $assoc_args['include-hidden'] ) ) as $level ) {
```

dparker1005 (commented):
Thanks Flint — the command design here is strong: the `--dry-run`/confirm pattern on every write, no-op detection before prompting, the `--user_id` rename to avoid WP-CLI's reserved `--user`, and the ambiguity guard on `subscription sync` by transaction ID are all exactly right. I ran the read commands and REST endpoints against a local site and the core flows work well.

One ordering note before the inline comments: **#3736 needs to merge first.** The `/pmpro/v1/subscriptions` endpoint calls `PMPro_Subscription::get_subscriptions()` with `return_count`, but that argument doesn't exist on dev — #3736 is the PR that adds it. Until it lands, the count call silently returns an array and the `(int)` cast makes `X-WP-Total` always 0 or 1 (verified locally: `X-WP-Total: 1` with 20 items on the page). Once #3736 merges, this branch will also need a rebase — both PRs add the same `OFFSET` hunks to `get_orders()`/`get_subscriptions()`, so those will conflict.

includes/rest-api.php
ideadude: this PR 3740 expects 3736 to be merged in first. I didn't do it just to keep Mirco with the credit.

includes/functions.php
ideadude: My guess is the desired behavior for people calling this would be to pick the latest status for each member of status=all was passed in. So sort by mu.id DESC. Very unlikely they are like "get me all history for all members" in one API request and expecting that to work.

They might want the full history for a single member. Do we have a good API endpoint for that? I don't think we want them calling this and then using search to limit to a single member.

includes/rest-api.php
ideadude: Nice catch. Core WP and WooCommerce have orderby and order parameters for similar endpoints, so it seems people like to use these. (I can imagine instances where I would appreciate not having to do my own sorting after I get something from the API.) So let's wire them up.

PR #3742, opened 2026-07-27
BUG FIX: Multi-value User Fields blank in UI after CSV import

flintfromthebasement:
## Description
Multi-value User Fields (`checkbox_grouped`, `multiselect`, `select2`, multi `select`) save as PHP arrays from the profile/checkout UI. Members List CSV export joins those arrays with commas (`email_summer,text_consent`). Import Users from CSV writes that cell back as a plain string.

On display, core did:

```php
if ( ! is_array( $value ) ) {
    $value = array( $value ); // "a,b,c" → ["a,b,c"]
}
in_array( $option_key, $value ); // never matches
```

So after import:
- Export still showed the correct comma-separated string
- Profile UI showed no boxes checked
- Re-saving the same selections in the UI stored a real array and the UI lit up

### Fix
- `PMPro_Field::stores_array_values()` — identifies multi-value field types
- `PMPro_Field::get_values_as_array()` — normalizes arrays, serialized arrays, and comma-separated strings into a list of option keys
- Used in `getHTML()` for multi-select field types and in `displayValue()`

Companion PR on Import Users from CSV stores multi-value fields as arrays on import so new imports match the UI save format. This core change is still needed for already-imported data and any other CSV-style integrations.

## Testing
1. Create a `checkbox_grouped` User Field with option keys like `email_summer`, `monthly_newsletter`, `text_consent`
2. Save selections via profile UI — confirm boxes stay checked
3. `update_user_meta( $user_id, 'field_name', 'email_summer,text_consent' );` to simulate CSV import
4. Reload profile — both options should now show checked (previously blank)
5. Confirm Members List export still outputs comma-separated values
6. Single-value string meta (`email_summer`) and real arrays still work

## Related
Found while testing member import on a Hosting Max staging site (LBCA). Not site-specific.

### DIFF
--- a/classes/class-pmpro-field.php
+++ b/classes/class-pmpro-field.php
@@ -654,7 +654,7 @@ class PMPro_Field {
 	/**
 	 * Whether this field stores multiple selected values as an array.
 	 *
-	 * @since TBD
+	 * @since 3.8.3
 	 *
 	 * @return bool
 	 */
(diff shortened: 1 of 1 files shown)


flintfromthebasement:
## Follow-up from review

Addressed the critical finding:

- Removed `maybe_unserialize()` from `get_values_as_array()`. That helper also runs on checkout re-display with request-sourced values — unserializing there was an unauthenticated object-injection surface.
- `get_user_meta()` already unserializes stored arrays; CSV import strings only need explode/trim.
- Set `@since 3.8.3`.

Still intentional:
- Comma-split assumes option keys don't contain commas (same contract as Members List export `join(',', $value)`).
- Export path already handles real arrays via `join` — no change needed there.

dparker1005 (approved):

PR #3756, opened 2026-08-18
Remove display grid from login widget

RachelRVasquez:
### All Submissions:

**Bug:** Form looks squished/broken on narrow widths due to `display: grid` styles that place the "Show Password" button inline, to the right of the field label.

image

**Flint's solution:** https://gist.github.com/flintfromthebasement/41e49dbd50842eb07b48a1da455b796b

**Alternate solution in this PR:** Remove that CSS. Let the form stack elements normally. Tab order now also matches the visual order for accessibility.

image

Styles have been removed from both the `variation_1.css` and `variation_high_contrast.css` files.

**Recommendation for a separate PR:** Adding `aria-pressed` to the "Show Password" button. Reference: https://medium.com/@web-accessibility-education/dos-and-donts-of-accessible-show-password-buttons-9a5fbc2c566b

Screen readers need to know when that "Show Password" button is active. Short video of NVDA reading the password field as protected bullets until you toggle the button so that it can read what's in the password field.

https://www.awesomescreenshot.com/video/55660850?key=8d1093d09b503911bff79f42a55c1cde

1. Easiest to test with Memberlite due to the "Columns Ratio" setting under Appearance > Customize > Posts & Archives.
2. Set the "Columns Ratio" to something really narrow, like 9x3.
3. If your sidebar doesn't already have the login widget, add it.
4. Check the front-end and confirm whether the "Show Password" now shows beneath the password field and the password label is no longer squished/broken.

### Other information:

### Changelog entry

> Remove display grid styles on login form widget. "Show Password" toggle now displays beneath the password field. Resolves issue on narrow screens where the "Show Password" was crashing into the label and breaking it.

### DIFF
--- a/css/frontend/variation_1.css
+++ b/css/frontend/variation_1.css
@@ -1202,11 +1202,6 @@
 		.pmpro_cols-3 > * {
 			width: auto;
 		}
-
-		#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle) {
-			width: 100%;
-		}
-
 	}
 
 }
--- a/css/frontend/variation_1.css
+++ b/css/frontend/variation_1.css
@@ -656,33 +656,6 @@
 		opacity: .7;
 	}
 
-	/* Move password toggle before field visually but structurally after */
-	#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle),
-	.pmpro_section #loginform .login-password {
-		align-items: center;
-		display: grid;
-		grid-template-areas:
-			"label toggle"
-			"input input";
-		grid-template-columns: 1fr auto;
-	}
-
-	#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle) label,
-	.pmpro_section #loginform .login-password label {
-		grid-area: label;
-	}
-
-	#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle) input,
-	.pmpro_section #loginform .login-password input {
-		grid-area: input;
-	}
-
-	#pmpro_user_fields .pmpro_form_field-password .pmpro_form_field-password-toggle,
-	.pmpro_section #loginform .login-password .pmpro_form_field-password-toggle {
-		grid-area: toggle;
-		justify-self: end;
-	}
-
 	#resetpassform .pmpro_cols-2 {
 		container: resetpassform / inline-size;
 	}
--- a/css/frontend/variation_high_contrast.css
+++ b/css/frontend/variation_high_contrast.css
@@ -651,33 +651,6 @@
 		border-radius: 0;
 	}
 
-	/* Move password toggle before field visually but structurally after */
-	#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle),
-	.pmpro_section #loginform .login-password {
-		align-items: center;
-		display: grid;
-		grid-template-areas:
-			"label toggle"
-			"input input";
-		grid-template-columns: 1fr auto;
-	}
-
-	#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle) label,
-	.pmpro_section #loginform .login-password label {
-		grid-area: label;
-	}
-
-	#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle) input,
-	.pmpro_section #loginform .login-password input {
-		grid-area: input;
-	}
-
-	#pmpro_user_fields .pmpro_form_field-password .pmpro_form_field-password-toggle,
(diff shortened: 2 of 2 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3756 — Remove display grid from login widget**
*RachelRVasquez → dev | 2 files, +0 -54*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3756>

---

**Summary**
Solid direction — removing the grid layout (rather than patching around it) is the right fix and improves keyboard/tab order. Not quite ready to merge: the removed selector is shared with the checkout account-fields password toggle, which isn't mentioned in the PR's testing steps, so that surface needs a look before this goes in.

---

**Issues**

- **[Major]** `css/frontend/variation_1.css:659-684`, `css/frontend/variation_high_contrast.css:654-679` — The deleted grid block targets two selectors, not one: `.pmpro_section #loginform .login-password` (the login widget) **and** `#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle)`. `#pmpro_user_fields` is the checkout "Account Information" fieldset (`pages/checkout.php:241`), which renders the same password-toggle markup (`pages/checkout.php:275`). This rule isn't inside a media query in either file, so removing it changes the checkout password field's layout at every screen width, not just narrow ones — the toggle moves from beside the label to below the input there too. The PR description and test steps only cover the login widget via Memberlite's narrow-column layout; the checkout page isn't mentioned. Worth a quick check on `/checkout/` before merge to confirm the new stacked layout there is intentional/acceptable, since it's a wider-reaching change than the PR describes.

- **[Minor]** `css/frontend/variation_1.css:1233`, `css/frontend/variation_high_contrast.css:1216` — The `@media (max-width: 767px)` rule `#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle) { width: 100%; }` still references the now-grid-less selector. With the grid removed, `.pmpro_form_field` already lays out as `display: flex; flex-direction: column` (`variation_1.css:299`), so the field is full-width by default and this rule is dead weight. Not harmful, fine to leave for a follow-up cleanup.

---

**Looks Good**
- Both variation files (`variation_1.css`, `variation_high_contrast.css`) are edited identically — no drift between the standard and high-contrast stylesheets.
- Removing the grid restores natural DOM/tab order matching visual order, a real accessibility win over the prior layout.
- `.pmpro_section #loginform .login-password` (`variation_1.css:568-569`) and the base `.pmpro_form_field` class both already declare `display: flex; flex-direction: column`, so neither the login widget nor the checkout field is left unstyled after the grid rules are removed — it falls back cleanly to a stacked layout rather than collapsing.
- This supersedes an earlier narrower fix (commit `5e3d182e9`, which hid the toggle text in narrow containers as a workaround) with a root-cause removal instead of another patch on top of the grid.
- Correctly scopes out the `aria-pressed` accessibility improvement into a separate follow-up PR rather than mixing concerns.

---

**Questions**
1. Was the checkout page's password field (`#pmpro_user_fields`, "Account Information" fieldset) tested at any width? The removed CSS applies there unconditionally, not just on narrow login widgets.

RachelRVasquez:
**Responding to Flint here:**

Checkout looks good on desktop and mobile. Screenshots below from my local w/ Memberlite:

image

image

The `@media (max-width: 767px)` rule `#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle) { width: 100%; }` has been removed from both stylesheets.

flintfromthebasement (approved):
**Follow-up review**

**PR: strangerstudios/paid-memberships-pro#3756 — Remove display grid from login widget**
*RachelRVasquez → dev | 2 files, +0 -64*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3756>

**Summary**
Both prior issues are resolved — this is ready to merge. Rachel verified the checkout password field on desktop and mobile with Memberlite screenshots, and removed the dead media-query rule I'd flagged as leftover cruft.

**Prior Issues — Status**

- **[Major]** `#pmpro_user_fields` grid removal affecting checkout — **Resolved.** Rachel tested checkout on desktop and mobile locally with Memberlite and posted screenshots showing correct stacked layout with the password toggle rendering properly. No regression.
- **[Minor]** Dead `@media (max-width: 767px)` rule referencing the now-grid-less selector — **Resolved.** Commit `d2129a26` removes `#pmpro_user_fields .pmpro_form_field-password:has(.pmpro_form_field-password-toggle) { width: 100%; }` from both `variation_1.css` and `variation_high_contrast.css`, exactly as suggested.

**New Findings**
None. The delta is a clean two-line-block removal in each file, matching what was already discussed — no new risk introduced.

**Looks Good**
Rachel closed the loop with visual proof (both breakpoints) rather than just asserting it works, and cleaned up the dead rule same-session instead of leaving it for a follow-up. Good final state.

PR #3757, opened 2026-08-18
Review follow-ups for Abilities API and pmpro-cli (v3.9)

dparker1005:

Follow-ups from the review threads on #3736 (Abilities API) and #3740 (pmpro-cli + REST collections), incorporating @ideadude's feedback on each point. Both PRs are now merged into `v3.9`; this branch sits on top of that merge.

**From the #3736 review:**

- `pmpro/member-memberships-get` now returns **active memberships by default** and adds an `include_inactive` boolean input for full history ("option 1" per Jason). The change/cancel abilities' `previous_memberships`/`current_memberships` snapshots are also active-only now. Behavior change is called out in the changelog entry below.
- `old_level_status` on `pmpro/member-membership-change` is now validated with `pmpro_sanitize_with_safelist()` against the statuses core itself passes (`inactive`, `changed`, `admin_changed`, `cancelled`, `admin_cancelled`, `expired`); invalid values return a 400 `WP_Error` before any mutation. Shipped without a new statuses-helper function per Jason's "okay to ship as is."

**From the #3740 review:**

- Fixed `--format=ids` on `level list`, `order list`, and `subscription list` (rows are now plucked to a flat ID list before formatting).
- `pmpro_get_memberships()` now joins only the **latest row matching the status filter for each user/level pair** (via a derived `MAX(id)` table), so output is deterministic per Jason's suggestion. `status=active` deliberately matches anyone PMPro considers to hold the level (any active row grants access via `pmpro_hasMembershipLevel()`), and historical statuses match rows like the underlying table does. Implemented as a single-pass derived-table join rather than a correlated subquery — the correlated version hit a pathological plan on a 38k-row table locally (120s+ vs 0.06s).
- Removed the `pmpro_get_members_sql` filter (no callers anywhere; fragile prepared-SQL contract).
- Wired up `orderby`/`order` on the `/pmpro/v1/orders` and `/pmpro/v1/subscriptions` endpoints with allowlisted column maps, matching `/members`. Unknown `orderby` values fall back to the default sort.
- `level list` (CLI) and the levels-query ability no longer pass the deprecated second argument to `pmpro_getAllLevels()`.
- **Renamed the memberships collection surfaces** per Slack discussion with Jason and Mirco: `GET /pmpro/v1/members` is now `GET /pmpro/v1/memberships`, `wp pmpro member list` is now `wp pmpro membership list`, and the shared query function is `pmpro_get_memberships()`. All three return one row per user/level pair, so the memberships name describes the row shape accurately (a member holding two levels appears once per level). Person-scoped surfaces (`wp pmpro member get|add-level|remove-level|change-level|cancel`, the `pmpro/member-*` abilities) keep the member noun since they act on a user. None of these had shipped, so no back-compat concern.
- Normalized the `@since 3.8.3` placeholders from #3736 to `@since TBD`.
- **From a final worktree review of this branch:** fixed `status=all` on the memberships collection (the REST endpoint passes statuses as an array, so `array( 'all' )` was treated as a literal status and matched nothing — now returns all rows via both the string and array paths, verified against live data), and reverted an over-greedy rename that had corrupted four legacy REST callback method names.

All verified by execution on a local site (WP 7.1, 38k-row memberships table):

1. `pmpro_abilities_execute_member_memberships_get()` for a user with 106 historical rows: returns 2 (active) by default, 106 with `include_inactive: true`.
2. `member-membership-change` with `old_level_status: "active"` → `WP_Error pmpro_abilities_invalid_old_level_status`, no data changed.
3. `wp pmpro level list --format=ids` → flat ID list, no warnings.
4. `pmpro_get_members( [ 'status' => 'all' ] )` for a pair with 36,959 duplicate rows: exactly one row, matching the highest-ID row's status/startdate; counts match row counts for both `all` (1094) and `active` (452); full query in 0.06s.
5. `/pmpro/v1/orders?orderby=total&order=asc|desc` sorts correctly; `orderby=evil; DROP` safely falls back to the default sort. `/pmpro/v1/subscriptions?orderby=id&order=asc` returns 1,2,3,4,5 with correct `X-WP-Total`.

### Open questions (flagging for discussion, not blocking):

1. ~~Per-user history on the REST side?~~ **Resolved: starting minimal.** The `pmpro/member-memberships-get` ability (`include_inactive: true`) and `wp pmpro member get <id> --include-inactive` cover single-member history; no new REST endpoint or filter in this PR.
2. ~~"Latest per pair" vs "latest per member"~~ **Resolved:** one row per user+level pair, and the query function is now named `pmpro_get_memberships()` so the row shape is self-describing (a member holding two levels returns two membership rows). Concurrent MMPU levels stay visible.
3. **Unknown `orderby` values silently fall back** to the default sort (matching `/members`). The alternative is a 400 like the `fields` param. Happy to switch if we prefer loud failures here.

### Changelog entry

> BUG FIX/ENHANCEMENT: Review follow-ups for the Abilities API and pmpro-cli features. The pmpro/member-memberships-get ability now returns active memberships by default (pass include_inactive to get full history), membership change abilities validate the old level status, the memberships query now returns one deterministic row per user/level pair, --format=ids works on CLI list commands, the orders/subscriptions REST endpoints support orderby/order, and the memberships collection is exposed as /pmpro/v1/memberships and wp pmpro membership list.

### DIFF
--- a/classes/class-pmpro-subscription.php
+++ b/classes/class-pmpro-subscription.php
@@ -315,7 +315,7 @@ class PMPro_Subscription {
 	 *
 	 * Defaults to returning the latest 100 subscriptions.
 	 *
-	 * @since 3.8.3 Added support for the `offset` and `return_count` arguments.
+	 * @since TBD Added support for the `offset` and `return_count` arguments.
 	 *
 	 * @param array $args The query arguments to use.
 	 *
--- a/classes/class.memberorder.php
+++ b/classes/class.memberorder.php
@@ -555,7 +555,7 @@
 		 * Get orders based on various parameters
 		 *
 		 * @since 2.9
-		 * @since 3.8.3 Added support for the `offset` argument.
+		 * @since TBD Added support for the `offset` argument.
 		 *
 		 * @param array $args Specify what you'd like to filter the query by
 		 *
--- a/includes/abilities.php
+++ b/includes/abilities.php
@@ -3,7 +3,7 @@
  * Register Paid Memberships Pro abilities for the WordPress Abilities API.
  *
  * @package PaidMembershipsPro
- * @since 3.8.3
+ * @since TBD
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
@@ -16,7 +16,7 @@ add_action( 'wp_abilities_api_init', 'pmpro_abilities_register_abilities' );
 /**
  * Check whether the WordPress Abilities API is available.
  *
- * @since 3.8.3
+ * @since TBD
  *
  * @return bool
  */
@@ -29,7 +29,7 @@ function pmpro_abilities_is_abilities_api_available() {
 /**
  * Check whether PMPro can register abilities.
  *
- * @since 3.8.3
+ * @since TBD
  *
  * @return bool
  */
@@ -37,7 +37,7 @@ function pmpro_abilities_can_boot() {
 	/**
 	 * Filter whether PMPro should register Abilities API categories and abilities.
 	 *
-	 * @since 3.8.3
+	 * @since TBD
 	 *
 	 * @param bool $enabled Whether ability registration is enabled.
 	 */
@@ -308,15 +308,21 @@ function pmpro_abilities_get_levels_query_definition() {
 function pmpro_abilities_get_member_memberships_definition() {
 	return array(
 		'label'               => __( 'Get Member Memberships', 'paid-memberships-pro' ),
-		'description'         => __( 'Retrieve the current membership assignments for a specific user.', 'paid-memberships-pro' ),
+		'description'         => __( 'Retrieve the active membership assignments for a specific user. Set include_inactive to also return the user\'s past membership history.', 'paid-memberships-pro' ),
 		'category'            => 'pmpro',
 		'execute_callback'    => 'pmpro_abilities_execute_member_memberships_get',
 		'permission_callback' => 'pmpro_abilities_can_manage_members',
 		'input_schema'        => array(
 			'type'       => 'object',
+			'default'    => array(
+				'include_inactive' => false,
+			),
 			'properties' => array(
-				'user_id' => array(
+				'user_id'          => array(
 					'type' => 'integer',
+				),
+				'include_inactive' => array(
+					'type' => 'boolean',
 				),
 			),
 			'required'   => array( 'user_id' ),
@@ -912,7 +918,7 @@ function pmpro_abilities_execute_levels_query( $input ) {
 	$query          = strtolower( trim( (string) $input['query'] ) );
 	$limit          = max( 1, min( 50, (int) $input['limit'] ) );
 	$page           = max( 1, (int) $input['page'] );
-	$levels         = array_values( pmpro_getAllLevels( $include_hidden, true ) );
+	$levels         = array_values( pmpro_getAllLevels( $include_hidden ) );
 
 	if ( '' !== $query ) {
 		$levels = array_values(
@@ -953,7 +959,9 @@ function pmpro_abilities_execute_member_memberships_get( $input ) {
 		return $user;
 	}
 
-	$memberships = pmpro_getMembershipLevelsForUser( $user->ID, true );
+	$include_inactive = ! empty( $input['include_inactive'] );
+
+	$memberships = pmpro_getMembershipLevelsForUser( $user->ID, $include_inactive );
 	if ( ! is_array( $memberships ) ) {
 		$memberships = array();
 	}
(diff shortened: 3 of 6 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3757 — Review follow-ups for Abilities API and pmpro-cli (v3.9)**
*dparker1005 → v3.9 | 6 files, +88 -40*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3757>

---

**Summary**
Solid follow-up PR with one correctness gap worth fixing before merge: the new deterministic `pmpro_get_members()` join still filters by status *inside* the "latest row" subquery, so status-filtered queries (including the default `status=active`) can return a pair's last row that matched the filter rather than its actual latest row. Everything else — the `old_level_status` safelist, `--format=ids`, REST `orderby`/`order` allowlisting, and the active-only membership snapshots — checks out against the code.

---

**Issues**

- **[Major]** `includes/functions.php` (derived-table join, ~5612-5636) — `$status_condition` is applied inside the `WHERE` of the `MAX(mu2.id)` subquery, so for a non-`'all'` status the query picks the highest-`id` row *among rows already matching that status*, not the pair's truly latest row. Example: a user/level pair has an older `active` row and a newer `cancelled` row. Calling with the default `status => 'active'` (the function's own default, and the default for `/pmpro/v1/members` and `wp pmpro member list`) matches the older active row inside the subquery, joins it back in, and reports that pair as active with a stale `startdate` — the user is actually canceled. The PR's own description says the join now returns "the latest row per user/level pair," but that's only true for `status=all`; for any other status the semantics are unchanged from the pre-PR `WHERE ... GROUP BY` version (same issue existed there too, just nondeterministically), so this fix doesn't actually close the gap for the common case. Test case 4 in the PR description only exercised `status=all` and `status=active` counts on data that apparently didn't include an active→cancelled transition for the same pair, so it wouldn't have caught this. Fix: compute `MAX(mu2.id)` over `mu2.membership_id > 0` only (no status predicate) to get the true latest row per pair, join that, then apply the status filter to the outer `mu.status` (or drop non-matching rows after the join) instead of inside the subquery.

- **[Minor]** `includes/functions.php:5626` (`$where[] = 'mu.membership_id > 0';`) — now redundant. The derived subquery already restricts to `mu2.membership_id > 0` before grouping, and `latest.id = mu.id` guarantees the joined `mu` row has the same `membership_id`. Harmless, but dead condition; safe to drop as part of the same cleanup.

---

**Looks Good**

- `old_level_status` validation (`abilities.php`): `pmpro_sanitize_with_safelist()` returns `false` on no match (confirmed at `functions.php:4035-4041`), `empty(false)` is true, so an invalid value returns a 400 `WP_Error` before any mutation. The safelist (`inactive`, `changed`, `admin_changed`, `cancelled`, `admin_cancelled`, `expired`) matches real call sites — `'changed'` is passed at `services/ipnhandler.php:598` and `includes/checkout.php:288`, and `'admin_changed'` is handled explicitly in `pmpro_changeMembershipLevel()` at `functions.php:1387`.
- Snapshotting `previous_memberships`/`current_memberships` as active-only in `member_membership_change`/`member_membership_cancel` (dropping the `true` argument) is a real improvement — the old code passed `include_inactive: true` on both snapshots, so a cancel's before/after would have shown identical historical rows either way.
- REST `orderby`/`order` on `/orders` and `/subscriptions`: both sanitized with `sanitize_key` and resolved through hardcoded column maps before touching SQL — an unrecognized value never reaches the query string, it just falls through to the existing default sort.
- `--format=ids` fix (`cli.php`) correctly plucks `$fields[0]` before formatting; `id` is the first field in all three affected formatters.
- `pmpro_getAllLevels()` second-argument removal is a clean drop of a deprecated, unused parameter.

---

**Questions**

1. Given the status-filter issue above, should the fix also settle open question #2 (latest-per-pair vs. latest-per-member)? They're related but distinct — worth confirming the intended semantics once the subquery is corrected.
2. Silent orderby fallback vs. a 400: keeping it consistent with `/members`' existing behavior seems reasonable, but flagging since the PR description raises it as open.

flintfromthebasement (approved):
**Follow-up review**

**PR: strangerstudios/paid-memberships-pro#3757 — Review follow-ups for Abilities API and pmpro-cli (v3.9)**
*dparker1005 → v3.9*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3757>

**Summary**
Clear to merge. My prior Major (status filter applied inside the latest-row subquery) does not hold up — I traced the actual write paths and the failure mode I described can't occur. The rest of the delta (function/route rename to `pmpro_get_memberships`/`/memberships`, `status=all`-as-array fix, restored REST callback names) is clean.

**Prior issue: status filter inside `MAX(mu2.id)` subquery — withdrawn**
`includes/functions.php:5636` still applies `$status_condition` inside the derived-table subquery, unchanged in shape from what I flagged last time. But I verified the write paths this time instead of reasoning from a hypothetical DB state: `pmpro_cancelMembershipLevel()` (`functions.php:1237`) does an in-place `UPDATE ... WHERE status = 'active'` on the same row (same `id`) rather than inserting a new one; `pmpro_changeMembershipLevel()`'s insert (`functions.php:1421`) never sets `status` explicitly so it takes the column default (`'active'`, confirmed in the `CREATE TABLE` at `includes/upgradecheck.php:616`), and its dedup step (`functions.php:1479`) force-flips any other `active` row for the same user/level pair to `changed` right after insert. Net effect: under normal PMPro flows there is never more than one `active` row per user/level pair, so "latest row matching the filter" and "true latest row" coincide for the case I described (stale active row + newer cancelled row, same pair). The scenario I raised requires pre-existing dirty data from outside these code paths, not a bug in this query — that's a data-hygiene edge case, not something this PR introduced or needs to fix. The docblock update in `8e91fd020` ("historical statuses such as 'cancelled' can match members who since re-subscribed... 'active' matches anyone PMPro considers to hold the level") is accurate. Conceding this one — my original repro didn't survive verification against the write paths, only against a diff read.

**Minor (unaddressed, still low priority):** `includes/functions.php` — `$where[] = 'mu.membership_id > 0';` is still redundant (the subquery already restricts to `mu2.membership_id > 0`). Harmless, not worth blocking on.

**New changes since my review — clean**
- `pmpro_get_members()` → `pmpro_get_memberships()`, CLI `member list` → `membership list`, REST `/members` → `/memberships`: renamed consistently. Grepped the full repo for `pmpro_get_members\b`, old REST callback names, and `member_list\b` — no stale references left anywhere.
- `7b556e796` fixed a real bug introduced mid-PR: an overly greedy rename had corrupted the REST callback method names to `pmpro_rest_api_get_membershipship_*`; confirmed the string no longer appears anywhere in `rest-api.php`.
- `status => array('all')` on the memberships collection now works — the REST endpoint passes statuses as an array, and the old `'all' !== $args['status']` check only matched a bare string. Now correctly checks `in_array('all', $statuses, true)` after normalizing to an array.
- `old_level_status` rejection message now lists valid values (`abilities.php`), matching the existing `fields` param error format — good self-correction signal for API/MCP callers.

**Looks Good**
Good instinct pushing back via docblock rather than restructuring a query that was already correct for its actual callers — and good catch on the `status=all`-array and rename-corruption bugs in the same pass.
Verdict: APPROVE.

Issue #3640, opened 2026-04-14
Spam protection: race condition in transient-based tracking allows brute force bypass


flintfromthebasement:
## Problem

The spam activity tracking in `includes/spam.php` uses a read-modify-write pattern on transients that is not atomic:

```php
// pmpro_track_spam_activity()
$activity = pmpro_get_spam_activity( $ip );     // 1. READ transient (array of timestamps)
array_unshift( $activity, $now );                // 2. MODIFY (add timestamp)
set_transient( $transient_key, $activity, ... ); // 3. WRITE back
```

Under concurrent requests from the same IP, the last writer wins and earlier additions are silently lost. This means the activity count for an IP never reaches `PMPRO_SPAM_ACTION_NUM_LIMIT` (10) under sustained attack.

### Real-world impact

On paidmembershipspro.com (2026-04-14), a single IP (`193.142.147.5`) sent **64,743 POST requests** to `/wp-login.php` over 19 hours. PMPro's spam transient for that IP only ever recorded **6 attempts** — the counter never reached 10 to trigger a block. Every request got HTTP 200.

The race condition is worse on sites using MyISAM for `wp_options` (table-level locking causes reads to queue, then writes all fire with stale data), but it also affects InnoDB sites under sufficient concurrency.

### Secondary issue: transient bloat

Each unique IP that triggers spam tracking creates two rows in `wp_options` (`_transient_pmpro_spam_activity_{ip}` + `_transient_timeout_pmpro_spam_activity_{ip}`). On paidmembershipspro.com, this accumulated **28,975 transient rows** from brute force attempts across thousands of IPs. These persist until they expire or are manually cleaned up, bloating the options table.

## Proposed Solution

Replace the transient-based array with a custom table where each spam event is an individual row. `INSERT` is always atomic — no read-modify-write, no race condition.

### Schema

```sql
CREATE TABLE {$wpdb->prefix}pmpro_spam_activity (
    id BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
    ip VARCHAR(45) NOT NULL,
    created DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
    INDEX idx_ip_created (ip, created)
);
```

### Track (replaces `pmpro_track_spam_activity`)

```php
$wpdb->insert(
    $wpdb->prefix . 'pmpro_spam_activity',
    array( 'ip' => $ip ),
    array( '%s' )
);
```

[...]

PR #3493, opened 2025-10-01
reCAPTCHA & Turnstile refactor/improvements.

andrewlimaza:
Improvements to CAPTCHA placement for reCAPTCHA and CloudFlare Turnstile:

* Reworked settings page to choose between captchas, instead of being able to use both at the same time.
* Implemented captchas to load on the default wp-login.php and PMPro's frontend login forms. The captcha will only show after a single failed login attempt - this helps prevent site owners from getting locked out if incorrect keys are used or keys do not match domains (and any other failure to load the captchas).
* Added new generic 'track failed login attempt' to trigger the above captcha to load, and clears up after successful actions are made.

This sets the foundation for us to create more integrations into other captchas or custom validation practices needed.

**Example of new settings UI update**
Screenshot 2025-10-01 at 12 58 45

**Examples of frontend login captcha V2 (for demonstrative purposes)**

Screenshot 2025-10-01 at 13 06 08

Screenshot 2025-10-01 at 13 06 20

PR #3765, opened 2026-08-21
Captcha on login and password reset forms (login hooks + single captcha setting)

dparker1005:
Phases 1 and 2 of the captcha-on-login work discussed for 3.9 (replaces the approach in #3493; phase 3 is the spam activity table in #3640).

## Phase 1: Generic login flow hooks (commit 1, `includes/login.php` only)

Gives captchas (and anything else) clean hook points in the login flow, mirroring how checkout isolates captcha code:

* **`pmpro_login_form_before_submit_button`** — fires inside the PMPro login form, via a `login_form_middle` filter added/removed around `wp_login_form()` (same pattern as the existing hidden-field filter), so it only ever runs on PMPro's forms.
* **`pmpro_authenticate_login_checks`** — filter on `authenticate` at priority 40, gated to PMPro-form submissions (`pmpro_login_form_used`) or real wp-login.php login submissions (`did_action( 'login_form_login' )`). XML-RPC, other plugins' login forms, and programmatic `wp_signon()` calls are never touched, so a captcha can't lock out a flow that didn't display it.
* **`pmpro_login_forms_handler_message` / `pmpro_login_forms_handler_msgt`** — lets custom error codes passed back to the login page display messages.
* Cleanup: `pmpro_authenticate_username_password()` now only redirects for `empty_username`/`empty_password` (the codes WP core never fires `wp_login_failed` for) and exits after redirecting. All other failures flow through `pmpro_login_failed()` — previously both functions called `wp_redirect()` in the same request and the second one happened to win.

## Phase 2: Captchas use the hooks (commit 2)

* **Single "Captcha" setting** (No / Google reCAPTCHA / Cloudflare Turnstile) on the Security Settings page replaces the two independent on/off settings, so only one captcha runs at a time. Back compat: the old options are read when the new setting has never been saved (a saved "No" does not fall back), and the old options are kept in sync on save for anything that reads them (e.g. Wisdom telemetry).
* **Captcha service registry**: new `pmpro_captcha_services` filter in `includes/captcha.php`. reCAPTCHA and Turnstile register themselves from their own files, and a future captcha add on can drop in the same way with no core changes. If a registered service disappears (add on deactivated), the site fails safe to no captcha.
* **Captcha on login and lost password forms** (PMPro pages + wp-login.php), shown **only after suspicious activity from the visitor's IP** — a failed login attempt while a captcha is enabled, or failed checkouts/invalid discount codes when Spam Protection is also on — and only when keys are configured, so a misconfigured captcha can't lock admins out. The signal is the existing spam activity API, which phase 3 (#3640) will re-point at the new table with no changes here. It is deliberately not cleared on successful login (on shared IPs that would let attackers reset it); it expires with the tracking window.
* Validation is fully server-side. reCAPTCHA v2 and Turnstile need no custom JS on login forms; v3-invisible uses a small inline script that fails open to server-side rejection if Google's script doesn't load. The checkout JS files are untouched.
* Failed logins are now tracked when a captcha is enabled even if Spam Protection is off.
* **Bug fix found along the way**: Turnstile checkout validation set the "validated" session var even when verification *failed*, so a failed captcha was bypassed on retry. Now only successful validations are remembered.

## Review follow-ups (commit 3)

Addressing @flintfromthebasement's review:

* **reCAPTCHA lib hardening**: `pmpro_ReCaptcha::_submitHTTPPost()` indexed `wp_remote_post()`'s return as an array, which fatals on PHP 8 when a network error to Google returns a `WP_Error`. Pre-existing on checkout, but this PR made it reachable from login — now guarded so verification fails safely (the user sees a captcha error and can retry) instead of a 500.
* **Shared spam signal documented as intentional**: the challenge trigger is a broader "recent suspicious activity" signal, not strictly failed logins — renamed `pmpro_captcha_has_recent_failed_login()` to `pmpro_captcha_has_recent_spam_activity()`, documented the semantics (including why it isn't cleared on successful login), and updated the setting description.
* **Turnstile login output** got the same show-once static guard the reCAPTCHA output already had.
* Not changed: the `action`/`errors`/`error` param handling in `pmpro_is_captcha_failed_request()` — the review confirmed the flow is correct; the code only ever emits the fixed `pmpro_captcha_failed` code. Verified against WP core that `lostpassword`/`retrievepassword` are the only wp-login.php lost-password action aliases, and any hypothetical third path would fail open (captcha skipped), not lock anyone out.

## Deep review sweep (commit 4)

A multi-agent adversarial review pass (5 dimensions, every finding independently verified against plugin + WP core source) surfaced four real issues, all fixed:

* **Third-party `wp_login_form()` forms were validated but never displayed the captcha.** Any form built with `wp_login_form()` posts to wp-login.php and trips the validation gate, but the widget only rendered on PMPro forms and wp-login.php itself — a theme header/popup login form would hard-fail for a challenged IP. Fixed by rendering through the core `login_form_middle` filter (late priority), which covers exactly the set of forms the gate validates. Narrowing the gate instead was rejected: any request-derived narrowing (e.g. requiring `testcookie`) would let bots bypass the captcha by omitting a field.
* **Only the first login form on a page got a widget while every form was validated** (e.g. a login block in a header template part + the login page form = an unsolvable challenge on one of them). Now every form gets its own widget; only shared script assets are deduped once per page.
* **reCAPTCHA v3-invisible JS hardening**: explicit per-widget rendering with tracked widget IDs (a bare `grecaptcha.execute()` could execute another plugin's widget or throw with no fallback, deadening the submit button when keys were invalid), single API script load reused if already present, and fail-open-to-server-rejection on any render/execute failure. v2 uses `defer` so all containers exist before auto-render.
* **Page caching**: the frontend login page now sends nocache headers while a captcha is enabled, so a cached captcha-less copy can't be served to an IP that will be required to solve one.

Found along the way: the **PMPro Magic Login add-on** echoes its `login_form_middle` output and returns null, wiping every other callback's content on that filter — core PMPro now defends against this (late priority + string cast), but the add-on needs its own one-line fix.

Confirmed-but-deferred (not this PR's scope): the challenge signal keys off `pmpro_get_ip()`, which trusts spoofable forwarded headers — an attacker rotating `X-Forwarded-For` per request evades tracking (and the pre-existing spam block) entirely. That's a property of the existing spam infrastructure shared by all its consumers; the fix (trusted-proxy-aware IP resolution) belongs in the #3640 table work.

## Testing done

Full curl-based pass on a local site: hook plumbing proven with a throwaway consumer; login/logout/lost password/reset password behavior verified unchanged with captcha off; with dummy keys, challenge lifecycle verified for both services on all four forms (no captcha before a failure, widget after, correct-password-but-no-token rejected with a displayed error, lost password blocked without sending email); checkout gating verified for both services incl. legacy-option derivation; registry verified with a fake third-party service (registers, saves, fails safe when removed); reCAPTCHA verify exercised over real HTTP with an invalid token (graceful failure, no fatal).

Still needs a browser pass with real keys: actual successful token round trips, the v3-invisible inline JS, and the settings show/hide toggling.

### DIFF
--- a/adminpages/securitysettings.php
+++ b/adminpages/securitysettings.php
@@ -174,7 +174,7 @@
 					'type'        => 'select',
 					'value'       => pmpro_captcha(),
 					'options'     => array( '' => __( 'No', 'paid-memberships-pro' ) ) + pmpro_get_captcha_services(),
-					'description' => __( 'Protect your checkout, login, and password reset forms with a captcha challenge. On login and password reset forms, the captcha is only shown after a failed login attempt from the visitor\'s IP address.', 'paid-memberships-pro' ),
+					'description' => __( 'Protect your checkout, login, and password reset forms with a captcha challenge. On login and password reset forms, the captcha is only shown after a failed login attempt or other suspicious activity from the visitor\'s IP address.', 'paid-memberships-pro' ),
 				),
 				array(
 					// The callbacks hooked here echo their own <tr> rows, so give them a table.
--- a/adminpages/securitysettings.php
+++ b/adminpages/securitysettings.php
@@ -18,6 +18,16 @@
 	// Save settings.
 	if( !empty( $_REQUEST['savesettings'] ) ) {
 		pmpro_setOption( "spamprotection", intval( $_POST['spamprotection'] ) );
+
+		// Save the captcha setting. Note: This must be saved before the
+		// pmpro_save_security_settings hook fires so that the captcha services
+		// saving their settings on that hook can see the updated value.
+		$captcha = isset( $_POST['captcha'] ) ? sanitize_text_field( $_POST['captcha'] ) : '';
+		if ( ! array_key_exists( $captcha, pmpro_get_captcha_services() ) ) {
+			$captcha = '';
+		}
+		pmpro_setOption( 'captcha', $captcha );
+
 		if ( isset( $_POST['use_ssl'] ) ) {
 			// REQUEST['use_ssl'] will not be set if the entire site is already over HTTPS.
 			pmpro_setOption( "use_ssl", intval( $_POST['use_ssl'] ) );
@@ -157,6 +167,14 @@
 						2 => __( 'Yes - Enable Spam Protection', 'paid-memberships-pro' ),
 					),
 					'description' => sprintf( esc_html__( 'Block IPs from checkout and login if there are more than %d failures within %d minutes.', 'paid-memberships-pro' ), (int) PMPRO_SPAM_ACTION_NUM_LIMIT, (int) round( PMPRO_SPAM_ACTION_TIME_LIMIT / 60, 2 ) ),
+				),
+				array(
+					'name'        => 'captcha',
+					'label'       => __( 'Captcha', 'paid-memberships-pro' ),
+					'type'        => 'select',
+					'value'       => pmpro_captcha(),
+					'options'     => array( '' => __( 'No', 'paid-memberships-pro' ) ) + pmpro_get_captcha_services(),
+					'description' => __( 'Protect your checkout, login, and password reset forms with a captcha challenge. On login and password reset forms, the captcha is only shown after a failed login attempt from the visitor\'s IP address.', 'paid-memberships-pro' ),
 				),
 				array(
 					// The callbacks hooked here echo their own <tr> rows, so give them a table.
--- a/adminpages/securitysettings.php
+++ b/adminpages/securitysettings.php
@@ -26,7 +26,7 @@
 		if ( ! array_key_exists( $captcha, pmpro_get_captcha_services() ) ) {
 			$captcha = '';
 		}
-		pmpro_setOption( 'captcha', $captcha );
+		update_option( 'pmpro_captcha', $captcha, false );
 
 		if ( isset( $_POST['use_ssl'] ) ) {
 			// REQUEST['use_ssl'] will not be set if the entire site is already over HTTPS.
--- a/includes/captcha.php
+++ b/includes/captcha.php
@@ -99,6 +99,35 @@ function pmpro_captcha_failed_error_message() {
 }
 
 /**
+ * Don't allow the frontend login page to be full-page cached while a captcha
+ * service is enabled.
+ *
+ * The captcha only renders for IPs with recent suspicious activity, so a
+ * cached copy of the page could serve a captcha-less form to a visitor whose
+ * submission would then be rejected.
+ *
+ * @since TBD
+ */
+function pmpro_captcha_nocache_login_page() {
+	// Bail if no captcha is enabled.
+	if ( empty( pmpro_captcha() ) ) {
+		return;
+	}
+
+	// Bail if we're not on the frontend login page.
+	$login_page_id = get_option( 'pmpro_login_page_id' );
+	if ( empty( $login_page_id ) || ! is_page( (int) $login_page_id ) ) {
+		return;
+	}
+
+	if ( ! defined( 'DONOTCACHEPAGE' ) ) {
+		define( 'DONOTCACHEPAGE', true );
+	}
+	nocache_headers();
+}
+add_action( 'template_redirect', 'pmpro_captcha_nocache_login_page' );
+
+/**
  * Check whether the current request includes a pmpro_captcha_failed error code
  * passed back to the login page in the URL.
  *
--- a/includes/captcha.php
+++ b/includes/captcha.php
@@ -68,16 +68,21 @@ function pmpro_captcha() {
 }
 
 /**
- * Check whether the current IP has recent failed login activity and should be
+ * Check whether the current IP has recent suspicious activity and should be
  * shown a captcha challenge on login and password reset forms.
  *
- * Uses the spam activity tracking in includes/spam.php.
+ * Uses the spam activity tracking in includes/spam.php. Failed logins always
+ * feed this signal while a captcha is enabled. If Spam Protection is also
+ * enabled, failed checkouts and invalid discount codes feed it too — that is
+ * intentional: an IP abusing checkout is also worth challenging at login.
+ * Activity is deliberately not cleared on a successful login, since on shared
+ * IPs that would let an attacker reset the signal; it expires on its own.
  *
  * @since TBD
  *
- * @return bool True if the current IP has recent failed login activity.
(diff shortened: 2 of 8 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3765 — Captcha on login and password reset forms (login hooks + single captcha setting)**
*dparker1005 → v3.9 | 7 files, +631 -70*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3765>

---

**Summary**
Well-structured feature: the new login hooks (`pmpro_login_form_before_submit_button`, `pmpro_authenticate_login_checks`) are correctly gated to PMPro/wp-login.php forms only, the legacy-option back-compat shim in `pmpro_captcha()` is sound, and the Turnstile "validated on failure" fix is a real, correctly-applied bug fix. Two issues worth fixing before merge: the failed-login trigger for the captcha challenge is fed by a shared spam-activity signal that isn't login-specific, and the reCAPTCHA verification path can hard-fail on a transient Google API error, newly reachable from the login/password-reset critical path.

---

**Issues**

- **Major** `includes/captcha.php:80-83` (`pmpro_captcha_has_recent_failed_login()`) — This checks `pmpro_get_spam_activity()`, but that transient is shared by three unrelated triggers, confirmed in `includes/spam.php`: failed logins (`pmpro_track_login_spam`, line 233), failed checkout/billing (`pmpro_track_failed_checkouts_for_spam`, line 160, hooked to `pmpro_checkout_processing_failed`/`pmpro_update_billing_failed`), and bad discount codes (`pmpro_check_discount_code_spam_check`, line 208). All three write to the same `pmpro_spam_activity_<ip>` transient via `pmpro_track_spam_activity()`. With Spam Protection and a captcha both enabled, a visitor who fails checkout or enters a bad discount code will be shown (and required to solve) a login captcha on their next login from that IP, even though they never failed a login — contrary to the PR's stated "only after a failed login attempt" behavior. Give login tracking its own bucket, or document that this is intentionally a broader "recent spam signal," not a login-specific one.

- **Major** `includes/recaptcha.php` (`pmpro_recaptcha_verify_token()`, calling `pmpro_ReCaptcha::verifyResponse()`) — `includes/lib/recaptchalib.php:77-80` does `wp_remote_post($path, ...)['body']`. If `wp_remote_post()` returns `WP_Error` (timeout, DNS failure, etc. reaching Google), indexing it with `['body']` is a fatal error in PHP 8 ("Cannot use object of type WP_Error as array"). This bug already exists for checkout (`includes/recaptcha.php:136-137`), but this PR newly puts it on the login and password-reset critical path — a transient network blip to Google now 500s a login attempt instead of just failing a checkout. Worth a guard (`is_wp_error()` check in `_submitHTTPPost`) given the new blast radius, even though it's not a regression this PR introduced.

- **Minor** `includes/cloudflare-turnstile.php` (`pmpro_cloudflare_turnstile_login_forms_html()`) — no `static $already_shown` guard, unlike `pmpro_recaptcha_get_login_html()` (`includes/recaptcha.php`). If both `login_form` and `lostpassword_form` fire in the same request the Turnstile script tag could render twice. Low impact, but worth matching for consistency.

- **Minor** `includes/captcha.php:104` (`pmpro_is_captcha_failed_request()`) — checks `$_REQUEST['action']` alongside `errors`/`error`, but `action` is also the generic login-error param (`failed`, `incorrect_password`, etc., see `includes/login.php:320`). I confirmed the flow is actually correct today — `pmpro_login_failed()` (`includes/login.php:1068`) puts the WP_Error code straight into `action`, so `action=pmpro_captcha_failed` is exactly how the captcha-failed case reaches this function — but overloading one param for three different semantics (single code vs. comma lists) is fragile if a future error code ever contains a comma or collides.

---

**Looks Good**
- `pmpro_authenticate_login_checks` at priority 40 and the narrowed `pmpro_authenticate_username_password()` (redirect+exit only for `empty_username`/`empty_password`) correctly resolve the prior double-`wp_redirect()` race.
- `pmpro_captcha()` back-compat: `false === $captcha` (never saved) falls back to legacy options; a saved `''` (explicit "No") does not — correctly distinguishes upgrade path from user choice.
- Turnstile session-var fix (`cloudflare-turnstile.php`) — `pmpro_set_session_var(..., true)` now only runs on success, closing the retry bypass.
- `pmpro_reset_password_form()` dropping `wp_kses_post()` around `pmpro_lost_password_form()` is correct — that function echoes directly and returns void.

---

**Questions**
1. `pmpro_clear_spam_activity()` is defined (`includes/spam.php:137`) but I found no call site anywhere in the repo — the login-failure signal that gates the captcha never gets cleared on a successful login, so the challenge stays live for the full 15-minute window regardless. Intended?
2. `pmpro_cloudflare_turnstile_lostpassword_check()`/`pmpro_recaptcha_lostpassword_check()` gate on `did_action('login_form_lostpassword')` or `did_action('login_form_retrievepassword')` — does WP core reliably fire one of those two for every lost-password submission path, or is there a third alias that'd skip the captcha check entirely? I couldn't confirm against WP core in this session.

dparker1005:
Thanks for the review! All addressed in 6493be66b:

- **reCAPTCHA lib `WP_Error` guard is in** — `_submitHTTPPost()` now returns an empty body on a failed request, so verification fails safely (the user sees a retryable captcha error) instead of a PHP 8 fatal when Google is unreachable. Exercised the path over real HTTP with an invalid token to confirm.
- **Shared spam signal: intentional, now documented.** Renamed the helper to `pmpro_captcha_has_recent_spam_activity()` and documented the semantics. Challenging an IP at login after it abused checkout is defense-in-depth we want, and a login-only bucket would add another per-IP transient system right before #3640 replaces that storage entirely. The setting description now says "a failed login attempt or other suspicious activity."
- **Turnstile login output** now has the same show-once static guard as the reCAPTCHA output.
- **Your Q1**: yes, intended — clearing activity on a successful login would let an attacker on a shared IP reset the signal, so it just expires with the tracking window. This is now documented in the helper's docblock.
- **Your Q2**: verified against wp-login.php — `lostpassword` and `retrievepassword` are the only two action aliases for that flow, and both are gated. A hypothetical third path would fail open (captcha skipped), never a lockout.
- **No change** to the `action`/`errors`/`error` handling in `pmpro_is_captcha_failed_request()` — agreed it's overloaded, but we only ever emit the fixed `pmpro_captcha_failed` code and you confirmed the flow is correct today. Happy to revisit if a real collision shows up.

PR #3767, opened 2026-08-24
Verify Stripe payment intent amount matches order total after 3DS

dparker1005:
## Problem

In the on-site Stripe flow, when a checkout requires 3D Secure, the browser completes authentication and then re-submits the checkout form with a `payment_intent_id`. On that resubmission, `process_payment_intent()` retrieved and confirmed the browser-supplied PaymentIntent **without verifying that its amount and currency matched the order being finalized**.

Because the order total is rebuilt server-side from the submitted membership level, a user could authenticate a low-priced PaymentIntent, then change the membership level on the resubmission so PMPro activated a more expensive level while Stripe only charged the low amount.

This affects the on-site (Payment Element) SCA/3DS flow. The Stripe Checkout (redirect) flow is not affected, as it reads the amount back from the Stripe session.

## Fix

`process_payment_intent()` now takes the order and rejects the PaymentIntent unless its `amount` and `currency` match the amount due for that order, computed the same way `create_payment_intent()` builds the charge. A legitimate resubmission (same level, same billing) always matches; a tampered level does not.

## Testing

On a Stripe sandbox with a 3DS test card:
- Reproduced the original issue, then confirmed the patched code rejects the mismatched PaymentIntent and does not activate the premium level.
- Confirmed a normal 3DS checkout still completes successfully and records a paid order.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -2231,7 +2231,7 @@ class PMProGateway_stripe extends PMProGateway {
 		if ( ! empty( $order->payment_intent_id ) ) {
 			// User has just tried to confirm their payment intent. We need to make sure that it was
 			// confirmed successfully, and then try to create their subscription if needed.
-			$payment_intent = $this->process_payment_intent( $order->payment_intent_id );
+			$payment_intent = $this->process_payment_intent( $order->payment_intent_id, $order );
 			if ( is_string( $payment_intent ) ) {
 				$order->error      = __( 'Error processing payment intent.', 'paid-memberships-pro' ) . ' ' . $payment_intent;
 				$order->shorterror = $order->error;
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3767 — Verify Stripe payment intent amount matches order total after 3DS**
*dparker1005 → dev | 1 file, +16 -3 lines*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3767>

---

**Summary**

Solid, well-targeted fix for a real price-tampering vulnerability, ready to merge. The exploit is real: `payment_intent_id` is submitted by the browser on 3DS resubmission while the order (and its total) is rebuilt server-side from whatever level is submitted — so a user could authenticate a cheap PaymentIntent, then swap to a pricier level and get it activated without paying the difference. The new check in `process_payment_intent()` (`classes/gateways/class.pmprogateway_stripe.php:3699-3707`) closes that gap by requiring the intent's `amount`/`currency` to match the order total before confirming.

---

**Issues**

- **Minor** `classes/gateways/class.pmprogateway_stripe.php:3704` — The expected amount is computed directly from `$order->subtotal + $order->getTax(true)` via `pmpro_round_price()`/`convert_price_to_unit_amount()`, but `create_payment_intent()` runs those same base params through the public filter `pmpro_stripe_payment_intent_params` at `classes/gateways/class.pmprogateway_stripe.php:4470` — after `amount`/`currency` are set (4454-4455) but before `Stripe_PaymentIntent::create()` is called (4473). I checked all repos under `repos/strangerstudios/` for consumers of this filter; the only shipped one is `pmpro-snippets-library/payment-gateways/stripe/change-statement-descriptor.php`, which only touches `statement_descriptor` and doesn't modify `amount`. So today, in the current codebase, this can't actually misfire. But the filter is documented as general-purpose ("Array of params sent to Stripe"), nothing scopes it away from `amount`, and if any site or future add-on used it to add a surcharge or adjust pricing dynamically, every legitimate 3DS resubmission on that site would be permanently rejected with "This payment does not match the amount due for this checkout." Worth either documenting on the filter that `amount`/`currency` must not be changed, or deriving the expected value by running the same filter (with a throwaway params array) rather than reimplementing the base calculation.

---

**Looks Good**

- The check is derived from the exact same source calculation `create_payment_intent()` uses, not a parallel reimplementation — avoids the classic "two computations drift apart" bug.
- `intval()` on both sides plus `strtolower()` on currency avoids type-juggling and case-sensitivity false negatives (Stripe returns lowercase currency codes; PMPro stores uppercase).
- Fails closed: a mismatch returns a string, which `process()` treats as a hard error and aborts (`classes/gateways/class.pmprogateway_stripe.php:2234-2239`) — no partial-activation path.
- Signature change to `process_payment_intent()` is safe — it has exactly one call site, and the docblock update matches the new parameter accurately.

PR #3768, opened 2026-08-24
Show active members and memberships in activity email

flintfromthebasement:
## Summary

- Report **Active Members** (distinct users) and **Active Memberships** (distinct user/level pairs) separately in the Admin Activity Email, showing the second only when it differs from the first.
- Pull the per-level breakdown from `pmpro_report_get_active_members_per_level()` instead of a second copy of the query, so the email and the Members per Level report can't disagree.
- Count distinct users per level in that report helper. Its column header reads "Number of Active Members" but the query counted rows.
- Clear the report's cached data when a user is deleted.

## Why

Sites that allow multiple memberships per user can have more active memberships than active members, but the email only showed a single "Total Members" figure while its per-level breakdown counted membership rows — so the two could disagree. Two further problems fed into that:

- **Deleted users.** `pmpro_delete_user()` only removes membership history when the admin opts in, so `active` rows for deleted users are the normal state and were inflating the totals.
- **Duplicate level names.** The breakdown grouped by `ml.name`, so two levels sharing a name were silently merged into one line.

Rather than adding queries to fix these, this reuses the report helper that already solved both. That leaves one small `COUNT(DISTINCT user_id)` for the member count, guarantees the email matches the Reports page, and picks up the helper's existing cache.

Counting distinct users per level also makes the summed per-level totals *exactly* equal to the number of distinct user/level pairs, rather than equal only on sites that haven't filtered `pmpro_remove_duplicate_membership_entries` off.

The cache is cleared on `pmpro_after_change_membership_level`, which covers signups, cancellations and expirations. Deleting a user doesn't fire that hook unless the admin also cancels their subscriptions, so `deleted_user` now clears it too — otherwise the cached membership count could keep counting a user who no longer exists.

## Testing

Against a dev database with 440 active members, 454 active memberships, 26 levels (two sharing the name "Free trial"), and one pre-existing orphaned membership row:

- Rendered the email section through `sendAdminActivity()` with `pre_wp_mail` short-circuited so nothing was sent. Confirmed `440 Active Members` / `454 Active Memberships` and the correct top-5 level list.
- Confirmed the two levels named "Free trial" now list separately (7 and 2) instead of merging into one line of 9.
- Confirmed the second heading is suppressed when the counts are equal, and that an empty result set renders without warnings.
- Verified the deleted-user case end to end: created a user with a level, primed the cache, deleted the user without cancelling subscriptions, and confirmed the cache is invalidated and both numbers drop. Before the `deleted_user` hook this rendered `440 / 455`; after, `440 / 454`.
- Confirmed `sum(per-level distinct users) == COUNT(DISTINCT user_id, membership_id)` on a dataset with synthetic duplicate rows injected — both return 454.
- `php -l` on both changed files.

## Proposed changelog entries

```
* ENHANCEMENT: The Admin Activity Email now reports active members and active memberships as separate numbers. #3768 (@flintfromthebasement, @dparker1005)
* BUG FIX: Fixed an issue where the Admin Activity Email and the Members per Level report counted memberships belonging to deleted users and merged membership levels that shared the same name. #3768 (@flintfromthebasement, @dparker1005)
```


dparker1005:
Thanks for this — both bugs you identified are real, and I reproduced both locally before touching anything. The orphaned-row inflation shows up on my dev DB (441 → 440), and I have two levels both named "Free trial" that the old `GROUP BY ml.name` was collapsing into a single line. Good catches, and the `pmpro_delete_user()` reasoning is right: we deliberately keep membership history unless the admin opts in, so leftover `active` rows for deleted users are the normal state, not corruption.

I pushed a commit that keeps the intent but changes the implementation. Three things:

**1. We already have this query.** `pmpro_report_get_active_members_per_level()` in `adminpages/reports/members-per-level.php` already does the `INNER JOIN $wpdb->users` hardening, already groups by `membership_id` (so the duplicate-name bug was already solved there), and is cached in a 24-hour transient invalidated on `pmpro_after_change_membership_level`. Reusing it means we delete a query instead of adding two — no new uncached full-table scans in cron — and more importantly the email and the Members per Level report are now literally the same function, so they can't drift apart. That was the underlying complaint in your description: two places counting the same thing differently.

**2. There was a correctness issue hiding under both our versions.** Summing per-level counts is only equal to `COUNT(DISTINCT user_id, membership_id)` if no user has two active rows for the same level. That's not guaranteed — `pmpro_remove_duplicate_membership_entries` (`includes/functions.php:1473`) defaults to `true` but is explicitly filterable off, and the early-return guard at `functions.php:1326` is skipped on the checkout path, so the dedupe `UPDATE` is the only thing preventing duplicates.

Chasing that down surfaced the actual root bug: that report's column header says **"Number of Active Members"** but the query was `count(*)` — counting rows, not members. It's been mislabeling its own numbers. I changed it to `COUNT(DISTINCT mu.user_id)`, which is a no-op on any site without duplicates, and which makes the summed per-level totals *exactly* equal to the distinct user/level pair count as an identity rather than a coincidence. Verified: with 3 synthetic duplicate rows injected, `COUNT(DISTINCT user_id, membership_id)` and the summed helper both return 454.

That one line is the only change outside the email, and it's the piece most worth a second opinion — it does change Reports output on sites that have opted out of dedupe.

**3. I dropped the tile layout.** The `<h3>` heading was being deleted outright, which left the section with no heading at all when a site has ≤5 levels, and the new tiles sat directly under the Joined/Expired/Cancelled tiles with near-identical styling. Keeping the existing heading style means this doesn't need a design review. I also made the "Active Memberships" line render only when it differs from "Active Members" — on a single-membership site those two numbers are always identical, and showing the same number twice reads as a bug.

**Verification:** rendered the actual email through `sendAdminActivity()` with `pre_wp_mail` short-circuited so nothing sent — 440 Active Members / 454 Active Memberships, correct top-5 list, no warnings. Also exercised the equal-counts branch (second line correctly suppressed) and the empty case. `php -l` clean on both files.

One note on the description: "ran the updated SQL against an affected staging dataset" — I'd rather these say exactly what was run and against what, or leave it out. I re-ran everything myself here, and your numbers held up, but that line isn't something a reviewer can check.

Worth flagging for the changelog: `Total Members` → `Active Members` is a user-visible label and translation string change.

PR #3771, opened 2026-08-26
fix: end membership when gateway-cancelled subscription never ended it

### LINKED ISSUE #3743: Gateway cancellation can permanently orphan a membership when the subscription is already marked cancelled
### Summary

`pmpro_handle_subscription_cancellation_at_gateway()` returns early when the `PMPro_Subscription` is already marked `cancelled`. When the subscription status was set by any path **other** than an in-WordPress cancellation, the membership is never ended, and because `customer.subscription.deleted` only fires once there is no second chance to recover.

The result is a membership left `status = 'active'` with `enddate = '0000-00-00 00:00:00'` forever, against a subscription the gateway has already cancelled. The member keeps full paid access indefinitely.

### Affected code

`includes/gateway-request-handlers.php`, in `pmpro_handle_subscription_cancellation_at_gateway()` (PMPro 3.8.2):

```php
// Check if we have already cancelled the subscription in PMPro.
if ( 'cancelled' === $subscription->get_status() ) {
    return 'We have already processed this cancellation. Probably originated from WP/PMPro. ( Subscription Transaction ID #' . $subscription_transaction_id . ')';
}
```

The comment is the crux. "Probably originated from WP/PMPro" holds for a cancellation initiated inside WordPress, where `pmpro_cancelMembershipLevel()` has already run. It does not hold when the subscription status arrives from a gateway-side sync, which updates `pmpro_subscriptions` without touching `pmpro_memberships_users`.

Both of the function's normal outcomes are correct. Either it calls `pmpro_set_expiration_date()` to run the member to their paid-through date, or it calls `pmpro_cancelMembershipLevel()`. Neither is reached once this guard fires.

### Why it does not self-heal

`pmpro_cron_expire_memberships` expires memberships by `pmpro_memberships_users.enddate`. An orphaned row has `0000-00-00 00:00:00`, which means "never expires", so the cron skips it permanently. Restoring or running the cron does not remediate affected members, which makes the problem easy to misdiagnose as a cron issue.

### Impact

On one production site running Stripe we found 44 orphaned subscription rows across 38 members. The oldest had been orphaned for 745 days. Every one had:

```
pmpro_subscriptions.status       = 'cancelled'
pmpro_subscriptions.enddate      = <a real past date>
pmpro_memberships_users.status   = 'active'
pmpro_memberships_users.enddate  = '0000-00-00 00:00:00'
```

Nothing surfaces this in the admin UI. The member looks like an ordinary active member.

### Detection query

```sql
SELECT s.user_id, s.membership_level_id, s.enddate
FROM wp_pmpro_subscriptions s
JOIN wp_pmpro_memberships_users mu
  ON mu.user_id = s.user_id
 AND mu.membership_id = s.membership_level_id
 AND mu.status = 'active'
WHERE s.status = 'cancelled'
  AND (mu.enddate IS NULL OR mu.enddate = '0000-00-00 00:00:00');
```

### Suggested fix

Make the guard verify the membership was actually ended, rather than inferring it from the subscription status. Roughly:

```php
if ( 'cancelled' === $subscription->get_status() ) {
    $level = pmpro_getSpecificMembershipLevelForUser( $subscription->get_user_id(), $subscription->get_membership_level_id() );
    $membership_still_open = ! empty( $level ) && empty( $level->enddate );
    if ( ! $membership_still_open ) {
        return 'We have already processed this cancellation. ...';
    }
    // Otherwise fall through: the subscription is cancelled but the membership
    // was never ended, so continue to the cancellation logic below.
}
```

That keeps the intended protection against double-processing while closing the case where the subscription status and the membership have diverged.

### Environment

- PMPro 3.8.2
- Gateway: Stripe, live mode
- Observed over roughly two years of data, so not specific to a recent release

faisalahammad:
### All Submissions:

When a subscription is cancelled at the gateway, `pmpro_handle_subscription_cancellation_at_gateway()` returned early if `pmpro_subscriptions.status` was already `cancelled`. That status is not only set by a cancel that originated in WordPress. It is also set by gateway sync, for example when syncing a subscription from the admin Subscriptions page or during `PMPro_Subscription::update()`. Sync marks the subscription cancelled without ending the membership.

In that state, `pmpro_memberships_users` kept `status = 'active'` with an empty end date. The expiry cron skips memberships with no end date, so the member kept paid access forever. This was reported with 44 affected members on one site (issue #3743).

The fix checks whether the user still has an active membership for the subscription's level with no end date before returning early. If yes, the handler continues and ends the membership as it normally would. If the membership was really ended by a WordPress-side cancel, the early return works like before.

Resolves #3743.

1. Create a member with a recurring Stripe subscription.
2. In admin, go to Memberships > Subscriptions and use the sync action so `pmpro_subscriptions.status` becomes `cancelled` while `pmpro_memberships_users` stays active with no end date.
3. Send the `customer.subscription.deleted` webhook (Stripe CLI: `stripe trigger customer.subscription.deleted`).
4. Before this patch the membership stays active. With this patch the membership row gets `status = 'cancelled'` and an end date, and the cancel emails are sent.

Regression checks:

5. Cancel a membership in WordPress, then let the returning webhook arrive. It should still be ignored with "We have already processed this cancellation" and no double process.
6. A gateway cancel where the next payment date is still in the future should still extend the membership to that date (the `pmpro_cancel_on_next_payment_date` path).

Result: all tested manually and working on WordPress 6.x with PMPro 3.8.5 and Stripe test mode.

### Other information:

There is no automated test suite in this repo at the moment, so I verified with `php -l`, PHPCS (`WordPress-Core`) against the changed block, and the manual steps above. Two extra notes on intended behavior:

- An admin who only cancels the subscription (not the membership) will now have the later webhook end the membership too. Previously it stayed open forever, which was the bug.
- A user with two active subscriptions on the same level: one gateway cancel now ends the shared level membership, same as what one webhook always did for a single subscription.

Backward compatible: no breaking changes to documented behavior; only the previously broken case behaves differently.

### Changelog entry

BUG FIX/ENHANCEMENT: Ended memberships are now correctly closed when a gateway cancellation webhook arrives after a subscription was already marked cancelled by a gateway sync.

### DIFF
--- a/includes/gateway-request-handlers.php
+++ b/includes/gateway-request-handlers.php
@@ -46,7 +46,14 @@ function pmpro_handle_subscription_cancellation_at_gateway( $subscription_transa
 
 	// Check if we have already cancelled the subscription in PMPro.
 	if ( 'cancelled' === $subscription->get_status() ) {
-		return 'We have already processed this cancellation. Probably originated from WP/PMPro. ( Subscription Transaction ID #' . $subscription_transaction_id . ')';
+		// Check if the user still has an active membership for this level with no end date.
+		$user_level = pmpro_getSpecificMembershipLevelForUser( $subscription->get_user_id(), $subscription->get_membership_level_id() );
+		$membership_still_open = ! empty( $user_level ) && empty( $user_level->enddate );
+		if ( ! $membership_still_open ) {
+			return 'We have already processed this cancellation. Probably originated from WP/PMPro. ( Subscription Transaction ID #' . $subscription_transaction_id . ')';
+		}
+		// The subscription is marked cancelled at the gateway but the membership
+		// was never ended. Continue to end it below.
 	}
 
 	// Store the old next payment date for the subscription for later.


dparker1005:
Hi @faisalahammad, thank you for tracking down the orphaned-membership case in #3743. The diagnosis is right: gateway sync can mark a subscription cancelled without ever ending the membership, and that leaves paid access open forever.

I don't think the webhook handler can safely make this call, though. That early return exists because "subscription cancelled, membership still active with no end date" can also be a deliberate admin state. On the Edit Member page an admin can check "cancel subscription" while leaving the expiration empty, and the Subscriptions admin page cancel does the same. Both mark the subscription cancelled in PMPro before the gateway sends its cancellation event. With this change, when that event arrives the handler would see an active level with no end date and end the membership, undoing what the admin just did.

The hard part is that by the time the webhook fires, the data looks identical in both cases. Nothing records whether the cancelled status came from an admin action or from sync, so there isn't a clean signal to key off. I'm going to close this so it doesn't sit in the queue, and keep the discussion on #3743 until we've settled on how to disambiguate those. Happy to look at a new PR once that's worked out.

PR #3772, opened 2026-08-26
Prevent fatal at checkout when an order filter callback drops the order object

### LINKED ISSUE #3720: Fatal: getRandomCode() on null in pmpro_checkout_order() at class.pmprogateway_stripe.php:1273 — order object null across all gateways (v3.8, PHP 8.3)
Uncaught Error: Call to a member function getRandomCode() on null
in PMProGateway_stripe::pmpro_checkout_order()
class.pmprogateway_stripe.php:1273
Called from: preheaders/checkout.php:607 → apply_filters()

faisalahammad:
### All Submissions:

During checkout, PMPro applies the `pmpro_checkout_order` filter (or `pmpro_checkout_order_free` for free checkout) to let other plugins adjust the order object. WordPress passes each callback's return value on to the next callback. When another plugin hooks one of these filters and does not return the order object it received, every later callback receives `null`. The Stripe callback then calls `$morder->getRandomCode()` on null and checkout dies with a fatal error on PHP 8, which was reported in issue 3720.

Core always builds a valid order object right before applying these filters, so the null always comes from a callback in the chain. Core cannot stop third-party code from breaking that contract, but core can keep checkout alive when it happens.

This PR:

1. Validates the value coming out of the filter at each apply site (`preheaders/checkout.php` for `pmpro_checkout_order` and `pmpro_checkout_order_free`, `preheaders/billing.php` for `pmpro_billing_order`). If a callback dropped or replaced the order object, we restore the object built before the filter ran and log a notice naming the filter under WP_DEBUG so site owners can find the misbehaving plugin.
2. Adds an early return guard at the top of the Stripe and Braintree `pmpro_checkout_order()` filter callbacks. If something invalid arrives, they pass the value through instead of dereferencing it.

Checkout continues with a known-good order object, gateway methods run as intended, and no 500 is possible from this failure mode again.

Repro snippet that triggered the reported fatal before this PR (add as a mu-plugin):

```php
add_filter( 'pmpro_checkout_order', function ( $order ) { return null; } );
```

Resolves #3720.

1. Install a stock copy of PMPro 3.8.5 without this patch, enable `WP_DEBUG`, add the mu-plugin snippet above, and submit checkout for a paid level with Stripe. You should see the fatal: `Call to a member function getRandomCode() on null` in `class.pmprogateway_stripe.php`.
2. Apply this branch, clear nothing, and submit the same checkout again. Expected result: no fatal. Checkout proceeds normally past the filter; the order processes or fails with a normal processing message instead of a white screen.
3. Check `debug.log`. Expected result: a notice `[PMPro] A callback on the pmpro_checkout_order filter did not return a valid order object.` appears once per broken submission.
4. Remove the mu-plugin and run a normal paid Stripe checkout, a free-level checkout, and a billing update page load/submission. Expected result: everything behaves exactly as before, since valid order objects pass the guards untouched.

Tested manually: confirmed the steps above against a staging copy running PHP 8.x with PMPro 3.8.5 plus Stripe enabled.

Backward compatible: yes, no breaking changes. Filters are unchanged; well-behaved callbacks see identical input and output.

Tests: no new test added. This repo has no PHPUnit test suite (`tests/` directory absent), so the verification path is the manual repro above plus coding standards checks.

Other information:

### Changelog entry

Prevent a fatal during checkout when another plugin hooks the pmpro_checkout_order, pmpro_checkout_order_free, or pmpro_billing_order filters but forgets to return the order object. Checkout now restores the order and logs which filter dropped it instead of fataling in gateway code.

### DIFF
--- a/classes/gateways/class.pmprogateway_braintree.php
+++ b/classes/gateways/class.pmprogateway_braintree.php
@@ -388,6 +388,11 @@ use Braintree\WebhookNotification as Braintree_WebhookNotification;
 		 */
 		static function pmpro_checkout_order($morder)
 		{
+			// Bail if a callback earlier on the filter did not return an order object.
+			if ( ! is_a( $morder, 'MemberOrder' ) ) {
+				return $morder;
+			}
+
 			//load up values
 			if(isset($_REQUEST['number']))
 				$braintree_number = sanitize_text_field($_REQUEST['number']);
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1295,6 +1295,11 @@ class PMProGateway_stripe extends PMProGateway {
 	 */
 	public static function pmpro_checkout_order( $morder ) {
 
+		// Bail if a callback earlier on the filter did not return an order object.
+		if ( ! is_a( $morder, 'MemberOrder' ) ) {
+			return $morder;
+		}
+
 		// Create a code for the order.
 		if ( empty( $morder->code ) ) {
 			$morder->code = $morder->getRandomCode();
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -3129,6 +3129,22 @@ function pmpro_setMessage( $message, $type, $force = false ) {
 }
 
 /**
+ * Log a notice when a callback on an order filter failed to return the order object.
+ *
+ * Some plugins hook order filters like pmpro_checkout_order but forget to return
+ * the order they received. That leaves null for later callbacks and used to cause
+ * a fatal in gateway callbacks. Checkout now recovers, so this just helps find the culprit.
+ *
+ * @since TBD
+ * @param string $filter_name Name of the filter whose callback did not return an order.
+ */
+function pmpro_log_filter_missing_order( $filter_name ) {
+	if ( WP_DEBUG ) {
+		error_log( '[PMPro] A callback on the ' . $filter_name . ' filter did not return a valid order object at checkout.' ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+	}
+}
+
+/**
  * Show a PMPro message set via pmpro_setMessage
  *
  * @since 1.8.5
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -3140,7 +3140,7 @@ function pmpro_setMessage( $message, $type, $force = false ) {
  */
 function pmpro_log_filter_missing_order( $filter_name ) {
 	if ( WP_DEBUG ) {
-		error_log( '[PMPro] A callback on the ' . $filter_name . ' filter did not return a valid order object at checkout.' ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+		error_log( '[PMPro] A callback on the ' . $filter_name . ' filter did not return a valid order object.' ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
 	}
 }
 
--- a/preheaders/billing.php
+++ b/preheaders/billing.php
@@ -251,9 +251,16 @@ if ( is_a( $pmpro_billing_subscription, 'PMPro_Subscription' ) ) {
 			 *
 			 * @since 1.8.13.2
 			 *
-			 * @param object $order the order object used to update billing			 
+			 * @param object $order the order object used to update billing
 			 */
-			$pmpro_billing_order = apply_filters( "pmpro_billing_order", $pmpro_billing_order );
+			$order_before_filters = $pmpro_billing_order;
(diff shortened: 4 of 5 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3772 — Prevent fatal at checkout when an order filter callback drops the order object**
*faisalahammad → dev | 5 files, +43 -2*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3772>

---

**Summary**
Ready to merge after a one-word fix to the log message. The two-layer design is right: type guards at the top of the Stripe and Braintree `pmpro_checkout_order()` callbacks protect sibling callbacks still queued in the same `apply_filters()` chain, and the restore-and-log after `apply_filters()` in `preheaders/checkout.php` and `preheaders/billing.php` protects everything downstream. I verified on `upstream/dev` that Stripe (`class.pmprogateway_stripe.php:156-157`) and Braintree (`class.pmprogateway_braintree.php:243-244`) are the only gateways hooking `pmpro_checkout_order` / `pmpro_billing_order`, so in-core coverage is complete. Both restore sites fall back to a value that is guaranteed to be a `MemberOrder`: `preheaders/checkout.php:589` constructs `$pmpro_review` immediately before the filter, and `preheaders/billing.php:47-53` builds `$pmpro_billing_order` before the filter block.

---

**Issues**
- **[Minor]** `includes/functions.php:3143` — `pmpro_log_filter_missing_order()` hardcodes "at checkout", but it is also called from `preheaders/billing.php:261` for `pmpro_billing_order`. A broken billing callback logs "…the pmpro_billing_order filter did not return a valid order object at checkout", which points a site owner at the wrong flow. Drop the "at checkout" suffix (the filter name already identifies the flow) or add a `$context` argument.

---

**Looks Good**
- `is_a( $morder, 'MemberOrder' )` is the right check: it rejects null, false, and foreign objects alike, and still passes subclasses.
- Gateway guards return the invalid value unchanged rather than null, so the chain completes and the preheader guard does the single authoritative restore. No silent data loss.
- Restoring the pre-filter object instead of constructing a new one preserves the sanitized billing fields set earlier in the preheader.
- Both gateways reuse `pmpro_checkout_order()` for the `pmpro_billing_order` hook, so one guard per gateway covers both filters.
- Bare `if ( WP_DEBUG )` matches existing usage in `classes/class-pmpro-recurring-actions.php` and `adminpages/wizard/save-steps.php`. `@since TBD` is correct per team convention.

---

**Questions**
1. `pmpro_build_order_for_checkout()` at `includes/functions.php:4274` (upstream/dev) also applies `pmpro_checkout_order` unguarded. It is deprecated since 3.2 (`_deprecated_function` at line 4241, CHANGELOG #3129) with no core callers, so this is low value. Guard it for description accuracy, or leave it? Either is fine.
2. Behavior change worth a conscious nod: a callback that returned `null`/`false` to abort checkout now has that overridden and payment proceeds. That was never a documented contract and already produced a 500 on PHP 8, so overriding is defensible. Any known Add On relying on a falsy return to halt checkout?

faisalahammad (commented):
Thanks for the review. Pushed a fix for the log message and answers to both questions below.

**Log message fix (a6fffe6ef)**: Dropped the "at checkout" suffix from `pmpro_log_filter_missing_order()`. The filter name in the message already tells the site owner which flow broke, so the billing log line now reads "A callback on the pmpro_billing_order filter did not return a valid order object." Also updated step 3 of the test instructions in the PR description to match.

**Question 1, the deprecated builder**: Left `pmpro_build_order_for_checkout()` unguarded. It is deprecated since 3.2, has no core callers, and adding a guard there grows the diff without protecting a real flow. New code should use the checkout preheader path, which is covered.

**Question 2, falsy return to abort checkout**: I searched the official docs, the add-on repos, and public code on GitHub for callbacks on `pmpro_checkout_order`, `pmpro_checkout_order_free`, and `pmpro_billing_order`. No code returns falsy on purpose to abort checkout. The documented abort mechanism is the `pmpro_checkout_*_checks` filters, and those are not touched here. So no known Add On relies on the old behavior, and the override is safe.

dparker1005:
Hi @faisalahammad, thank you for digging into this and for the clear write-up of the root cause.

I'm going to close this one, though. The fatal in #3720 comes from a third-party callback on `pmpro_checkout_order` that doesn't return the order it was given. That's a bug in that callback, and core intentionally trusts filter return values rather than validating and restoring them.

We'd rather have the fatal here. It's loud and it points straight at the broken callback, so the site owner finds and fixes it. If core quietly restored the order, checkout would appear to work while every callback later in the chain (including our own gateway callbacks) had received `null` and skipped its work, and nobody would know anything was wrong. A visible failure is the better outcome than a hidden one.

I'll follow up on #3720 with the diagnosis so the reporter can track down the plugin that's dropping the order.

PR #3773, opened 2026-08-26
Track attachments in email log entries

### LINKED ISSUE #3647: Email Log: Track attachments in logged email entries
## Feature Request

The email log report (`Memberships > Reports > Email Log`) currently captures a solid set of email data (to, from, subject, body, headers, CC, BCC, template, status) but does **not** track attachments.

It would be useful to log attachment information alongside the rest of the email data so admins can see what files were included in a sent email.

## Current Schema

The `wp_pmpro_email_log` table (defined in `includes/upgradecheck.php`) has no `attachments` column. The `pmpro_log_email()` function in `includes/email-logging.php` also does not capture attachment data from the `$mail_data` array.

## Proposed Change

1. **DB:** Add an `attachments` column (e.g. `text`, serialized array of file paths/names) to `wp_pmpro_email_log`.
2. **Logging:** Update `pmpro_log_email()` to capture the `attachments` key from `$mail_data` (WordPress passes this via `wp_mail` args).
3. **Report UI:** Display attachment info in the email log detail view — at minimum the filename(s), ideally with a count shown in the list view.

## Notes

- `wp_mail()` accepts an `$attachments` argument (array of file paths). This is available in the `phpmailer_init` hook or via the `wp_mail` filter, which is already where PMPro hooks in for logging.
- Attachment file paths may be transient (e.g. temp files), so storing the **filename** (not full path) is probably safer for the log record.
- For invoices or other generated PDFs attached at checkout, this would give admins visibility into what was sent.

## Reported By

Andrew Lima

faisalahammad:
### All Submissions:

The email log stored every part of a sent email except its attachments. Attachment paths are usually temp files that do not survive the request, so only filenames are stored.

This PR adds an `attachments` column to `{prefix}pmpro_email_log` (TEXT, `maybe_serialize()`d array of filenames, same convention as the existing `headers` column) via a new 3.9 db version gate in `pmpro_checkForUpgrades()`. Filenames are captured from `$mail_data['attachments']` in `pmpro_log_email()`, which accepts both an array and the newline-separated string form that `wp_mail()` itself accepts. String lists are split on newline, matching how WordPress core normalizes attachments in `wp_mail()`.

They are then shown in two places:

- **Email log details modal**: an "Attachments" row listing the filenames, shown only when the entry has attachments.
- **Member Edit > Email Log panel** and the **personal data exporter**: the exporter includes the same "Attachments" row for GDPR completeness.

The parsing lives in one helper, `pmpro_get_email_log_attachments( $log )` in `includes/email-logging.php`, called by the modal renderer and the privacy exporter.

Per review feedback, the list-table columns that were originally added to **Memberships > Reports > Email Log** and the member edit email log panel have been dropped. The details modal covers the debugging use case, and those columns would show a dash on nearly every row.

All readers use `SELECT *`, so the REST API endpoint picks up the new field automatically. Old rows simply have no attachments data.

Resolves #3647.

1. With email logging enabled, add this snippet and trigger any PMPro email:

```php
add_filter( 'pmpro_email_attachments', function( $attachments, $email ) {
	$file = tempnam( get_temp_dir(), 'pmpro-test-' ) . '.pdf';
	file_put_contents( $file, 'test attachment' );
	return array( $file );
}, 10, 2 );
```

2. Load an admin page once so the upgrade routine runs `dbDelta`. Verify with `wp db query "DESCRIBE wp_pmpro_email_log"` that an `attachments` column exists, and with `wp option get pmpro_db_version` that the db version is now `3.9`.

3. Open Memberships > Reports > Email Log and click "View Content" on the new entry. Expected: an Attachments row listing just the filename, not the full temp path. Also expected: no Attachments column in the list table itself.

4. Remove the snippet, send another email, and open its modal. Expected: no Attachments row at all.

5. Test the string form, which is the case that was logging wrong before. Change the snippet to return two temp files as a single newline-separated string instead of an array:

```php
add_filter( 'pmpro_email_attachments', function( $attachments, $email ) {
	$a = tempnam( get_temp_dir(), 'pmpro-test-a-' ) . '.pdf';
	$b = tempnam( get_temp_dir(), 'pmpro-test-b-' ) . '.pdf';
	file_put_contents( $a, 'a' );
	file_put_contents( $b, 'b' );
	return $a . "\n" . $b;
}, 10, 2 );
```

Expected: the modal shows both filenames. Before this revision only one filename was stored.

6. Edit a member and open the Email Log panel. Expected: no Attachments column in the list, and the same modal behavior as step 3.

7. Tools > Export Personal Data for that user. Expected: each email log group in the HTML report includes an Attachments row when attachments exist.

8. Verify pre-upgrade log rows still display correctly (no PHP notices).

Result: all scenarios pass on my test site (WP 6.x, PHP 8.x).

### Other information:

Note: no automated test suite exists in this repository, so testing was done manually against a live WordPress install (scenarios above plus old-row upgrade simulation and failed-send path where status stays Failed but attachments are still recorded).

Backward compatible: yes, no breaking changes. The new column starts empty, old rows render unchanged, and rollback is inert (column remains but unused).

### Changelog entry

Added: Track email attachments in email log entries. The email log details modal and the personal data exporter now show which files were attached to each logged email.

### DIFF
--- a/adminpages/member-edit/pmpro-class-member-edit-panel-email-log.php
+++ b/adminpages/member-edit/pmpro-class-member-edit-panel-email-log.php
@@ -60,6 +60,7 @@ class PMPro_Member_Edit_Panel_Email_Log extends PMPro_Member_Edit_Panel {
 						<th><?php esc_html_e( 'To', 'paid-memberships-pro' ); ?></th>
 						<th><?php esc_html_e( 'Subject', 'paid-memberships-pro' ); ?></th>
 						<th><?php esc_html_e( 'Template', 'paid-memberships-pro' ); ?></th>
+						<th><?php esc_html_e( 'Attachments', 'paid-memberships-pro' ); ?></th>
 						<th><?php esc_html_e( 'Status', 'paid-memberships-pro' ); ?></th>
 					</tr>
 				</thead>
@@ -95,6 +96,22 @@ class PMPro_Member_Edit_Panel_Email_Log extends PMPro_Member_Edit_Panel {
 									}
 
 									echo $description ? esc_html( $description ) : esc_html__( '&#8212;', 'paid-memberships-pro' );
+								?>
+							</td>
+							<td data-colname="<?php esc_attr_e( 'Attachments', 'paid-memberships-pro' ); ?>">
+								<?php
+									$entry_attachments = ! empty( $entry->attachments ) ? maybe_unserialize( $entry->attachments ) : array();
+									if ( is_string( $entry_attachments ) ) {
+										$entry_attachments = preg_split( '/[,;]/', $entry_attachments );
+									}
+									$entry_attachments = array_filter( array_map( 'trim', (array) $entry_attachments ), 'strlen' );
+
+									if ( ! empty( $entry_attachments ) ) {
+										echo '<span class="dashicons dashicons-paperclip" title="' . esc_attr( implode( ', ', $entry_attachments ) ) . '"></span> ';
+										echo esc_html( number_format_i18n( count( $entry_attachments ) ) );
+									} else {
+										echo '—';
+									}
 								?>
 							</td>
 							<td data-colname="<?php esc_attr_e( 'Status', 'paid-memberships-pro' ); ?>">
--- a/adminpages/member-edit/pmpro-class-member-edit-panel-email-log.php
+++ b/adminpages/member-edit/pmpro-class-member-edit-panel-email-log.php
@@ -60,7 +60,6 @@ class PMPro_Member_Edit_Panel_Email_Log extends PMPro_Member_Edit_Panel {
 						<th><?php esc_html_e( 'To', 'paid-memberships-pro' ); ?></th>
 						<th><?php esc_html_e( 'Subject', 'paid-memberships-pro' ); ?></th>
 						<th><?php esc_html_e( 'Template', 'paid-memberships-pro' ); ?></th>
-						<th><?php esc_html_e( 'Attachments', 'paid-memberships-pro' ); ?></th>
 						<th><?php esc_html_e( 'Status', 'paid-memberships-pro' ); ?></th>
 					</tr>
 				</thead>
@@ -96,22 +95,6 @@ class PMPro_Member_Edit_Panel_Email_Log extends PMPro_Member_Edit_Panel {
 									}
 
 									echo $description ? esc_html( $description ) : esc_html__( '&#8212;', 'paid-memberships-pro' );
-								?>
-							</td>
-							<td data-colname="<?php esc_attr_e( 'Attachments', 'paid-memberships-pro' ); ?>">
-								<?php
-									$entry_attachments = ! empty( $entry->attachments ) ? maybe_unserialize( $entry->attachments ) : array();
-									if ( is_string( $entry_attachments ) ) {
-										$entry_attachments = preg_split( '/[,;]/', $entry_attachments );
-									}
-									$entry_attachments = array_filter( array_map( 'trim', (array) $entry_attachments ), 'strlen' );
-
-									if ( ! empty( $entry_attachments ) ) {
-										echo '<span class="dashicons dashicons-paperclip" title="' . esc_attr( implode( ', ', $entry_attachments ) ) . '"></span> ';
-										echo esc_html( number_format_i18n( count( $entry_attachments ) ) );
-									} else {
-										echo '—';
-									}
 								?>
 							</td>
 							<td data-colname="<?php esc_attr_e( 'Status', 'paid-memberships-pro' ); ?>">
--- a/adminpages/reports/email_log.php
+++ b/adminpages/reports/email_log.php
@@ -383,6 +383,7 @@ function pmpro_report_email_log_page() {
 						<th><?php esc_html_e( 'Subject', 'paid-memberships-pro' ); ?></th>
 						<th><?php esc_html_e( 'Template', 'paid-memberships-pro' ); ?></th>
 						<th><?php esc_html_e( 'User', 'paid-memberships-pro' ); ?></th>
+						<th><?php esc_html_e( 'Attachments', 'paid-memberships-pro' ); ?></th>
 						<th><?php esc_html_e( 'Status', 'paid-memberships-pro' ); ?></th>
 					</tr>
 				</thead>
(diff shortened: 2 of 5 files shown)


dparker1005 (changes requested):
Hi @faisalahammad, thank you for picking up #3647. The shape matches the issue well: filenames only, serialized like the existing headers column, and the privacy exporter is updated, which is easy to miss.

Before we go further I've asked @andrewlimaza to weigh in, since the request came from support. Nothing in core or our Add Ons attaches files to PMPro emails, so we want to confirm this is answering a question support actually gets before adding a permanent column to the log table. If it is, here's what would need to change:

1. **Upgrade version.** The database version gate is 3.86, but 3.8.6 has already shipped. Since this adds a column it's a 3.9 change (I've retargeted the PR to `v3.9`), so the gate and comment should match whatever release it ships in.
2. **Drop the list-table columns.** Both the Email Log report and the Member Edit email panel gain an Attachments column that will show a dash on almost every row. The row in the details modal covers the debugging use case. We'll leave the final UI call to Kim, but my recommendation is modal only.

Thanks again.

flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3773 — Track attachments in email log entries**
*faisalahammad → v3.9 | 5 files, +89 -2*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3773>

---

**Summary**
Not ready to merge. The shape is right (filenames only, serialized like `headers`, exporter covered), but string-form attachments are split on the wrong delimiter, the same parsing block is pasted into four files, and dparker1005's version-gate and column-placement points are still open.

---

**Issues**

- **[Major]** `includes/email-logging.php:259` — `preg_split( '/[,;]/', (string) $mail_data['attachments'] )` assumes comma/semicolon separation. WordPress does not use that form. `wp_mail()` normalizes a string attachment list with `explode( "\n", str_replace( "\r\n", "\n", $attachments ) )` (`wp-includes/pluggable.php:263-264`), and that happens *after* the `wp_mail` filter fires at `pluggable.php:209`. PMPro's `pmpro_stash_outgoing_email()` hooks that filter (`includes/email-logging.php:304` on `v3.9`) and the stashed array is what `pmpro_log_email()` receives from `wp_mail_succeeded`/`wp_mail_failed`. So a valid `"/tmp/a.pdf\n/tmp/b.pdf"` is treated as one element and `wp_basename()` logs a single mangled name. Since logging covers all site `wp_mail()` traffic, third-party callers will hit this. Split on newline to match core; drop the `[,;]` split in the readers too, since nothing writes that form.

- **[Major]** `includes/upgradecheck.php:466` — Gate is `< 3.86` with a "Version 3.8.6" docblock, but 3.8.6 has shipped and this lands in v3.9 (dparker1005 already flagged this). The last gate on `v3.9` is `3.84` (`upgradecheck.php:458`), so it won't double-fire, but the number and comment should reflect the release it ships in.

- **[Minor]** `includes/email-logging.php:569`, `includes/privacy.php:404`, `adminpages/reports/email_log.php:444`, `adminpages/member-edit/pmpro-class-member-edit-panel-email-log.php:103` — Identical unserialize/split/filter block in four places. Extract `pmpro_get_email_log_attachments( $log )` in `includes/email-logging.php` and call it from the others. This also means the delimiter fix above happens once, and the `is_string()` fallback (dead today, since storage always serializes an array) is justified in one spot.

- **[Minor]** `adminpages/reports/email_log.php:451`, `adminpages/member-edit/pmpro-class-member-edit-panel-email-log.php:113` — `echo '—';` while the adjacent Template column uses `esc_html__( '&#8212;', ... )`. Match the neighbor. Moot if the list-table columns are dropped per dparker1005's recommendation.

---

**Looks Good**
- `wp_basename()` at write time (`includes/email-logging.php:263`) keeps temp paths out of the DB, modal, REST response, and GDPR export in one stroke.
- `$wpdb->insert()` format array (`includes/email-logging.php:294`) checked against `$log_data`: 16 keys, 16 specifiers, `%d` on `user_id`.
- Every output path escapes: `esc_attr()` on the title attribute, `esc_html()` in the modal, plain string to the core exporter.
- New column matches the existing `cc`/`bcc` definition (`text NOT NULL`), and the privacy exporter update is the part people usually forget.

---

**Questions**
1. Since the list-table column decision is with Kim, do you want to hold the next push until that lands, or fix the delimiter and version gate now and strip the columns later if asked?

faisalahammad:
Thanks both. Pushed 19911ebf5 with the changes.

@dparker1005
1. Version gate is now 3.9. The docblock comment and the `update_option` value match. The previous gate in the file is 3.84, so it runs once and does not run again after.
2. List table columns are dropped. I removed the Attachments column from both the Email Log report and the member edit email log panel. Both files are back to the same code as the v3.9 base now. Attachments only show in the details modal.

@flintfromthebasement
1. You are right about the delimiter. I checked core again and `wp_mail()` splits a string attachment list with `explode( "\n", str_replace( "\r\n", "\n", $attachments ) )`, and that runs after the `wp_mail` filter we stash from. I now split on newline in the same way. I ran the old code against `"/tmp/a.pdf\n/tmp/b.pdf"` and it stored one mangled name, the new code stores both. The `[,;]` split is gone from the write and read paths.
2. Version gate fixed, as above.
3. Extracted `pmpro_get_email_log_attachments( $log )` in `includes/email-logging.php`. The helper is now the only place that parses the stored value. The modal and the privacy exporter call it. The two list table copies are gone with the columns.
4. The `echo '—';` lines were in the columns, so they are gone with them.

On your question about waiting for Kim: I fixed the delimiter and version gate now and also dropped the columns, since that was the recommendation in the review. If Kim wants a column later it is a small change on top of this.

PR description is updated. I removed the column bullets, changed the comma and semicolon wording to the newline form, and added a test step for the string case. `php -l` passes on all five files and CodeRabbit found no issues.

ideadude:
Nice catch to remove the columns from main table. I thought we could add an icon or indicator there, but couldn't think of the right spot. I don't think "scanning for the email with the attachment" is a thing folks are doing.

However, I do like that if you open the content you can see the attachment names. Storing files would be problematic (sending attachments from WP is difficult and rare in my experience anyway). We can add a gist or something for that if folks request and consider it later if it seems more useful.

This feels good to ship to me now.

A small amount of additional code, but we are showing the full contents of the email when it has an attachment like this.

dparker1005 (approved):
Thanks @faisalahammad, the follow-up covers everything. Merging into v3.9.

PR #3775, opened 2026-08-27
ENHANCEMENT: Stripe Connect disconnect hash + dead-key reconnect

flintfromthebasement:
## Description
Combines hashed Connect disconnect protection with a payment-settings key health check.

When a site Connects, we send `supports_hash=1`. The Connect server stores `sha256(access_token)`. Disconnect sends `key_hash` only if this site is hashed. Mismatch does not deauthorize other sites — reconnect instead.

After Payment settings load, we ping Stripe `GET /v1/customers?limit=1` with the keys actually used (API keys take precedence over Connect). That is a permission PMPro actually needs; Restricted Keys usually cannot `GET /v1/account`. An empty Stripe account still returns a list, so we only care that the key authenticates. Auth failures show an inline notice. Connect: **Reconnect with Stripe** (same authorize URL, including `supports_hash`). Restricted Key: **Check your Restricted Key**. Working keys stay quiet. Secrets never hit the browser.

This supersedes #3776 (merged into this branch).

## Files
- `classes/gateways/class.pmprogateway_stripe.php`
- `css/admin.css`
- `includes/scripts.php`
- `js/pmpro-admin.js`

## To Test
**Hash handshake**
1. Connect on a site running this branch. `pmpro_{live|sandbox}_stripe_connect_hashed` is set. Disconnect URL includes `key_hash`.
2. Second site, same Stripe account, unhashed: Disconnect is refused. Reconnect notice with Reconnect / Learn More.
3. Reconnect on the unhashed site issues new keys without deauthorizing the first site.

**Key health**
1. Valid Connect keys: AJAX returns empty results, no banner.
2. Flip one character of the Connect secret: inline error + Reconnect with Stripe.
3. Restricted Key path: same ping (`customers?limit=1`), copy is “Check your Restricted Key”, no reconnect link. Empty Stripe account (no customers) must stay quiet.

### DIFF
--- a/adminpages/wizard/save-steps.php
+++ b/adminpages/wizard/save-steps.php
@@ -133,6 +133,7 @@ function pmpro_init_save_wizard_data() {
 				array(
 					'action' => 'authorize',
 					'gateway_environment' => $environment,
+					'supports_hash' => '1',
 					'return_url' => rawurlencode( add_query_arg( 'pmpro_stripe_connect_nonce', wp_create_nonce( 'pmpro_stripe_connect_nonce' ), $next_step ) ),
 				),
 				esc_url( $connect_url_base )
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1304,7 +1304,8 @@ class PMProGateway_stripe extends PMProGateway {
 	}
 
 	/**
-	 * Test a Stripe secret (Connect or Restricted Key) with a cheap authenticated GET.
+	 * Test a Stripe secret (Connect or Restricted Key) with a cheap customers list.
+	 * Empty accounts still return a list; we only care that the key authenticates.
 	 *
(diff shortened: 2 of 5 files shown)


flintfromthebasement:
Converting back to draft — this needs a rethink.

The connect server now requires a Stripe sign-in (OAuth) to disconnect, verified server-side against the account being disconnected. That closes the unauthenticated-disconnect vulnerability for every plugin version, including everything already in the wild — so the `supports_hash`/`key_hash` handshake in this PR is no longer solving a security problem. On the server, a matching `key_hash` is now only used to show the list of connected sites on the disconnect interstitial.

Parts of this PR may still be worth keeping in some form — the dead-key detection and reconnect UX in particular — but the hash plumbing shouldn't ship as-is. Needs a fresh look at what's still useful.

ideadude:
Closed in favor of https://github.com/strangerstudios/paid-memberships-pro/pull/3791

PR #3776, opened 2026-08-27
ENHANCEMENT: Detect invalid Stripe Connect keys and offer reconnect

flintfromthebasement:
## Description
When Stripe credentials are saved, the payment settings page now tests those keys in the background. If Stripe rejects them (invalid key, 401/403), we show an inline error. Working keys stay quiet — no extra green banner.

This is separate from hashed Connect options (PR #3775).

## What it does
1. After Memberships → Settings → Payment (Stripe) loads, admin JS fires `pmpro_stripe_check_connect_keys`.
2. The AJAX handler never receives a secret. It reads the keys actually used from options and calls Stripe `GET /v1/account`.
3. If the site is using API keys (`pmpro_stripe_secretkey` + publishable key), only that Restricted Key is tested. Otherwise live and sandbox Connect secrets are tested when credentials exist.
4. Connect auth failures show: “We could not reach Stripe with the saved keys.” plus **Reconnect with Stripe** (same `action=authorize` URL as Connect with Stripe, including `pmpro_stripe_connect_nonce`).
5. API-key auth failures show: “We could not reach Stripe with the saved API keys. Check your Restricted Key.” No reconnect — that is a Connect action.
6. Transient/network errors do not nag. Page load is not blocked.

## Files
- `classes/gateways/class.pmprogateway_stripe.php`
- `includes/scripts.php`
- `js/pmpro-admin.js`

## To Test
**Connect happy path**
1. On flintdev.test, go to Memberships → Settings → Payment with Stripe Connect already connected (live and/or test).
2. Load the page. DevTools → Network: `admin-ajax.php?action=pmpro_stripe_check_connect_keys` returns `{ success: true, results: {} }`.
3. No extra success banner. Secret key is not in the AJAX request payload.

**Connect failure path**
1. Temporarily replace `pmpro_live_stripe_connect_secretkey` or `pmpro_sandbox_stripe_connect_secretkey` with garbage (keep user id + publishable key so Connect still looks connected).
2. Reload Payment settings.
3. Inline error appears under Disconnect From Stripe in that environment’s section (collapsed sections expand).
4. Reconnect with Stripe uses the same authorize URL as Connect with Stripe and completes a normal reconnect.

**API key / Restricted Key path**
1. Site using Restricted Keys (no Connect, or API keys taking precedence).
2. Load Payment settings. Valid keys stay quiet.
3. Replace `pmpro_stripe_secretkey` with garbage and reload.
4. Inline error under Restricted Key: check your Restricted Key. API key table is shown. No Reconnect with Stripe link.

**Also**
- Invalid keys should not break the rest of the settings page.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -131,6 +131,7 @@ class PMProGateway_stripe extends PMProGateway {
 		add_action( 'wp_ajax_pmpro_stripe_create_webhook', array( 'PMProGateway_stripe', 'wp_ajax_pmpro_stripe_create_webhook' ) );
 		add_action( 'wp_ajax_pmpro_stripe_delete_webhook', array( 'PMProGateway_stripe', 'wp_ajax_pmpro_stripe_delete_webhook' ) );
 		add_action( 'wp_ajax_pmpro_stripe_rebuild_webhook', array( 'PMProGateway_stripe', 'wp_ajax_pmpro_stripe_rebuild_webhook' ) );
+		add_action( 'wp_ajax_pmpro_stripe_check_connect_keys', array( 'PMProGateway_stripe', 'wp_ajax_pmpro_stripe_check_connect_keys' ) );
 
 		//code to add at checkout if Stripe is the current gateway
 		$default_gateway = get_option( 'pmpro_gateway' );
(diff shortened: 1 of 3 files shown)


flintfromthebasement:
Merged into #3775 so reconnect always sends supports_hash.

PR #3783, opened 2026-09-04
Detect rejected Stripe credentials and offer a reconnect

dparker1005:
## What this does

When a site has Stripe credentials saved but Stripe no longer accepts them (a revoked or rolled Restricted Key, or a Stripe Connect access token that was revoked, most commonly because a staging site clicked Disconnect on the same Stripe account), nothing in PMPro tells the admin. Worse, the existing webhook notice fires on every PMPro admin page saying the webhook is disabled or missing, because the webhook lookup fails with the dead key and its failure is indistinguishable from "no webhooks found".

This PR adds a cached connection check to the Stripe gateway and surfaces the result where admins look:

- **Payment settings**: the Status tag becomes "Connection Error" and an inline notice shows Stripe's error message with a **Reconnect with Stripe** link (Connect sites) or a prompt to enter a new Restricted Key (API key sites). The webhook section stops probing webhooks and explains that the connection has to be restored first.
- **Admin notice** on PMPro admin pages (including the payment settings gateway list) for users with `manage_options`, with the same reconnect link. When it fires, the misleading webhook notice is suppressed.
- **Dashboard checklist**: "Configure Payment Settings" loses its checkmark while the cached result says the credentials are rejected.
- **Site Health**: the gateway line includes the connection error.

**Reconnect** starts the normal Connect *authorize* flow, not disconnect. Disconnecting would revoke access for every site on that Stripe account. The existing return handler already overwrites the stored credentials, so no server changes are needed.

## How the check works

- `PMProGateway_stripe::get_connection_error()` lists one Customer with the active environment's key passed per request. Only an HTTP 401 counts as rejected credentials. A 403 (restricted key without permission for that endpoint) and network failures do not, so a working narrow Restricted Key never gets flagged.
- Classification is by HTTP status, not exception class, so it still works when another plugin loaded an older stripe-php first.
- Cached in a transient per environment: 1 hour when healthy or on a permanent non-401 failure, 5 minutes on a 401 or a transient network/5xx/429 failure. Cache is cleared on Stripe settings save, Connect return, and disconnect.
- `get_cached_connection_error()` reads the cache only and is the only thing the dashboard checklist touches, so no Stripe call is ever made from a non-admin request.

## Also in this PR

- **Same-account guard**: the Connect return handler now refuses to save credentials for a *different* Stripe account than the one already stored for that environment, with an explanation to disconnect first. Reconnecting to the wrong account would orphan every existing customer and subscription, and this PR actively funnels admins into the authorize flow while already connected, so the guard felt necessary. Easy to drop if you'd rather rely on the copy alone.
- The inline Connect URL building in the settings section moved into a small `get_connect_url()` helper so the notice can reuse it. Output is byte-identical to the old code for all four live/sandbox × authorize/disconnect cases (checked with a stub harness).
- `pmpro_is_ready()` is intentionally **not** changed. Flipping `$pmpro_gateway_ready` there would resurrect the Setup Wizard menu on sites that never ran it and add a duplicate "Next step: Set up your payment gateway" header message.

## Verification

Reviewed by three independent Opus agents (correctness/fatal risks, adversarial break-it, backward compat/UX) and their findings were folded in. Verified by reading code, `php -l`, and the URL equivalence harness. **Not yet exercised in a browser or against a real revoked key**, so a sandbox pass of the following would be good before merging:

1. Connect in test mode, then revoke the token from the Stripe dashboard (Settings → Authorized applications) or roll a Restricted Key. Within 5 minutes (or after saving Stripe settings) the status, notice, and checklist should all flip.
2. Click Reconnect with Stripe with the same account. Everything should clear immediately.
3. Click Reconnect with a *different* account. The connection should stay unchanged and the error message should explain why.
4. Confirm a working Restricted Key that lacks Customers read permission is **not** flagged (should return 403, which is ignored).

## Notes for follow-up (not in this PR)

- A "Check again" link to clear the cache on demand.
- Styling the settings-page reconnect link as a Stripe Connect button (Kim).
- The notice text says members cannot check out, which is technically moot on free-only sites. The webhook notice has the same behavior today.

## Proposed changelog entries

```
* ENHANCEMENT: Detect when Stripe no longer accepts the saved Stripe credentials and show the admin a notice with an option to reconnect. #3783 (@dparker1005)
* BUG FIX: The Stripe webhook notice no longer reports a missing webhook when the real problem is that Stripe is rejecting the saved credentials. #3783 (@dparker1005)
* BUG FIX: Connecting to Stripe now refuses to overwrite an existing connection with a different Stripe account. #3783 (@dparker1005)
```

### DIFF
--- a/adminpages/metaboxes/welcome.php
+++ b/adminpages/metaboxes/welcome.php
@@ -56,9 +56,13 @@ function pmpro_dashboard_welcome_callback() {
 				<?php } ?>
 
 				<?php if ( current_user_can( 'pmpro_paymentsettings' ) ) { ?>
+					<?php
+					// Stripe credentials that Stripe has rejected don't count as configured. This only reads the last cached check.
+					$stripe_rejected = 'stripe' === get_option( 'pmpro_gateway' ) && class_exists( 'PMProGateway_stripe' ) && PMProGateway_stripe::get_cached_connection_error();
+					?>
 					<li>
 						<a href="<?php echo esc_url( admin_url( 'admin.php?page=pmpro-paymentsettings' ) ); ?>">
-							<i class="dashicons <?php echo ( empty( $pmpro_gateway_ready ) || empty( $pmpro_level_ready ) ) ? 'dashicons-marker' : 'dashicons-yes-alt'; ?>"></i>
+							<i class="dashicons <?php echo ( empty( $pmpro_gateway_ready ) || empty( $pmpro_level_ready ) || $stripe_rejected ) ? 'dashicons-marker' : 'dashicons-yes-alt'; ?>"></i>
 							<span><?php esc_html_e( 'Configure Payment Settings', 'paid-memberships-pro' ); ?></span>
 						</a>
 					</li>
--- a/adminpages/metaboxes/welcome.php
+++ b/adminpages/metaboxes/welcome.php
@@ -56,13 +56,9 @@ function pmpro_dashboard_welcome_callback() {
 				<?php } ?>
 
 				<?php if ( current_user_can( 'pmpro_paymentsettings' ) ) { ?>
-					<?php
-					// Stripe credentials that Stripe has rejected don't count as configured. This only reads the last cached check.
-					$stripe_rejected = 'stripe' === get_option( 'pmpro_gateway' ) && class_exists( 'PMProGateway_stripe' ) && PMProGateway_stripe::get_cached_connection_error();
-					?>
 					<li>
 						<a href="<?php echo esc_url( admin_url( 'admin.php?page=pmpro-paymentsettings' ) ); ?>">
-							<i class="dashicons <?php echo ( empty( $pmpro_gateway_ready ) || empty( $pmpro_level_ready ) || $stripe_rejected ) ? 'dashicons-marker' : 'dashicons-yes-alt'; ?>"></i>
+							<i class="dashicons <?php echo ( empty( $pmpro_gateway_ready ) || empty( $pmpro_level_ready ) ) ? 'dashicons-marker' : 'dashicons-yes-alt'; ?>"></i>
 							<span><?php esc_html_e( 'Configure Payment Settings', 'paid-memberships-pro' ); ?></span>
 						</a>
 					</li>
--- a/classes/class-pmpro-site-health.php
+++ b/classes/class-pmpro-site-health.php
@@ -272,6 +272,11 @@ class PMPro_Site_Health {
 				$gateway_text .= ' (' . __( 'API Version', 'paid-memberships-pro' ) . ': ' . PMPRO_STRIPE_API_VERSION . ')';
 			}
 
+			$connection_error = $stripe->get_connection_error();
+			if ( ! empty( $connection_error ) ) {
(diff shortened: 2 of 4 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3783 — Detect rejected Stripe credentials and offer a reconnect**
*dparker1005 → v3.9 | 2 files, +255 -28*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3783>

---

**Summary**
The detection, caching, settings-page status, admin notice, and webhook-notice suppression are all implemented correctly and verified against the head commit. Not ready to merge as described: the dashboard checklist integration the PR description promises is not in the diff. Fix that or update the description, then this is good to go.

---

**Issues**

- **Major** `classes/gateways/class.pmprogateway_stripe.php:3240` — `get_cached_connection_error()` is defined but never called. At head `31960495`, `git grep get_cached_connection_error` returns only the definition, and `includes/functions.php:3319-3356` (`pmpro_is_ready()`) still sets `$pmpro_gateway_ready` purely from whether the key options are non-empty. So "Configure Payment Settings" keeps its checkmark on a rejected key, contradicting the description. Either wire the cache-only helper into the checklist condition in `adminpages/metaboxes/welcome.php` (leaving `pmpro_is_ready()` alone, as the description argues), or drop the unused method and the claim from the description and changelog.

- **Minor** `classes/gateways/class.pmprogateway_stripe.php:1713` — `show_stripe_webhook_setup_notice()` has no capability check, and the PR adds a live Stripe probe to it. Any logged-in user who loads `admin.php?page=pmpro-anything` triggers the probe via `admin_notices` before the page's own cap check runs. Exposure is bounded by the transient and matches the pre-existing `does_webhook_exist()` call, but since the new sibling `show_stripe_connection_error_notice()` gates on `manage_options`, add the same one-liner here.

- **Minor** `classes/class-pmpro-site-health.php:275` — `get_gateway()` was pure option reads and now makes a (cached) outbound Stripe call. Site Health runs over REST and WP-CLI where timeouts bite. Consider `get_cached_connection_error()` here instead. That would also give the currently unused helper a caller.

---

**Looks Good**
- `get_connection_error()` caches `array( 'error' => false )` and checks with `array_key_exists`, so a healthy result is not mistaken for a cache miss. The false-vs-unset transient trap is avoided.
- Classification by HTTP status with 401 only, key passed per request via `array( 'api_key' => ... )`. A narrow Restricted Key returning 403 is correctly not flagged, and it survives a foreign stripe-php load.
- Webhook notice bails early on a connection error rather than special-casing the lookup result. Correct placement.
- `get_connect_url()` preserves the `pmpro_stripe_connect_url` filter, the `test`/`live` naming, and the nonces. The same-account guard sits after the missing-field check so there is no undefined-index risk, and it returns false on first connect.
- All error-message output paths are escaped.

---

**Questions**
1. Was the checklist wiring deliberately deferred, or lost between drafts?
2. The guard forces disconnect-then-reconnect to switch accounts, and disconnect revokes access for every site on that Stripe account. Is that trade-off acceptable, or should the error copy warn about it?

dparker1005:
Closing in favor of #3792

PR #3784, opened 2026-09-06
fix(wizard): require pmpro_wizard cap in save handler

### LINKED ISSUE #3761: Setup wizard save handler has no capability check
Minor hardening from an authorization audit.

`adminpages/wizard/save-steps.php` (hooked to `admin_init`) verifies per-step nonces and returns on failure — correct as far as it goes — but never checks a capability. The handler creates pages, saves site settings, and can initiate Stripe connect.

Today the nonces are only printed on the wizard page (requires the `pmpro_wizard` cap), so this holds — but it's one nonce leak (future template change, or a user granted `pmpro_wizard` but not settings caps) away from a low-privilege user creating pages and overwriting site/payment settings.

**Fix:** add `if ( ! current_user_can( 'pmpro_wizard' ) ) return;` (or `manage_options`) at the top of the handler.

faisalahammad:
### All Submissions:

`pmpro_init_save_wizard_data()` runs on `admin_init` and handles all wizard step saves. It verified the per-step nonces but never checked a capability, while the handler creates pages, saves site and payment settings, writes membership levels, can start Stripe connect, and can install the Update Manager plugin. The nonces are only printed on the wizard page today, but that is one nonce leak or a `pmpro_wizard`-without-settings-caps role grant away from a low-privilege user triggering those writes.

This PR adds a `current_user_can( 'pmpro_wizard' )` check at the top of the handler, right after the page check. It uses the same capability that gates the wizard page itself (the `add_submenu_page()` registration), so the save handler can no longer run for a user who could not load the page. The check also covers the `step=done` cleanup, which previously deleted an option and reset `wizard_step` on any GET request with no nonce and no submit check.

Repro on the old code: as a wp-admin user without `pmpro_wizard` (e.g. after `wp cap remove administrator pmpro_wizard`), request `wp-admin/admin.php?page=pmpro-wizard&step=done&submit=1`. The wizard completion data is reset silently. With this change the request is denied.

Resolves #3761.

1. Log in as an administrator and complete the setup wizard from Membership > Dashboard > "Set Up Membership Site". Each step should save and redirect as before.
2. Remove the `pmpro_wizard` capability from a role (e.g. `wp cap remove administrator pmpro_wizard`) and log in as that user.
3. Request `wp-admin/admin.php?page=pmpro-wizard&step=done&submit=1` directly. You should get a "Sorry, you are not allowed to access this page." error instead of a silent wizard data reset.
4. Re-add the capability (`wp cap add administrator pmpro_wizard`) and confirm the wizard works again.

### Other information:

### Changelog entry

SECURITY: Added a missing capability check to the setup wizard save handler so it can only run for users with the `pmpro_wizard` capability. #3761 (@faisalahammad)

### DIFF
--- a/adminpages/wizard/save-steps.php
+++ b/adminpages/wizard/save-steps.php
@@ -11,6 +11,11 @@ function pmpro_init_save_wizard_data() {
 		return;
 	}
 
+	// Check that the current user can access the wizard.
+	if ( ! current_user_can( 'pmpro_wizard' ) ) {
+		wp_die( esc_html__( 'Sorry, you are not allowed to access this page.', 'paid-memberships-pro' ) );
+	}
+
 	// Clear things up on the completed page.
 	if ( ! empty( $_REQUEST['step'] ) && $_REQUEST['step'] === 'done' ) {
 		delete_option( 'pmpro_wizard_collect_payment' );


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3784 — fix(wizard): require pmpro_wizard cap in save handler**
*faisalahammad → dev | 1 file, +5 -0*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3784>

---

**Summary**
Correct, minimal fix for a real authorization gap. Ready to merge. `pmpro_init_save_wizard_data()` is registered unconditionally on `admin_init` (`adminpages/wizard/save-steps.php:300`) and on `dev` has no capability check at all, so any authenticated user could hit `admin.php?page=pmpro-wizard&step=done` and trigger `delete_option( 'pmpro_wizard_collect_payment' )` plus `pmpro_setOption( 'wizard_step', 'done' )` with no nonce and no submit check. The new `current_user_can( 'pmpro_wizard' )` guard sits after the page bail and before that cleanup block, so it covers every state-changing path in the handler with one check.

Verified on `dev`:
- `pmpro_wizard` is the exact capability used to register the wizard submenu (`includes/adminpages.php:114`), so page access and save access are now gated identically.
- `pmpro_wizard` is in the PMPro capability list (`includes/capabilities.php:55`) and `pmpro_check_admin_capabilities` runs on `admin_init` at priority 5 (`includes/capabilities.php:22`), ahead of this handler's default priority 10. No ordering hazard on a fresh install's first admin load.
- No other call site invokes the function; it is only reached via the single `add_action`. No tests reference wizard save behavior, so nothing to update.

---

**Issues**
- **[Minor]** `adminpages/wizard/save-steps.php:20-23` — The `step=done` cleanup still runs on a plain GET with no nonce and no `submit` check. The new capability check shrinks the exposure to users who can already use the wizard, and the impact is limited to marking the wizard complete, so this is not blocking. Worth a follow-up: gate the cleanup behind a nonce on the "done" redirect, or move it under the `submit` branch like the other steps, so a linked or cross-site URL can't silently finish a capable admin's wizard.

---

**Looks Good**
- Fix is placed upstream of both the `step=done` cleanup and the per-step submit branches rather than patched into each `wizard-action` block. One check, whole handler.
- Reuses the existing `pmpro_wizard` capability instead of widening to `manage_options`, so correctly provisioned roles see no behavior change.
- `wp_die( esc_html__( ... ) )` matches WordPress core's permission-denied pattern, and since `wp_die` terminates, there is no fall-through risk.
- PR description includes a concrete repro for the vulnerable path and test steps for both states.

PR #3785, opened 2026-09-07
fix(orders): full-width expanded rows on mobile

### LINKED ISSUE #3713: Admin Orders Listing Layout Broken When Viewing on Mobile Browsers (Chrome, Firefox)
**Describe the bug**
After login as admin, when viewing the Orders Listing page in mobile browser (Firefox, Chrome), the layout is broken (refer attached screenshot)

**To Reproduce**
1. Using mobile browser, login as admin user
2. Visit Orders Listing page (wp-admin/admin.php?page=pmpro-orders)
3. In mobile view, you will see the ID and Code column
4. Click the expand icon on the right of a record
5. The details of the record are shown but the layout are broken. Each field should be shown in full width of the table, but currently it only take up about 25% width of the displayable area

**Screenshots**
See attached

**Expected behavior**
Each field should be shown in full width of the table

- [X ] I have deactivated other plugins and confirmed this bug occurs when only Paid Memberships Pro plugin is active.
- [X ] This bug happens with a default WordPress theme active, or [Memberlite](https://memberlitetheme.com).
- [ X] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
<details>
```
WordPress 7.0
Paid Memberships Pro 3.8
PHP 8.1

<img width="738" height="485" alt="Image" src="https://github.com/user-attachments/assets/d66cc7c4-64b9-4beb-9383-64c1fa3888c1" />

```
</details>

faisalahammad:
### All Submissions:

On mobile (viewport 782px or narrower), WordPress core renders list table rows as flex containers and already gives expanded cells a full-width flex basis (`tr td:nth-child(n+3)`). The Orders table has no `cb` column, so the ID cell is the odd one out: as the first cell in the row it keeps its default sizing and shares the first line with the primary column instead of taking the full width (screenshot in #3713).

This change adds one small orders-scoped rule to the existing `@media only screen and (max-width: 782px)` section of `css/admin.css`:

```css
.pmpro_admin-pmpro-orders .wp-list-table .is-expanded td,
.pmpro_admin-pmpro-orders .wp-list-table .is-expanded th {
	flex: 0 1 100%;
}
```

Every cell in an expanded Orders row now takes a full row on mobile, so expanded fields (User, Level, Total, Billing, Gateway, Transaction IDs, Status, Date) span the full table width. The rule is scoped to `.is-expanded`, which the core `toggle-row` script only ever adds to `tbody` rows, so the header row (`thead`/`tfoot`) and the collapsed mobile view are untouched. Both `td` and `th` are targeted because on newer WordPress versions the primary column cell is rendered as a `th` (see the WordPress 7.1 field guide note on list table row headers). Desktop rendering and all other admin pages (the rule is scoped to `.pmpro_admin-pmpro-orders`) are untouched. CSS only, no PHP or JS changes.

The same root cause likely affects the Members, Subscriptions, and Discount Codes list tables. Left out of this PR to keep the change scoped to the reported issue; happy to extend it in a follow-up.

Resolves #3713.

1. Go to Memberships > Orders on desktop and confirm the table renders exactly as before with all columns.
2. Narrow the viewport below 782px (or use a phone browser). Confirm the collapsed view still shows ID and Code with the expand toggle, and the header row still renders as a single line, same as before.
3. Tap the expand toggle on a row. Expected result: every field in the expanded section takes the full table width instead of about 25%.
4. Tap the toggle again to collapse, expand a different row, and confirm the same behavior.
5. Open the Members, Subscriptions, and Discount Codes pages on mobile and confirm their layout is unchanged (the new rule does not apply to them).

### Other information:

Tested manually on a WP install at 375px and desktop widths. The test confirmation did not include the WordPress version. The flex row layout used by this fix is in WordPress 7.1 and later.

### Changelog entry

BUG FIX: Order fields in expanded rows on the Orders list now take the full table width on mobile. #3713 (@faisalahammad)

### DIFF
--- a/css/admin.css
+++ b/css/admin.css
@@ -4812,6 +4812,17 @@ button.pmpro_report_th_closed:before {
 		word-wrap: break-word;
 	}
 
+	/* Orders list expanded rows: each cell (including the primary cell) takes the full row width. Fixes #3713. */
+	.pmpro_admin-pmpro-orders .wp-list-table tr:not(.inline-edit-row):not(.no-items) td,
+	.pmpro_admin-pmpro-orders .wp-list-table tr:not(.inline-edit-row):not(.no-items) th {
+		flex: 0 1 100%;
+	}
+
+	.pmpro_admin-pmpro-orders .wp-list-table tr:not(.inline-edit-row):not(.no-items) td.column-primary,
+	.pmpro_admin-pmpro-orders .wp-list-table tr:not(.inline-edit-row):not(.no-items) th.column-primary {
+		flex: 1 1 100%;
+	}
+
 	.wrap.pmpro_admin .row-actions {
 		align-items: center;
 		color: #a7aaad;
--- a/css/admin.css
+++ b/css/admin.css
@@ -4812,17 +4812,6 @@ button.pmpro_report_th_closed:before {
 		word-wrap: break-word;
 	}
 
-	/* Orders list expanded rows: each cell (including the primary cell) takes the full row width. Fixes #3713. */
-	.pmpro_admin-pmpro-orders .wp-list-table tr:not(.inline-edit-row):not(.no-items) td,
-	.pmpro_admin-pmpro-orders .wp-list-table tr:not(.inline-edit-row):not(.no-items) th {
-		flex: 0 1 100%;
-	}
-
-	.pmpro_admin-pmpro-orders .wp-list-table tr:not(.inline-edit-row):not(.no-items) td.column-primary,
-	.pmpro_admin-pmpro-orders .wp-list-table tr:not(.inline-edit-row):not(.no-items) th.column-primary {
-		flex: 1 1 100%;
-	}
-
 	.wrap.pmpro_admin .row-actions {
 		align-items: center;
 		color: #a7aaad;
@@ -4908,6 +4897,12 @@ button.pmpro_report_th_closed:before {
 	.pmpro_responsive_table tbody tr td::before,
 	.pmpro_responsive_table tfoot tr td::before {
 		content: attr(data-colname) ": ";
+	}
+
+	/* Expanded Orders rows: every cell takes the full width. Fixes #3713. */
+	.pmpro_admin-pmpro-orders .wp-list-table .is-expanded td,
+	.pmpro_admin-pmpro-orders .wp-list-table .is-expanded th {
+		flex: 0 1 100%;
 	}
 
 }


flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3785 — fix(orders): full-width expanded rows on mobile**
*faisalahammad → dev | 1 file, +11 -0*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3785>

---

**Summary**
Ready to merge. Small, additive CSS-only fix, correctly nested inside the existing `@media only screen and (max-width: 782px)` block opened at `css/admin.css:4768` (verified on `dev`: the insertion point sits between the `.pmpro_report-holder` rule and `.wrap.pmpro_admin .row-actions`, both inside that media query, so no brace mismatch). The premise holds: WordPress core's `wp-admin/css/list-tables.css` sets `.wp-list-table tr { display: flex; flex-wrap: wrap; }` at the 782px breakpoint (checked in the local WP trunk mirror, `7.2-alpha`, `list-tables.css:1935-1938`), so `flex` on the cells is a real override, not a no-op. Scoping to `.pmpro_admin-pmpro-orders` matches the body-class convention already used for orders overrides elsewhere in this file.

Context on why the shrink happens, for whoever reviews the promised follow-up: core gives `tr td:nth-child(n+3)` a full-width flex basis but leaves `td:nth-child(1)` and the primary cell at `auto` / `1 1 0`. The Orders table has no `cb` column and its primary column is `order_code` (`classes/class-pmpro-orders-list-table.php:245-252`), so the ID cell in position 1 is the odd one out. The PR's blanket rule covers that case cleanly.

---

**Looks Good**
- Selector scope mirrors core's own `tr:not(.inline-edit-row):not(.no-items)` pattern exactly, so it only ever competes with the rules it intends to override.
- Test plan explicitly checks Members, Subscriptions, and Discount Codes remain untouched, and calls out the same root cause there for a follow-up rather than silently widening scope.
- Verified `get_columns()` (`class-pmpro-orders-list-table.php:146-159`) has no checkbox column, so the broad `td, th` selector can't full-width an empty `check-column` cell.

---

**Questions**
1. The `th` half of both rules matches the `<thead>`/`<tfoot>` header cells too, including `th.column-primary`. Core hides `th.column-primary ~ th` at this breakpoint, but ID's `th` precedes the primary and stays visible. Did you confirm the header row still renders as one line (ID / Code side by side or stacked, either intentionally) after this change, not just the body rows? Screenshot on the PR would settle it.
2. Non-blocking: with `flex-basis: 100%` already forcing one cell per line, `flex-grow` 0 vs 1 between the two rules has no visible effect. If the split isn't intentional, collapse to a single rule for `td, th`.
3. Which WP version did you test on? The flex `tr` behavior is what the fix depends on, and I could only verify it against WP trunk locally. If you tested on the same 7.0 the reporter used, note it in the PR so nobody chases a "no-op" theory later.

faisalahammad (commented):
Thanks for the review.

1. I changed the rule to match `.is-expanded` rows only. WordPress core adds this class to tbody rows, so the header row is not changed. I confirmed the header stays on one line.
2. I collapsed the two rules into one rule with `flex: 0 1 100%`.
3. The flex list table layout used by this fix was added in WordPress 7.1. The test confirmation did not include the site WordPress version, so I did not add an unverified version to the PR. The rule is harmless on WordPress 7.0, where core uses the older block layout.

kimcoleman:
@faisalahammad can you share screenshots of what changes you are seeing with this PR in place? The only change I see is when the ID row is shown (need to turn it on in screen options), this PR makes the "code" row jump to a new line when expanded. That feels like a negative experience when toggling.

Without your code (collapsed):
Screenshot 2026-09-17 at 1 18 00 PM

Without your code (expanded):
Screenshot 2026-09-17 at 1 17 57 PM

With your code (collapsed):
Screenshot 2026-09-17 at 1 17 31 PM

With your code (expanded) - the row jumps:
Screenshot 2026-09-17 at 1 17 35 PM

kimcoleman:
Ok - I've found it testing on other list tables. I see what happens specifically relies on the column after the column-primary being one like an ID column that has very little text and only at specific screen widths.

This seems like an extremely small edge case, but it seems safe to add. I'm going to scope this to all wp-list-tables in the `.pmpro_admin` wrapper class instead of targeting orders only then I'll merge it.

Screenshot 2026-09-17 at 1 37 24 PM

kimcoleman (approved):
Made the change to expand this - edge case that it comes up. I did another PR that fixes some other responsive issues introduced in the WP 7.0+ admin "redesign".

kimcoleman:
@faisalahammad to continue contributing to PMPro, please check the box to "Allow Edits From Maintainers" when opening PRs in our organization's repos.

We need to be able to push changes to PRs before merging them.

For now when, we have these changes, we will be closing your PR and opening a new one crediting you for the change with our fixes.

Issue #3634, opened 2026-04-09
Feature Request: Decouple "View As" Toolbar from manage_options Capability  Current Behavior


kimwhite:
**Current Behavior**

The "View As" toolbar feature in PMPro is gated behind the manage_options capability (see adminpages.php line 458). This means only Administrators can use it out of the box. Site owners who want to grant this tool to other roles (e.g., Editors, Support Agents) must grant manage_options — which also unlocks the PMPro License settings page and core WordPress settings pages as a side effect.

**Requested Change**

Introduce a dedicated PMPro capability for the "View As" feature — something like pmpro_view_as — and gate the toolbar on that cap instead of manage_options.

This would allow:
- Admins to grant pmpro_view_as to specific roles without elevating their broader permissions
- Compatibility with role management plugins (Members, User Role Editor, etc.) for fine-grained control
- A cleaner, more intentional permission model consistent with how other PMPro admin pages are already gated

**Suggested Implementation**

1. Add pmpro_view_as to the capabilities assigned via pmpro_set_capabilities_for_role() for the Administrator role
2. Change the gate in adminpages.php from manage_options to current_user_can( 'pmpro_view_as' )
3. Document the new cap so site owners know they can grant it selectively

**Why This Matters**

Membership sites often have support staff, community managers, or content editors who need to preview the member experience — but should not have access to payment settings, license keys, or WordPress core options. Right now, there's no clean way to enable this without a workaround snippet and an unintended capability grant.

PR #3786, opened 2026-09-07
fix(caps): add pmpro_view_as capability for the View As toolbar

faisalahammad:
### All Submissions:

The "View As" (Admin Membership Access) toolbar in the admin bar is gated on `manage_options`, so an Editor or support staff member who needs to preview the member experience must also be granted full site settings access, including WP Settings and the PMPro License page. This was raised in #3634.

This PR adds a dedicated `pmpro_view_as` capability and gates both sides of the feature on `manage_options` OR `pmpro_view_as`, following the existing dual-capability pattern used elsewhere in the plugin:

- `pmpro_admin_membership_access_menu_bar()` in `includes/adminpages.php` now shows the toolbar for users with either capability.
- `pmpro_getMembershipLevelsForUser()` in `includes/functions.php` applies the same gate so the preview actually takes effect for these users, not just the toolbar.

`pmpro_view_as` is added to the administrator capability list in `pmpro_get_capability_defs()`. Since `pmpro_check_admin_capabilities()` re-applies the administrator caps on `admin_init` when any are missing, existing sites pick up the new capability automatically on the next dashboard load with no upgrade script. Site owners can grant `pmpro_view_as` to Editor or a custom support role (e.g. with Members or User Role Editor) without handing over `manage_options`. Multisite super admins keep working through the `manage_options` check.

Note for site owners: the "View With Access" preview uses `pmpro_getAllLevels( true )`, which includes hidden levels. Granting `pmpro_view_as` therefore also lets the holder preview content restricted to hidden levels. This is the same behavior admins already had, but it is worth knowing when deciding who to grant the capability to. Restricting non-admins to visible levels would be a reasonable follow-up if the team prefers a narrower grant; this PR does not change that behavior.

1. As an administrator, load any frontend page and confirm the "View As" toolbar still appears and all three modes (with access, current levels, no access) work as before.
2. Grant the `pmpro_view_as` capability to the Editor role (e.g. with the Members plugin) and log in as an Editor. The toolbar should now appear and the preview should change what content and levels that user sees on the frontend.
3. As that same Editor, confirm they still cannot access WP Settings, the PMPro settings pages, or the PMPro License page.
4. View the Account page as the Editor and confirm it still shows their real membership levels (the preview correctly does not apply there, same as for admins).
5. On a multisite install, confirm a super admin still sees the toolbar via `manage_options`.

### Other information:

The changelog entry is kept in this description only. `CHANGELOG.txt` and `readme.txt` are not touched by this PR.

### Changelog entry

ENHANCEMENT: Added a new `pmpro_view_as` capability so the "View As" toolbar in the admin bar can be granted to roles that should not have `manage_options`. #3634 (@faisalahammad)

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,7 @@
 == Changelog ==
+= TBD =
+* ENHANCEMENT: Added a new `pmpro_view_as` capability so the "View As" toolbar in the admin bar can be granted to roles that should not have `manage_options`. #3634 (@faisalahammad)
+
 = 3.8.5 - 2026-08-24 =
 * SECURITY: Fixed an issue where the payment amount was not re-verified after Strong Customer Authentication (3D Secure) in the Stripe on-site checkout. #3767 (@proxydom)
 * SECURITY: Fixed an access check bypass in the `getfile` service that could allow protected files to be served through `admin-ajax.php` without a membership access check. #3762 (@dparker1005)
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,7 +1,4 @@
 == Changelog ==
-= TBD =
-* ENHANCEMENT: Added a new `pmpro_view_as` capability so the "View As" toolbar in the admin bar can be granted to roles that should not have `manage_options`. #3634 (@faisalahammad)
-
 = 3.8.5 - 2026-08-24 =
 * SECURITY: Fixed an issue where the payment amount was not re-verified after Strong Customer Authentication (3D Secure) in the Stripe on-site checkout. #3767 (@proxydom)
 * SECURITY: Fixed an access check bypass in the `getfile` service that could allow protected files to be served through `admin-ajax.php` without a membership access check. #3762 (@dparker1005)
--- a/includes/adminpages.php
+++ b/includes/adminpages.php
@@ -455,7 +455,7 @@ function pmpro_admin_membership_access_menu_bar() {
 	}
 
 	// View menu at all?
-	if ( ! current_user_can( 'manage_options' ) || ! is_admin_bar_showing() ) {
+	if ( ( ! current_user_can( 'manage_options' ) && ! current_user_can( 'pmpro_view_as' ) ) || ! is_admin_bar_showing() ) {
 		return;
 	}
 
--- a/includes/capabilities.php
+++ b/includes/capabilities.php
@@ -75,7 +75,8 @@ function pmpro_get_capability_defs($role)
 		'pmpro_discountcodes',
 		'pmpro_userfields',
 		'pmpro_updates',
-		'pmpro_manage_pause_mode'
+		'pmpro_manage_pause_mode',
+		'pmpro_view_as'
 	);
 
     return apply_filters( "pmpro_assigned_{$role}_capabilities", $cap_array);
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2564,8 +2564,8 @@ function pmpro_getMembershipLevelsForUser( $user_id = null, $include_inactive = 
 	// make sure user id is int for security
 	$user_id = intval( $user_id );
 
-	// Admins have special rules for membership levels. Check them here.
-	if ( $user_id == $current_user->ID && current_user_can( 'manage_options' ) ) {
+	// Admins and users with the View As capability have special rules for membership levels. Check them here.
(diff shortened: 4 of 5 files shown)


dparker1005 (changes requested):
Hi @faisalahammad, thank you for picking this up. The capability change itself looks good and is a clean answer to #3634. We'd like to get it in, so I've retargeted the PR at our `v3.9` release branch and added it to the 3.9 milestone.

The PR is currently showing merge conflicts, and they're coming from the CHANGELOG.txt and readme.txt edits. Please don't add changelog entries in PRs. We assemble the changelog at release time, and any edit to the top of that file conflicts with every other open PR as soon as a release ships. Could you revert those two files and put the entry in the PR description instead? We'll pull it into the changelog from there when we release.

Once the changelog files are reverted, this should merge cleanly against `v3.9`.

flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3786 — fix(caps): add pmpro_view_as capability for the View As toolbar**
*faisalahammad → v3.9 | 5 files, +11 -4*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3786>

---

**Summary**
The capability change is correct and narrowly scoped. Both gates are OR'd consistently, the preview only ever reads and writes the current user's own meta, and the administrator auto-adoption path via `pmpro_check_admin_capabilities()` on `admin_init` works as described. Not mergeable yet only because the changelog and readme edits dparker1005 asked to revert are still in the diff at head `e9f60c1`.

---

**Issues**

- **Major** `CHANGELOG.txt:2-4`, `readme.txt:213-215` — dparker1005's review asked for these two files to be reverted because changelog assembly happens at release and top-of-file edits conflict with every other open PR. The current head still contains both edits, and the PR still reports merge conflicts from them. Revert both files. The entry already lives in the PR description's "Changelog entry" section, which is where the team pulls it from.

- **Minor** `includes/functions.php:2591` — In "View With Access" mode the preview returns `pmpro_getAllLevels( true )`, which includes hidden levels. That means `pmpro_view_as` grants frontend read access to content gated to hidden levels too. Admins already had this, but bundled with `manage_options`. No code change required; consider noting in the docs for this capability what a site owner is actually handing to support staff. Restricting non-`manage_options` holders to visible levels would be a reasonable alternative if the team wants a narrower grant.

---

**Looks Good**

- `includes/adminpages.php:458` and `includes/functions.php:2568` are gated on the identical condition, so the set of users who can see the toggle is exactly the set for whom toggling has effect. Half-applying this (toolbar only, or preview only) is the easy mistake here, and the PR avoids it.
- The write path in `adminpages.php:463-470` sits behind the new gate and keeps its `check_admin_referer()` plus the `yes`/`current`/`no` whitelist. No new CSRF or injection surface.
- The `$user_id == $current_user->ID` guard at `functions.php:2568` is untouched, so a `pmpro_view_as` holder cannot preview or alter another user's levels.
- The page-exclusion list (account, billing, checkout, invoice, levels, etc.) and the `pmpro_disable_admin_membership_access` filter apply unchanged to the new capability, so transactional pages still show real levels.
- `pmpro_isAdmin()` at `functions.php:1741` still checks only `manage_options`. Its single caller shows a settings link at checkout when no gateway is configured. Leaving it alone is correct: a `pmpro_view_as` user cannot reach settings anyway.
- Stale `pmpro_admin_membership_access` meta is harmless if the capability is later removed from a role, since the branch that reads it never executes.

---

**Questions**

1. dparker1005 retargeted this at `v3.9` rather than `dev`. Is that the intended landing branch for the rest of the 3.9 cycle, so future caps changes should follow suit?

faisalahammad:
Reverted `CHANGELOG.txt` and `readme.txt` in b472fbd. The conflict came from the `= 3.8.6 - 2026-09-08 =` section landing on `v3.9` at the same lines where the `= TBD =` entry was added. Both files are now back to their state before this PR, and the entry stays in the PR description where the release changelog is assembled from. GitHub reports the PR as mergeable into `v3.9` now, and the diff is down to the 3 code files.

On the hidden levels note: no code change, as you said. I added a short note to the description explaining that `pmpro_view_as` also lets the holder preview content gated to hidden levels, since the "View With Access" preview uses `pmpro_getAllLevels( true )`. Site owners can weigh that when deciding who to grant the capability to. Happy to narrow it to visible levels for non-admins if the team wants that instead.

On the landing branch question: `v3.9` makes sense for this one, since you retargeted it and put it in the 3.9 milestone.

dparker1005 (approved):
Changelog files reverted, diff is down to the three code files. Approving.

Issue #3778, opened 2026-08-28
Review PMPro admin script/data loading on every wp-admin screen


flintfromthebasement:
## Summary

`pmpro_admin_enqueue_scripts()` is attached to `admin_enqueue_scripts` without checking the current screen or user capability. As a result, every wp-admin page available to any logged-in user loads Select2, `pmpro-admin.js`, PMPro admin CSS, and a localized global `pmpro` object.

The localized object currently includes:

- every level object from `pmpro_getAllLevels( true, true )`, augmented with formatted price, expiration, and description;
- level ID/name choices;
- rendered blank user-field group and field HTML;
- checkout and REST URLs;
- Stripe webhook and `wp_rest` nonces.

This is not automatically a privilege escalation: a REST nonce authenticates the current user but does not grant capabilities, and PMPro's REST routes still need correct permission callbacks. However, the broad loading expands the data/nonce exposure and runs level formatting plus user-field rendering on unrelated admin screens for low-privilege users.

## Why this needs a product decision

A blind "PMPro pages only" scope is risky. The same localized level arrays are consumed by PMPro blocks in the post editor, and `pmpro-admin.js` also supports PMPro UI outside the main settings screens. The data could instead be split/localized by consumer, or the enqueue could use an explicit allowlist that includes PMPro screens, relevant profile/editor screens, and any other supported integrations.

## Questions

1. Which non-PMPro admin screens intentionally require `pmpro-admin.js`, Select2, and PMPro admin CSS?
2. Do authors/editors need full level objects and user-field template HTML, or only the smaller level ID/name list used by blocks?
3. Can the REST and Stripe webhook nonces be localized only on screens where their corresponding UI exists?
4. Should the expensive level/user-field payload be split from the base admin script so unrelated screens get neither the query/rendering cost nor the data?

## Relevant code

- `includes/scripts.php`: `pmpro_admin_enqueue_scripts()` and its unconditional `admin_enqueue_scripts` hook.
- `js/pmpro-admin.js`: Stripe webhook actions, user-field builder, email log popup, and quick search.
- `blocks/src/`: several editor blocks consume `pmpro.all_level_values_and_labels` or `pmpro.all_levels_formatted_text`.

dparker1005:
A status note so anyone interested in this has the current thinking. We reviewed a contributor PR in #3787 that scoped admin assets to PMPro pages plus an allowlist of editor and profile screens, and closed it.

[...]

PR #1724, opened 2021-08-16
Ensure our admin scripts/styles only load on our pages

sc0ttkclark:
### All Submissions:

* [x] Have you followed the [Contributing guideline](https://github.com/strangerstudios/paid-memberships-pro/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/strangerstudios/paid-memberships-pro/pulls) for the same update/change?

### Changes proposed in this Pull Request:

Resolves #1723.

### How to test the changes in this Pull Request:

1. Load any admin page that isn't ours
2. Confirm our admin script/css is not loaded
3. Load any admin page that is ours
4. Confirm our admin script/css is loaded

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you successfully run tests with your changes locally?

### Changelog entry

> Ensure our admin scripts/styles only load on PMPro admin pages

PR #3787, opened 2026-09-07
fix(admin): only load admin assets where needed

faisalahammad:
### All Submissions:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/strangerstudios/paid-memberships-pro/pulls) for the same update?

PMPro currently enqueues `pmpro-admin.js`, `admin.css`, Select2, and the localized `pmpro` object on every wp-admin screen. Any admin loading a plain post edit screen or a settings page of another plugin downloads Select2 plus a payload that includes membership levels, checkout URLs, and REST nonces. That is unnecessary load and it exposes more client side data than those screens need. This was reported in #3778, and an earlier fix for the same problem (#1724) had regressed.

This change rewrites `pmpro_admin_enqueue_scripts()` so assets load only where they are actually used:

- PMPro admin pages (any `$_REQUEST['page']` starting with `pmpro-`): full payload.
- Editor screens (`post.php`, `post-new.php`, `widgets.php`, `customize.php`, `site-editor.php`): assets load with a reduced payload, because the PMPro blocks and widget controls read `all_level_values_and_labels`, `all_levels_formatted_text`, and `checkout_url` from the localized object.
- User profile screens (`profile.php`, `user-edit.php`): Select2 and `admin.css` only, needed by PMPro select2 user fields and the membership table styling.
- Everything else: no PMPro admin assets.

Inside the payload, page specific keys are now only localized on the pages that use them: `user_fields_blank_group`/`user_fields_blank_field` on `pmpro-userfields`, `stripe_webhook_nonce` on `pmpro-paymentsettings`, `rest_url`/`nonce` on PMPro pages (quick search and email log popup). `all_levels` and `plugin_updated_successfully_text` have no core JS consumers but are kept for add-on backwards compatibility. The duplicate `pmpro_getAllLevels( true, true )` call is also removed; levels were fetched twice per page load.

Two supporting changes:

- The admin bar Memberships icon was styled inside `admin.css`, so the icon would have lost its image on non PMPro screens. The `#pmpro-ab-icon` rule is moved to a new `css/admin-bar-icon.css` that is enqueued before the gate, only for users with `pmpro_memberships_menu` when the admin bar shows.
- A new `pmpro_load_admin_assets` filter lets add-ons that rely on PMPro admin assets on their own screens force the load.

1. As an administrator, open any non PMPro wp-admin screen (Posts, Settings, Plugins) with the browser dev tools network tab open. Expected result: no Select2, no `pmpro-admin.js`, no PMPro admin CSS except the small admin bar icon style, and no `pmpro` JS global.
2. Open Memberships > Dashboard and a few other PMPro pages (Orders, Settings, Memberships > Settings). Expected result: everything behaves as before, including the orders level filter dropdown (Select2) and the admin quick search.
3. On Memberships > User Fields, edit a group/field name inline. Expected result: inline editing still works (this page supplies the blank group/field templates to JS).
4. On Memberships > Payment Settings with Stripe, confirm the Stripe settings save and the webhook status refresh still works (page supplies `stripe_webhook_nonce`).
5. Edit a post with a PMPro block (e.g. Checkout Page) in the block editor. Expected result: the level dropdowns in the block sidebar still populate (the editor screen keeps the reduced payload).
6. Open Appearance > Widgets with a PMPro block and the Customizer with a PMPro block. Expected result: block controls still render and function.
7. Edit a user profile that has a PMPro select2 user field and a membership level row. Expected result: the field and table render styled and functional.
8. On any admin screen, confirm the Memberships icon in the admin bar still shows its logo image.
9. Load `wp-admin/admin.php?page=pmpro-wizard&step=done` and confirm the confetti behavior is unchanged on a fresh wizard completion.

### Other information:

There is no automated test suite in this repository, so I verified with the manual steps above on a local WordPress install, plus PHPCS (`WordPress-Core`/`WordPress-Docs`) on the changed files and `npm run lint:css` on the new stylesheet.

One compatibility note for reviewers: add-ons that read the removed keys or use the REST nonce from `pmpro` on non PMPro admin screens would lose that data. The new `pmpro_load_admin_assets` filter is the documented escape hatch, e.g. `add_filter( 'pmpro_load_admin_assets', '__return_true' )` on their screens.

### Changelog entry

ENHANCEMENT: PMPro admin scripts, styles, and localized data now only load on PMPro admin pages, editor screens, and user profile screens instead of every wp-admin screen. #3778 (@faisalahammad)


dparker1005:
Hi @faisalahammad, thank you for taking on #3778. The goal is right, and pieces like the admin bar icon split, the sanitized `page` check, and dropping the duplicate `pmpro_getAllLevels()` call are all good.

I'm going to close this one, though. We talked it through internally, and scoping admin assets is a change we'd want to make deliberately and on our own terms. Our assets have loaded on every wp-admin screen for years, partly for reasons we're still reconstructing, and a lot of code inside and outside core quietly depends on that. #3778 is tagged Needs Thought for exactly this reason, and we haven't made the call yet on whether or how to do it.

The branch also shows why an allowlist approach is risky without a full consumer audit. On this branch, dismissible PMPro notices (pause mode, spam protection, deprecation warnings) lose their styling and the dismiss button on most screens, since the handler lives only in `pmpro-admin.js`. The category and tag edit screens lose the "Require Membership" checkbox styling, since `edit-tags.php` and `term.php` aren't in the list. Add Ons that check for or depend on the `pmpro_admin` handle, PMPro Downloads for one, break because registration never happens on their screens. And `customize.php` never fires `admin_enqueue_scripts`, so that entry is inert.

If we pick this up, it'll be shaped as: register handles and the `pmpro` object everywhere, gate only the enqueue, and give things like `pmpro.all_levels` a deprecation window, with a sweep of the Add Ons alongside it. That's release-planning work rather than a standalone PR. Thanks again for the effort here.

PR #3788, opened 2026-09-07
fix(multisite): store expiration notices per site

### LINKED ISSUE #3683: Consider migrating per-site user_meta to update_user_option for multisite correctness
## Context

PMPro has two multisite deployment models:

1. **`pmpro-network-subsite`** (the supported/documented model): PMPro tables
   live on the main site; subsites share them via `$wpdb` remapping. User data
   is intentionally shared network-wide.
2. **Independent PMPro per subsite**: each subsite has its own PMPro install
   with its own `pmpro_memberships_users` table. Tolerated but not really
   designed for.

Both models share `wp_usermeta` (it is a single global table). Today, PMPro
writes all of its user-level state with `update_user_meta()` — i.e. globally.
For Model 1 this is correct. For Model 2 it causes cross-subsite collisions:
e.g. issue [strangerstudios/pmpro-extra-expiration-warning-emails#39](https://github.com/strangerstudios/pmpro-extra-expiration-warning-emails/issues/39),
where two subsites with the same `membership_id` clobber each other's
"expiration notice sent" timestamp.

The right primitive for per-site user data is WP core's `update_user_option()` /
`get_user_option()`, which auto-prefixes the meta_key with
`$wpdb->get_blog_prefix()`. This is WP's native pattern for scoping user data
per-site (same mechanism core uses for capabilities, dashboard prefs, etc.).

## Proposal

Audit PMPro's `*_user_meta` calls and split them by semantic:

**Should be per-site (`update_user_option` candidates):**

- `pmpro_expiration_notice_<id>` — tied to a per-site `pmpro_memberships_users` row
- `pmpro_approval_<id>` — approval is per-(site, level)
- `pmpro_stripe_customerid` — each site has its own Stripe account
- `pmpro_braintree_customerid` — same, per-gateway-instance
- `pmpro_stripe_dont_cancel` — per-subscription tracking
- `pmpro_consent_log` — consent given to *this* site's terms
- `pmpro_admin_membership_access` — admin override per blog
- `pmpro_archived_notifications` — per-site dashboard UI state
- `pmpro_disable_notifications` — per-site admin preference
- `pmpro_dashboard_metabox_order` — WP admin widget order (per-site by convention)
- `pmpro_delete_file_*` — tied to per-site forms/uploads

**Should stay global:**

- `pmpro_avatar` — user's identity, network-wide
- Billing address fields (`pmpro_bfirstname`, `pmpro_blastname`, `pmpro_baddress`, `pmpro_baddress2`, `pmpro_bcity`, `pmpro_bstate`, `pmpro_bzipcode`, `pmpro_bcountry`) — properties of the person
- Core WP fields (`first_name`, `last_name`) — never PMPro's to namespace

## The paired change in `pmpro-network-subsite`

Switching core in isolation would break the network-subsite addon: it
currently relies on user_meta being global so that data written on the main
site is visible everywhere. If core switches a key to `update_user_option`,
subsite 2 would read `wp_2_pmpro_stripe_customerid` (empty) and behave as if
the user has no Stripe customer.

The migration should land in two coordinated steps:

1. Core swaps per-site keys to `update_user_option` / `get_user_option`.
2. `pmpro-network-subsite` wraps reads/writes in `switch_to_blog( main_site_id )`
   (or otherwise explicitly reads from main), the same way it already remaps
   `$wpdb->pmpro_memberships_users`.

After both, independent-per-subsite installs get per-site correctness for
free, and network-subsite continues to behave as it does today.

## Migration considerations

- Existing user_meta rows need to either be (a) read-back-compatible during a
  transition window or (b) migrated on upgrade. The simplest pattern: on
  read, fall back to the legacy key if the new one is absent; write only the
  new one going forward; ship a one-time upgrade routine that renames
  existing rows.
- Per-key migration paths differ slightly (e.g. `pmpro_expiration_notice_<id>`
  needs the membership IDs enumerated for each user; gateway customer IDs are
  one row per user). Worth doing one key at a time rather than as a single
  flag-day change.

## Out of scope

- "Add a column to `pmpro_memberships_users` instead" — that's a cleaner data
  model for the expiration-notice case specifically, but it is a per-key
  conversation. This issue is just about the user_meta migration pattern in
  general.
- Network-activated PMPro — that deployment is not under consideration here.

## Why now

Issue raised while reviewing
[strangerstudios/pmpro-extra-expiration-warning-emails#40](https://github.com/strangerstudios/pmpro-extra-expiration-warning-emails/pull/40)
(addon-side patch for the same multisite bug). That PR adds an
`is_multisite()`-gated meta_key prefix; if core eventually handles per-site
scoping correctly, the addon would not need to invent its own scheme.

faisalahammad:
### All Submissions:

Resolves #3683 (first step, one key at a time as suggested in the issue).

The expiration reminder timestamp is stored as user meta keyed `pmpro_expiration_notice_{membership_id}`. On a multisite network where each subsite runs its own copy of PMPro with its own `pmpro_memberships_users` table, `wp_usermeta` is still shared network-wide, so two subsites with the same membership ID overwrite each other's timestamps. A reminder sent on one site suppresses the reminder on the other, or a site re-sends a reminder the other site already sent. The same collision is what [pmpro-extra-expiration-warning-emails#39](https://github.com/strangerstudios/pmpro-extra-expiration-warning-emails/issues/39) works around on the add-on side.

This PR moves that one key to the per-site user option API:

- `send_expiration_reminder_email()` writes with `update_user_option()` and `send_membership_expired_email()` deletes with `delete_user_option()`, both of which prefix the key with `$wpdb->get_blog_prefix()` (`wp_2_pmpro_expiration_notice_5` on subsite 2).
- The reminder query in `membership_expiration_reminders()` joins on the same prefixed key, with the blog prefix passed through `$wpdb->prepare()` as a bound parameter.
- A new upgrade routine (`includes/updates/upgrade_3_8_5.php`, gated at db version 3.85) renames existing `pmpro_expiration_notice_%` rows to the prefixed key. On multisite only the main site claims the legacy rows, because a shared row carries no record of which site wrote it; subsites start fresh with their own keys. The rename is idempotent since prefixed rows no longer match the LIKE pattern.

Other keys listed in #3683 (gateway customer IDs, consent log, admin UI state) are deliberately not touched here and can follow one at a time. No change is needed in pmpro-network-subsite: it clears PMPro crons on subsites, so only the main site ever writes these notices under that deployment model.

1. On a single site with a member expiring within 7 days, confirm a `pmpro_expiration_notice_{level_id}` row exists in `wp_usermeta`, then run the upgrade. Expected result: the row is renamed to `wp_pmpro_expiration_notice_{level_id}`, the reminder sends once, and repeated cron runs do not resend it.
2. On a multisite network with PMPro active independently on the main site and a subsite, create the same level ID on both and give one user an expiring membership on each. Run the reminder task on both sites. Expected result: each site sends its own reminder and writes its own prefixed timestamp (`wp_...` on the main site, `wp_2_...` on subsite 2), and a second cron run on either site sends nothing.
3. On a network using the Multisite Membership Add On, confirm behavior is unchanged: subsite crons stay cleared and only the main site sends reminders.

### Other information:

There is no automated test suite in this repository, so I verified with the manual steps above plus PHP lint and PHPCS (`WordPress-Core`/`WordPress-Docs`) on the changed files. Worst case during the window between deployment and the upgrade running is one duplicate reminder per user, after which the prefixed key suppresses normally.

### Changelog entry

BUG FIX: Membership expiration reminder tracking is now per-site on multisite networks. The expiration notice is stored as a user option, and an upgrade script renames existing notices so each site on a network running its own PMPro copy tracks reminders separately. #3683 (@faisalahammad)

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,7 @@
 == Changelog ==
+= TBD =
+* BUG FIX: Membership expiration reminder tracking is now per-site on multisite networks. The expiration notice is stored as a user option, and an upgrade script renames existing notices so each site on a network running its own PMPro copy tracks reminders separately. #3683 (@faisalahammad)
+
 = 3.8.5 - 2026-08-24 =
 * SECURITY: Fixed an issue where the payment amount was not re-verified after Strong Customer Authentication (3D Secure) in the Stripe on-site checkout. #3767 (@proxydom)
 * SECURITY: Fixed an access check bypass in the `getfile` service that could allow protected files to be served through `admin-ajax.php` without a membership access check. #3762 (@dparker1005)
--- a/classes/class-pmpro-recurring-actions.php
+++ b/classes/class-pmpro-recurring-actions.php
@@ -143,6 +143,8 @@ class PMPro_Recurring_Actions {
 		$interval_end   = date( 'Y-m-d H:i:s', strtotime( "+{$pmpro_email_days_before_expiration} days", current_time( 'timestamp' ) ) );
 
 		// look for memberships that are going to expire within one week (but we haven't emailed them within a week)
+		// User options are stored in usermeta with the blog prefix, so the
+		// expiration notice key has to match what update_user_option() writes.
 		$sqlQuery = $wpdb->prepare(
 			"SELECT DISTINCT
 						mu.user_id,
@@ -152,7 +154,7 @@ class PMPro_Recurring_Actions {
 						um.meta_value AS notice
 					FROM {$wpdb->pmpro_memberships_users} AS mu
 					LEFT JOIN {$wpdb->usermeta} AS um ON um.user_id = mu.user_id
-					AND um.meta_key = CONCAT( 'pmpro_expiration_notice_', mu.membership_id )
+					AND um.meta_key = CONCAT( %s, 'pmpro_expiration_notice_', mu.membership_id )
 				WHERE ( um.meta_value IS NULL OR DATE_ADD(um.meta_value, INTERVAL %d DAY) < %s )
 					AND mu.status = 'active'
 					AND mu.enddate IS NOT NULL
@@ -162,6 +164,7 @@ class PMPro_Recurring_Actions {
 					AND mu.membership_id <> 0
 				ORDER BY mu.enddate
 				",
+			$wpdb->get_blog_prefix(),
 			$pmpro_email_days_before_expiration,
 			$today,
 			self::PMPRO_MAGIC_MIN_DATE,
@@ -247,8 +250,8 @@ class PMPro_Recurring_Actions {
 			}
 		}
 
-		// Update user meta so we don't email them again
-		update_user_meta( $user_id, 'pmpro_expiration_notice_' . $membership_id, current_time( 'Y-m-d H:i:s' ) );
+		// Update the user option so we don't email them again on this site.
+		update_user_option( $user_id, 'pmpro_expiration_notice_' . $membership_id, current_time( 'Y-m-d H:i:s' ) );
 	}
 
 	/**
@@ -371,8 +374,8 @@ class PMPro_Recurring_Actions {
 			$euser      = get_userdata( $user_id );
 			if ( ! empty( $euser ) ) {
 				$pmproemail->sendMembershipExpiredEmail( $euser, $membership_id );
-				// Delete the expiration notice for this membership
-				delete_user_meta( $user_id, 'pmpro_expiration_notice_' . $membership_id );
+				// Delete the expiration notice for this membership on this site
+				delete_user_option( $user_id, 'pmpro_expiration_notice_' . $membership_id );
 				if ( WP_DEBUG ) {
 					error_log( sprintf( __( 'Membership expired email sent to %s. ', 'paid-memberships-pro' ), $euser->user_email ) );
 				}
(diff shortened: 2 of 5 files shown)


dparker1005:
Hi @faisalahammad, thank you for taking a run at this. The direction (moving this key to per-site user options) is the one we want, and the write and read changes in the recurring actions class are right.

I'm going to close this one, though, because it can't ship on its own and the remaining pieces are bigger than a revision to this PR.

The main problem is the Multisite Membership Add On (pmpro-network-subsite). The PR says no change is needed there because it clears PMPro crons on subsites, but that stopped being true in PMPro 3.5. The Add On clears the legacy WP-Cron hook list, which is empty on current core. Reminders now run on `pmpro_schedule_quarter_hourly` through Action Scheduler, which is per-site and which the Add On doesn't touch, so every subsite runs the reminder query against the shared memberships table. Today the shared, unprefixed meta key is what stops those runs from sending duplicate emails. With per-site keys, each subsite would send its own reminder to the same member. That Add On has to be updated to stop subsites running the recurring hooks before core can change this key, and that work hasn't been scoped yet.

Two smaller things for whoever picks this up next:

- The migration shouldn't rename and let the main site claim the rows. Since every site currently reads the same shared row, the behavior-preserving migration is for each site to copy the legacy value into its own prefixed key on its own upgrade, so no site resends a reminder it already sent. Legacy rows can be left in place.
- The upgrade routine should be named and gated for the release it actually ships in (this one is labeled 3.8.5, which has already shipped), and the changelog files should not be edited in the PR. We assemble the changelog at release time, and those edits are what's causing the merge conflicts.

I've written the details up on #3683 so the plan is in one place.

PR #3789, opened 2026-09-08
BUG FIX: Refresh Stripe Connect publishable keys and detect rejected Stripe credentials

flintfromthebasement:
### All Submissions:

Stripe Connect currently saves the platform publishable key returned during OAuth forever. Rotating that platform key breaks on-site Stripe Elements checkout and billing updates, even though the connected account and access token remain valid.

This PR:

- Fetches the combined live/test platform-key manifest proposed in [pmpro-stripe-connect-server#1](https://github.com/strangerstudios/pmpro-stripe-connect-server/pull/1) when an on-site Connect checkout or billing page is rendered.
- Validates the complete response before atomically replacing one durable one-week cache record.
- Keeps both the prior validated manifest and the original OAuth-returned publishable key as last-known-good fallbacks when HTTP, JSON, validation, or storage fails.
- Recognizes Stripe.js `api_key_expired`, `platform_api_key_expired`, and its code-less `Invalid API Key provided` response, performs one nonce-protected refresh through `admin-ajax.php`, and reloads Elements once if the key changed.
- Records an environment-specific issue when automatic recovery fails, returns the same key, or the refreshed key is rejected again.
- Shows a contextual Payment Settings error and nonce-protected “Check for an updated Stripe key” button only while that issue is present. Manual checks also show a standard top-of-page notice that distinguishes a temporary connection failure from a completed check where the key was unchanged; a successful changed-key refresh clears the issue.
- Prevents all manifest refresh and recovery behavior from engaging when a complete local API-key pair takes precedence, including mixed states where old Connect credentials remain stored.
- Prevents reload loops with session storage and explains that payment details must be re-entered after recovery because an existing Stripe Element cannot be moved to a new Stripe instance.
- Adds filters for the manifest URL and cache lifetime.

This intentionally does not redesign the payment settings UI or establish a longer-term retry/backoff policy.

1. Configure Stripe Connect in sandbox mode and select the on-site payment flow.
2. Filter `pmpro_stripe_connect_publishable_key_manifest_url` or `pre_http_request` to return a schema-version-1 manifest with dummy live/test publishable keys.
3. Load checkout and confirm the current-mode manifest key is cached and localized to `pmpro-stripe.js`.
4. Return malformed JSON, a non-200 response, or an `sk_...` value and confirm the prior cached key remains unchanged.
5. Simulate `api_key_expired` from Stripe.js and confirm one AJAX refresh/reload occurs when the manifest key changes; repeat with the refreshed key and confirm there is no second reload.
6. Confirm a failed, unchanged, or repeatedly rejected recovery records an issue and displays the conditional check link in the active Stripe Connection section.
7. Use the link and confirm a changed key clears the issue, while an unchanged or failed response preserves the prior key and warning.
8. Configure a complete restricted-key pair while leaving Connect credentials stored. Confirm no manifest request occurs, the local keys remain effective, no Connect account is localized to Stripe.js, and the warning/link remains hidden.
9. Submit a normal card error such as `card_declined` and confirm it does not trigger key recovery.

Verified on `flintdev.test` with intercepted HTTP responses:

- valid live/test manifests cache successfully;
- invalid manifests preserve the prior cache;
- a two-day-old cache makes no additional request under the one-week lifetime;
- a missing cache refreshes lazily from the manifest;
- anonymous AJAX refresh succeeds with a valid nonce and rejects an invalid nonce with HTTP 403;
- the check link is hidden normally, appears for a detected Connect issue, and remains hidden for restricted-key sites;
- restricted keys bypass manifest HTTP and remain the effective secret/publishable-key pair even when Connect credentials remain stored;
- isolated Stripe.js tests pass for one-shot recovery, reload-loop prevention, and non-key errors;
- PHP and JavaScript syntax checks pass.

### Merged from #3783 (@dparker1005)

This PR now also includes the Stripe connection check from #3783, rebased onto `dev`, so #3783 can be closed:

- `get_connection_error()` probes Stripe with the active environment's secret key (one Customer list, 401 = rejected credentials), cached per environment in a transient (1 hour healthy, 5 minutes on a 401 or transient failure). Cache clears on Stripe settings save, Connect return, and disconnect.
- Payment settings: the Status tag becomes **Connection Error** with an inline notice showing Stripe's message and a **Reconnect with Stripe** link (Connect) or a prompt to enter a new Restricted Key (API keys). The webhook section stops probing until the connection is restored.
- Admin notice on all PMPro admin pages for `manage_options` users with the same reconnect link. The misleading webhook notice is suppressed while the connection error is present.
- Site Health gateway line includes the connection error.
- Same-account guard: the Connect return handler refuses to overwrite an existing connection with a different Stripe account.
- Inline Connect URL building moved into `get_connect_url()`.

### How the two error states combine

- **Secret key rejected (with or without a publishable-key issue):** only the secret-key messaging above is shown. Reconnecting fixes the publishable key too, so the publishable-key row and notice are suppressed until the secret key works again.
- **Publishable key issue only:** the existing "Stripe checkout may be unavailable" row in the payment settings table, plus a new admin notice on all PMPro admin pages wired like the connection error notice. Both offer **Check for an updated Stripe key** or **Reconnect with Stripe**.

### Additional testing for the merged behavior

10. Connect in sandbox, then revoke the token from the Stripe dashboard (Settings → Authorized applications) or roll the Restricted Key. Within 5 minutes (or after saving Stripe settings) the Status tag, inline notice, and admin notice on other PMPro pages should all flip; the webhook notice should stop showing.
11. Click **Reconnect with Stripe** with the same account. Everything should clear immediately.
12. Reconnect with a *different* account. The connection should stay unchanged and the error should explain why.
13. With both a rejected secret key and a recorded publishable-key issue, only the secret-key messaging should appear. After reconnecting, both should clear.
14. With a working secret key and a recorded publishable-key issue, the publishable-key notice should appear on PMPro admin pages (not on the Stripe settings screen, where the inline row covers it).

### Changelog entry

BUG FIX: Refresh Stripe Connect platform publishable keys for on-site Stripe Elements checkout and billing updates.
ENHANCEMENT: Detect when Stripe no longer accepts the saved Stripe credentials and show the admin a notice with an option to reconnect. #3783 #3789 (@dparker1005)
BUG FIX: The Stripe webhook notice no longer reports a missing webhook when the real problem is that Stripe is rejecting the saved credentials. #3783 #3789 (@dparker1005)
BUG FIX: Connecting to Stripe now refuses to overwrite an existing connection with a different Stripe account. #3783 #3789 (@dparker1005)

### DIFF
--- a/classes/class-pmpro-site-health.php
+++ b/classes/class-pmpro-site-health.php
@@ -272,6 +272,11 @@ class PMPro_Site_Health {
 				$gateway_text .= ' (' . __( 'API Version', 'paid-memberships-pro' ) . ': ' . PMPRO_STRIPE_API_VERSION . ')';
 			}
 
+			$connection_error = $stripe->get_connection_error();
+			if ( ! empty( $connection_error ) ) {
+				$gateway_text .= ' (' . __( 'Connection Error', 'paid-memberships-pro' ) . ': ' . wp_strip_all_tags( $connection_error ) . ')';
+			}
+
 			if ( $legacy ) {
 				$gateway_text .= ' (' . __( 'Legacy Keys', 'paid-memberships-pro' ) . ')';
 				return $gateway_text . ' [' . $gateway . ':legacy-keys]';
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1689,9 +1689,6 @@ class PMProGateway_stripe extends PMProGateway {
 					<?php esc_html_e( 'This may be a temporary connection problem. Wait a few minutes, then try again. Onsite checkout and billing updates may remain unavailable in the meantime.', 'paid-memberships-pro' ); ?>
 				<?php } ?>
 			</p>
-			<p>
-				<a class="button button-secondary" href="<?php echo esc_url( self::get_publishable_key_refresh_url() ); ?>"><?php esc_html_e( 'Try checking again', 'paid-memberships-pro' ); ?></a>
-			</p>
 		</div>
 		<?php
 	}
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -131,6 +131,8 @@ class PMProGateway_stripe extends PMProGateway {
 		add_action( 'wp_ajax_pmpro_stripe_create_webhook', array( 'PMProGateway_stripe', 'wp_ajax_pmpro_stripe_create_webhook' ) );
 		add_action( 'wp_ajax_pmpro_stripe_delete_webhook', array( 'PMProGateway_stripe', 'wp_ajax_pmpro_stripe_delete_webhook' ) );
 		add_action( 'wp_ajax_pmpro_stripe_rebuild_webhook', array( 'PMProGateway_stripe', 'wp_ajax_pmpro_stripe_rebuild_webhook' ) );
+		add_action( 'wp_ajax_pmpro_stripe_refresh_publishable_key', array( 'PMProGateway_stripe', 'wp_ajax_pmpro_stripe_refresh_publishable_key' ) );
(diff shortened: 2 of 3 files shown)


ideadude:
Closed in favor of: https://github.com/strangerstudios/paid-memberships-pro/pull/3791

PR #3790, opened 2026-09-08
Update orders directly in the 3.8.4 transaction ID recovery instead of calling saveOrder()

dparker1005:
## Summary

The 3.8.4 upgrade's Stripe Checkout transaction ID recovery task called `MemberOrder::saveOrder()` for every affected order. `saveOrder()` rewrites the entire order row and fires `pmpro_update_order` / `pmpro_updated_order`, so every recovered order produced an "order updated" event. Sites with many affected orders hit Zapier zap limits and had other order-update integrations run for a change that only touched transaction IDs and a note.

This replaces both `saveOrder()` calls with direct `$wpdb->update()` writes:

- Failure path (Checkout Session could not be retrieved): writes `notes` only.
- Success path: writes `payment_transaction_id`, `subscription_transaction_id`, and `notes`.

`add_order_note()` is still used so note formatting is unchanged. Nothing else in the recovery depended on `saveOrder()` side effects; the subscription record is already created explicitly afterward.

Only sites that have not yet run (or are mid-way through) the recovery chain are affected, since processed orders are marked in order meta.

## Testing

- PHP lint passes.
- Verified by reading code; not run against a live Stripe account.

## Proposed changelog entry

- BUG FIX: The 3.8.4 upgrade no longer fires order update hooks (e.g. Zapier) for every order whose Stripe transaction IDs were recovered.

### DIFF
--- a/includes/updates/upgrade_3_8_4.php
+++ b/includes/updates/upgrade_3_8_4.php
@@ -138,8 +138,17 @@ function pmpro_stripe_recover_checkout_transaction_ids() {
 			$checkout_session = null;
 		}
 		if ( empty( $checkout_session ) ) {
+			// Write the note directly rather than calling saveOrder(), which would
+			// fire the pmpro_update_order/pmpro_updated_order hooks (Zapier, etc.)
+			// for a change that no integration needs to be notified about.
 			$order->add_order_note( __( 'Could not retrieve the Stripe Checkout Session for this order while trying to recover its missing transaction IDs.', 'paid-memberships-pro' ) );
-			$order->saveOrder();
+			$wpdb->update(
+				$wpdb->pmpro_membership_orders,
+				array( 'notes' => $order->notes ),
+				array( 'id' => $order->id ),
+				array( '%s' ),
+				array( '%d' )
+			);
 			update_pmpro_membership_order_meta( $order->id, 'stripe_checkout_transaction_id_recovery', 'failed' );
 			continue;
 		}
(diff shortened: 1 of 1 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3790 — Update orders directly in the 3.8.4 transaction ID recovery instead of calling saveOrder()**
*dparker1005 → dev | 1 file, +24 -2*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3790>

---

**Summary**
Ready to merge. Swapping `saveOrder()` for column-scoped `$wpdb->update()` calls is the right fix: `add_order_note()` (`classes/class.memberorder.php:1973`) only mutates `$this->notes` in memory, so the direct writes are necessary, and `saveOrder()`'s UPDATE branch (`classes/class.memberorder.php:1395`) has no side effects this task relies on beyond the hooks being removed. One pre-existing robustness gap is worth a follow-up but is not introduced by this PR.

---

**Issues**
- **[Minor]** `includes/updates/upgrade_3_8_4.php:210` — the success-path `$wpdb->update()` return is unchecked. On a DB error the order row keeps empty transaction IDs, the meta is still set to `recovered` (so the order is permanently excluded from the batch query's `tried.meta_id IS NULL` filter), and `PMPro_Subscription::create()` still runs from the in-memory `subscription_transaction_id`, leaving a subscription whose ID the order row never got. This is not a regression: the pre-PR `dev` code ignored `saveOrder()`'s return in exactly the same spot, so a failed write had the same outcome. But since you're now holding the raw result, a cheap guard is available: `if ( false === $wpdb->update(...) ) { update meta to 'db_error' (or leave it unset for retry); continue; }`. Same guard would fit the notes-only write at `:145`, though the stakes there are just a lost diagnostic note.

---

**Looks Good**
- Column list matches what actually changed: `notes` in the failure path; `payment_transaction_id`, `subscription_transaction_id`, `notes` in the success path. No full-row rewrite, so none of `saveOrder()`'s incidental normalization (`cancelled`→`success`, `checkout_id` backfill) can touch these rows.
- Format arrays (`%s` for the string columns, `%d` for the `id` WHERE) are correct, so the writes are prepared. All values come from the Stripe API or static translated strings passed through `add_order_note()`'s `wp_kses()`, so nothing unsanitized reaches `notes`.
- The downstream `PMPro_Subscription::create()` reads from the in-memory `$order`, not a DB read-back, so dropping `saveOrder()` doesn't change its inputs.
- Confirmed the recovery task shipped in `3.8.4` and `3.8.5` (`git grep` on the `3.8.5` tag finds `pmpro_stripe_recover_checkout_transaction_ids` in this file), so the PR description's scoping to "sites that have not yet run or are mid-way through the chain" is accurate.

dparker1005:
Thanks both. I'm going to leave the `$wpdb->update()` returns unchecked.

A single-row UPDATE by primary key only fails if the database is unreachable or the table is broken. In that state the `update_pmpro_membership_order_meta()` call on the very next line fails the same way, so the order is not marked and simply gets picked up again on the next run. The "silently skipped forever" outcome requires the orders write to fail while the ordermeta write succeeds a moment later, which isn't a realistic failure mode. The pre-PR code ignored the `saveOrder()` return in the same spots, so this isn't a regression, and I'd rather keep this diff scoped to the hook storm.

One behavior change worth noting for the record: `saveOrder()` also ran the billing-limit check after saving and could cancel a recovered subscription at Stripe if the limit was already reached. That no longer runs during recovery. The next renewal webhook performs the same check, so I'm comfortable with it.

PR #3791, opened 2026-09-08
BUG FIX: Refresh Stripe Connect platform publishable keys

flintfromthebasement:
### All Submissions:

Stripe Connect sites save the platform publishable key returned during OAuth forever. Rotating that platform key breaks onsite Stripe Elements checkout and billing updates even though the connected account and secret key are still valid.

This is the minimal fix, replacing the larger #3789:

- When the checkout or billing page renders for a Connect site, fetch the current platform keys from `https://connect.paidmembershipspro.com/stripe/v1/platform-keys.json` if the cache is missing or older than a week (`pmpro_stripe_connect_publishable_key_cache_lifetime` filter, manifest URL filterable via `pmpro_stripe_connect_publishable_key_manifest_url`).
- Cache both live and test keys in the `pmpro_stripe_connect_platform_keys` option. `get_publishablekey()` prefers the cached key for the current environment and falls back to the key saved during OAuth.
- When Stripe.js rejects the publishable key in the browser (`api_key_expired`, `platform_api_key_expired`, `invalid_api_key`, or the code-less "Invalid/Expired API Key provided" message), the checkout JS fires a nonce-protected AJAX request that refreshes the cache and shows "There was a problem connecting to the payment gateway. Please reload this page and try again." No automatic reload.
- Refreshes are throttled to one attempt per 5 minutes regardless of outcome. A failed or invalid manifest response leaves the existing cache untouched.
- Sites using their own API keys never fetch the manifest and never get the refresh nonce.

No settings UI, admin notices, or Site Health changes. Those can follow separately.

1. Configure Stripe Connect in sandbox mode with the onsite payment flow and load checkout. Confirm `pmpro_stripe_connect_platform_keys` is populated and `pmproStripe.publishableKey` matches the manifest test key.
2. Set `pmpro_sandbox_stripe_connect_publishablekey` to a bogus `pk_test_…` value and delete the `pmpro_stripe_connect_platform_keys` option. Load checkout: the manifest is fetched on render and the correct key is used, so checkout works.
3. Set the cached `test` key in `pmpro_stripe_connect_platform_keys` to a bogus `pk_test_…` value and delete the `pmpro_stripe_connect_platform_keys_checked` transient. Submit checkout: Stripe.js rejects the key, the "reload this page and try again" message shows, and the option is refreshed with the real key. Reload and checkout succeeds.
4. Repeat step 3 immediately: the throttle transient prevents a second manifest request within 5 minutes.
5. Configure a restricted key pair instead of Connect. Confirm no manifest request and no `publishableKeyRefreshNonce` in the localized script data.
6. Submit a normal card error such as `card_declined` and confirm it is displayed unchanged.

### Changelog entry

BUG FIX: Stripe Connect sites now refresh the platform publishable key from Paid Memberships Pro instead of using the key saved during the initial connection forever, so onsite checkout keeps working after the platform key is rotated.

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -1213,7 +1213,7 @@ class PMProGateway_stripe extends PMProGateway {
 	public static function wp_ajax_pmpro_stripe_refresh_publishable_key() {
 		check_ajax_referer( 'pmpro_stripe_refresh_publishable_key', 'nonce' );
 
-		if ( self::using_api_keys() ) {
+		if ( self::using_api_keys() || ! self::has_connect_credentials() ) {
 			wp_send_json_error();
 		}
 
@@ -1239,7 +1239,7 @@ class PMProGateway_stripe extends PMProGateway {
 				$stripe = new PMProGateway_stripe();
 				$localize_vars = array(
 					'publishableKey' => $stripe->get_publishablekey(),
-					'publishableKeyRefreshNonce' => self::using_api_keys() ? '' : wp_create_nonce( 'pmpro_stripe_refresh_publishable_key' ),
+					'publishableKeyRefreshNonce' => ( self::using_api_keys() || ! self::has_connect_credentials() ) ? '' : wp_create_nonce( 'pmpro_stripe_refresh_publishable_key' ),
 					'msgPublishableKeyRefreshed' => __( 'There was a problem connecting to the payment gateway. Please reload this page and try again.', 'paid-memberships-pro' ),
 					'user_id'        => $stripe->get_connect_user_id(),
 					'verifyAddress'  => apply_filters( 'pmpro_stripe_verify_address', get_option( 'pmpro_stripe_billingaddress' ) ),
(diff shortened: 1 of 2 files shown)


flintfromthebasement:
Ran the PR review council on this. Disposition:

**Applied in 1083300c1**
- AJAX handler now requires Connect credentials, matching the proactive path, and the refresh nonce is only issued when Connect is actually configured. Closes the gap where a Stripe site with no credentials could be made to fire outbound manifest requests.
- Publishable key validation is now a strict `^pk_(live|test)_[A-Za-z0-9]+$` match on both the manifest response and the cached value, instead of prefix-only.

**Intentionally not addressed**
- Blocking manifest fetch on checkout render during a manifest outage (one stall of up to 5s per site every 5 minutes). We are consciously letting things break when the service is down rather than adding backoff state. The 5-minute throttle already bounds load on the Connect server to one request per site per 5 minutes.
- Swapping the "reload and try again" message only when the refresh actually succeeded. Same reasoning: minimal scope, and the member can retry.

**Answer to the question:** the billing update page loads the same `pmpro-stripe.js` and calls the same `stripe.createPaymentMethod()` path, so the error object shape is identical to checkout.

PR #3792, opened 2026-09-09
ENHANCEMENT: Stripe connection test on the payment settings page

dparker1005:
### All Submissions:

Screenshot 2026-09-09 at 10 14 56 AM
Screenshot 2026-09-09 at 10 15 09 AM
Screenshot 2026-09-09 at 10 16 28 AM
Screenshot 2026-09-09 at 10 16 39 AM

Follow-up hardening to #3791, which shipped the minimal platform publishable-key refresh. This replaces the closed #3789 and absorbs the secret-key probe and same-account guard from #3783 (left open for now).

**Stripe Connection Test.** `PMProGateway_stripe::run_connection_test()` runs four checks and stores one result in the `pmpro_stripe_connection_test` option (not autoloaded, scoped to the active gateway environment):

| Check | How | Fails when |
|---|---|---|
| Stripe API | One `Customer::all( limit 1 )` request with the active secret key | No response or a 5xx |
| PMPro Connect Server | Fetches the platform-key manifest, bypassing the 5-minute throttle | Manifest unreachable or invalid (Connect sites only) |
| Secret Key | Same request as above | HTTP 401 only. A 403 from a restricted key counts as accepted. |
| Publishable Key | API-key sites: prefix/mode check. Then a read-only `PaymentIntent::retrieve` of a non-existent ID authenticated with the publishable key checkout uses, with the `Stripe-Account` header for Connect (valid key = 404, rejected = 401, revoked account = 403) | HTTP 401 only |

Each check is pass / fail / skipped with a message and, on failure, a suggested fix. Stripe calls use 5s connect / 10s total timeouts and restore the library defaults afterward.

**When it runs**
- Daily via the existing `pmpro_schedule_daily` action (only when Stripe is the primary gateway).
- Asynchronously on load of the Stripe settings page when there is no result from the last hour (AJAX, nonce + capability checked). The stored result renders immediately, a spinner shows while checking, and the Status cell is replaced with the server-rendered result.
- On demand via a **Run Connection Test** button in the same cell. The button's href is a nonce URL that runs the test and redirects, so it also works without JavaScript.

**Where it shows**
- The **Status** row of the active Stripe Connection section: the tag reads *Connection Error* when a key is rejected or *Connected, Issues Detected* when a server was unreachable, followed by "Last checked on…" and a *Show details* link. Details open automatically only when something failed.
- One admin notice on other PMPro admin pages, only for rejected keys (reachability problems never trigger it), linking back to the settings page. This reads the stored result and makes no outbound calls.
- The payment gateways list on the main Payment Settings page appends a *Connection Error* or *Issues Detected* tag after Stripe's "Enabled (Primary Gateway)" status. This goes through a new `pmpro_payment_settings_gateway_status_html` filter on the status HTML (passes the gateway slug), which Stripe hooks from its own class. Other gateways and add-ons can use the same filter.
- Site Health gateway line gets a `Connection Test: passed|<check:status list> on date` fragment.
- The existing webhook admin notice is unchanged except it now bails when the last test shows a rejected secret key, since the webhook check would fail for the wrong reason.

**Same-account guard (live mode only).** The Connect return handler refuses to overwrite an existing live connection with a different Stripe account and redirects immediately so the OAuth token never lingers in the URL. The error copy warns that disconnecting affects all sites on the original account. Sandboxes are exempt since Stripe issues a new account ID per sandbox. This makes the *Reconnect with Stripe* fix link safe. The inline connect/disconnect URL builders were consolidated into `get_connect_url()`.

**Deliberately not included:** email alerts, a failed-call log or HTTP decorator, browser-side auto-reload, and anything on the checkout request path. The webhook is not part of the test because it has its own section and notice.

**Notes for review**
- Every request the test makes is read-only. Nothing is created on the Stripe account. The settings page only auto-runs when there is no result from the last hour.
- Sites where Stripe is not the primary gateway are never tested.
- The off-settings-page notice currently lists each failed check. Kim's spec is a thinner "something is wrong, go look here" message; happy to trim before this leaves draft.
- Verified via wp-cli on dev.local: happy path, rejected secret key (bogus key and a genuinely revoked Connect token), rotated publishable key, Stripe and Connect server unreachable, environment switch, notice visibility rules, AJAX handler with a valid nonce. The browser flow (spinner, cell replacement, toggle) was verified manually by David.

1. Connect Stripe in sandbox mode and load the Stripe settings page. The Status row should show a spinner briefly, then *Connected*, "Last checked on…", and a *Show details* link. Details should list four passing checks.
2. Click **Run Connection Test**. The spinner should appear and the cell should refresh in place.
3. Replace `pmpro_sandbox_stripe_connect_secretkey` with a bogus `sk_test_…` value and reload. The tag should read *Connection Error*, details should open with Secret Key failed and a *Reconnect with Stripe* link, and other PMPro admin pages should show the connection notice while the webhook notice is suppressed. Restore the key.
4. Set the `test` key in the `pmpro_stripe_connect_platform_keys` option to a bogus `pk_test_…` value, filter `pmpro_stripe_connect_publishable_key_manifest_url` to an unreachable URL, and reload. Connect Server and Publishable Key should both fail.
5. Filter the manifest URL to an unreachable host and set `\Stripe\Stripe::$apiBase` to an unreachable host (a small mu-plugin works). The tag should read *Connected, Issues Detected*, Stripe API and Connect Server should fail, and the two key checks should show *Skipped*. Other PMPro admin pages should show no connection notice.
6. Switch the gateway environment. The Status row should show "not tested yet" until the next run.
7. With a site connected in sandbox, complete the Connect flow with a different Stripe account. The connection should stay unchanged and the error should explain why.
8. Use restricted API keys instead of Connect. There should be no Connect Server row, and a live-mode publishable key on a sandbox site should fail with a mode-mismatch message.

```
ENHANCEMENT: Added a Stripe connection test to the payment settings page that checks daily, on page load, and on demand whether Stripe and the Paid Memberships Pro Connect server can be reached and whether Stripe accepts the saved keys, with an admin notice when a key is rejected. #3783 #3789 #3791 (@dparker1005)
BUG FIX: Connecting to Stripe now refuses to overwrite an existing connection with a different Stripe account. #3783 (@dparker1005)
BUG FIX: The Stripe webhook notice no longer reports a missing webhook when the real problem is that Stripe is rejecting the saved credentials. #3783 (@dparker1005)
```

### DIFF
--- a/classes/gateways/class.pmprogateway_stripe.php
+++ b/classes/gateways/class.pmprogateway_stripe.php
@@ -14,6 +14,7 @@ use Stripe\ApplePayDomain as Stripe_ApplePayDomain;
 use Stripe\WebhookEndpoint as Stripe_Webhook;
 use Stripe\StripeClient as Stripe_Client; // Used for deleting webhook as of 2.4
 use Stripe\Account as Stripe_Account;
+use Stripe\Token as Stripe_Token;
 use Stripe\Checkout\Session as Stripe_Checkout_Session;
 
 define( "PMPRO_STRIPE_API_VERSION", "2025-09-30.clover" );
@@ -175,6 +176,12 @@ class PMProGateway_stripe extends PMProGateway {
 		add_action( 'admin_init', array( 'PMProGateway_stripe', 'stripe_connect_save_options' ) );
 		add_action( 'admin_notices', array( 'PMProGateway_stripe', 'stripe_connect_show_errors' ) );
 		add_action( 'admin_notices', array( 'PMProGateway_stripe', 'stripe_connect_deauthorize' ) );
+
+		// Connection test: run daily, run on demand from the payment settings page, and warn admins about failures.
+		add_action( 'pmpro_schedule_daily', array( 'PMProGateway_stripe', 'run_scheduled_connection_test' ) );
+		add_action( 'admin_init', array( 'PMProGateway_stripe', 'maybe_run_connection_test_on_demand' ) );
+		add_action( 'wp_ajax_pmpro_stripe_run_connection_test', array( 'PMProGateway_stripe', 'wp_ajax_pmpro_stripe_run_connection_test' ) );
+		add_action( 'admin_notices', array( 'PMProGateway_stripe', 'show_stripe_connection_notice' ) );
 
 		// Show warning if webhooks are not set up.
 		add_action( 'admin_notices', array( 'PMProGateway_stripe', 'show_stripe_webhook_setup_notice' ) );
@@ -1067,6 +1074,9 @@ class PMProGateway_stripe extends PMProGateway {
 				update_option( 'pmpro_' . $setting, sanitize_text_field( $_REQUEST[ $setting ] ) );
 			}
 		}
+
+		// The saved keys may have changed, so the last connection test no longer applies.
+		self::clear_connection_test_results();
 	}
 
 	/**
@@ -1592,6 +1602,9 @@ class PMProGateway_stripe extends PMProGateway {
 			|| ! isset( $_REQUEST['pmpro_stripe_access_token'] )
 		) {
 			$error = __( 'Invalid response from the Stripe Connect server.', 'paid-memberships-pro' );
+		} elseif ( self::is_different_connected_account( $_REQUEST['pmpro_stripe_connected_environment'], $_REQUEST['pmpro_stripe_user_id'] ) ) { // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			// Reconnecting with a different account would orphan every existing customer and subscription.
+			$error = __( 'The Stripe account you just connected is not the account this site was already connected to, so the connection was left unchanged. Existing memberships are tied to the original account. To switch Stripe accounts, disconnect from Stripe first. Note that disconnecting will disconnect all sites using the original Stripe account.', 'paid-memberships-pro' );
 		} else {
 			// Update keys.
 			if ( $_REQUEST['pmpro_stripe_connected_environment'] === 'live' ) {
@@ -1610,6 +1623,9 @@ class PMProGateway_stripe extends PMProGateway {
 			// Delete option for user API key.
 			delete_option( 'pmpro_stripe_secretkey' );
 			delete_option( 'pmpro_stripe_publishablekey' );
+
+			// The credentials changed, so the last connection test no longer applies.
+			self::clear_connection_test_results();
 
 			unset( $_GET['pmpro_stripe_connected'] );
 			unset( $_GET['pmpro_stripe_connected_environment'] );
@@ -1700,6 +1716,9 @@ class PMProGateway_stripe extends PMProGateway {
 			delete_option( 'pmpro_sandbox_stripe_connect_secretkey' );
 			delete_option( 'pmpro_sandbox_stripe_connect_publishablekey' );
 		}
+
+		// The credentials changed, so the last connection test no longer applies.
+		self::clear_connection_test_results();
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3792 — ENHANCEMENT: Stripe connection test on the payment settings page**
*dparker1005 → dev | 2 files, +607 -37*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3792>

---

**Summary**
Well-built feature. The HTTP-status classification (401-only means a rejected key, so restricted keys that 403 on `Customer::all` pass), the environment-scoped result storage, and the nonce + capability gating on all three trigger paths are all right. Two correctness gaps should be fixed before merge: the same-account guard doesn't actually leave the connection "unchanged," and the test misreports when the Stripe library isn't loaded. Rest is cleanup.

---

**Issues**

- **Major** `classes/gateways/class.pmprogateway_stripe.php:1589-1590` vs `:1605` — The guard's error copy says "the connection was left unchanged," but `stripe_connect_save_options()` already ran `update_option( 'pmpro_gateway', 'stripe' )` and `update_option( 'pmpro_gateway_environment', ... )` at the top of the function before `is_different_connected_account()` is evaluated. A sandbox site whose admin completes live OAuth with the wrong account gets its environment flipped to live while the keys stay as they were. The ordering is pre-existing (the other error branches share it), but this PR is the one advertising the guard as a safe reconnect path. Move the two `update_option()` calls below the error checks, or restore the previous values in the new `elseif`.

- **Minor** `classes/gateways/class.pmprogateway_stripe.php:2082-2085` and `:2151-2161` — When `self::$is_loaded` is false (the constructor only sets it after `dependencies()` passes, `:49-63`) but a secret key is saved, `secret_key` is recorded as `fail` with "No Stripe credentials are saved for this environment," which is untrue. Because `stripe_api` is `unknown` rather than `fail`, the publishable-key branch also falls through to `Stripe_Token::create()` on a class that was never loaded, and the caught Error surfaces as "The publishable key could not be checked: Class Stripe\Token not found." Since `secret_key` is `fail`, this also lights up the "Connection Error" notice across PMPro admin pages. Branch on `! self::$is_loaded` separately, skip both key checks, and reuse the missing-extension messaging `dependencies()` already produces.

- **Minor** `classes/gateways/class.pmprogateway_stripe.php:2126-2131` — `refresh_connect_publishable_keys()` returns `false` both on a fetch/validation failure and when the throttle transient is already set (`:1997-1999`, read at head). The test deletes the transient first, so this only misfires when two runs overlap (daily cron plus a settings-page AJAX, or two admin tabs), but in that case the loser records "Connect server could not be reached." Consider a `$force` parameter that bypasses the transient check, or return a distinct value for throttled.

- **Minor** `classes/gateways/class.pmprogateway_stripe.php:3446-3450` — `$connection_selector` is still computed in `show_connection_settings_section()` but is no longer read there; `show_connection_status_cell()` computes its own. Delete the block.

- **Minor** `classes/gateways/class.pmprogateway_stripe.php:2090-2092`, `:2162-2164` — `catch ( \Exception )` after `catch ( \Throwable )` is unreachable. Copilot flagged this and it's correct; drop the second catch in both spots.

- **Minor** `classes/class-pmpro-site-health.php:275-277` — `get_connection_test_results()` and `get_failed_connection_tests()` are static but called on the instance. Consistent with the surrounding file, but since these lines are new, call them statically.

---

**Looks Good**
- Timeout save/restore around the Stripe client has no early returns between set and restore, and every call site is wrapped in a catch.
- `get_connect_url()` is a faithful consolidation: same args, same nonce actions, same `test` mapping for the Connect server.
- Output escaping is consistent: stored Stripe messages go through `esc_html()` at every render site, fix links through a tight `wp_kses()` allowlist, and JS strings through `wp_json_encode()`.
- `get_connection_test_results()` returning false on an environment mismatch is a clean way to handle environment switches without extra clearing.

---

**Questions**
1. The auto-run creates a live-mode PII token on the connected account for any user with `pmpro_paymentsettings`. Should the on-load auto-run be limited to `manage_options`, or is that capability tier acceptable for a write on the merchant's account?
2. Is the thinner off-settings-page notice from Kim's spec landing here before the draft closes, or as a follow-up?

dparker1005:
Thanks, this was a useful pass. Fixed in f69ecea5c: the gateway/environment updates in `stripe_connect_save_options()` now happen only in the success branch, so a failed or refused connection leaves settings untouched; the unloaded-library case stores every check as Skipped and returns early; `refresh_connect_publishable_keys()` takes a `$force` flag instead of the test deleting the throttle transient; the dead `$connection_selector` is gone; and the Site Health calls are static. Leaving the `\Throwable` / `\Exception` pair as-is, that's the pattern used throughout this class.

On Q1, `pmpro_paymentsettings` can already replace the keys outright, so a throwaway PII token under that capability seems proportionate. On Q2, deciding that before this leaves draft.

andrewlimaza (approved):

kimcoleman (approved):

kimcoleman:
Added a pmpro_tag-small selector and icons to tighten this up a bit + adjusted wording slightly for one failure. This is great!

Screenshot 2026-09-18 at 11 11 59 AM

Note that we can use this same pmpro_tag-small I think on the page settings / templates area. There's an override there that would be fine using the pmpro_tag-small. If I see this PR merge to 3.9 I'll do that as well.

classes/gateways/class.pmprogateway_stripe.php
```diff
@@ -1979,6 +2083,406 @@ public static function maybe_refresh_connect_publishable_keys() {
...
+	 * Only enforced for live mode. Sandboxes get a new account ID whenever one is created, so switching them is routine.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $gateway_environment The environment being connected, 'live' or 'sandbox'.
+	 * @param string $stripe_user_id      The Stripe account ID returned by the Connect server.
+	 * @return bool True if a different account is already saved for this environment.
+	 */
+	private static function is_different_connected_account( $gateway_environment, $stripe_user_id ) {
+		$saved_user_id = get_option( 'pmpro_' . ( 'live' === $gateway_environment ? 'live' : 'sandbox' ) . '_stripe_connect_user_id' );
```
flintfromthebasement: The guard keys on `pmpro_live_stripe_connect_user_id` alone, while `has_connect_credentials()` defines "connected" as user id + secret key + publishable key. A site that has a stale user id but no keys shows **Not Connected** in the Status row, yet a live Connect to any other account is refused with the "already connected to a different account" error. Nothing on the page explains why, since the page itself says the site isn't connected.

Our own paths don't produce that state: `stripe_connect_deauthorize()` deletes all three options, and the hosting staging scrub does too. Where it does show up is manual "disconnects" (someone clears the secret key by hand), customer-side migrations that copy `wp_options` and then strip only what looks like a secret, and dev sites with a placeholder user id (pmprodev.test has `acct_testonly` set and hits this today).

Suggest only enforcing when the environment is actually connected:

```php
$saved_user_id = get_option( 'pmpro_' . $env . '_stripe_connect_user_id' );
return self::has_connect_credentials( $env ) && ! empty( $saved_user_id ) && $saved_user_id !== $stripe_user_id;
```

That keeps the protection for real live connections (the case the guard exists for) and lets a half-cleared site connect normally. Edge case rather than a blocker, but the failure is silent and confusing, and the fix is one condition.

Issue #2293, opened 2022-12-01
User Fields First Radio Buttons show "required" asterisk *, even when not required. 


kimwhite:
**Describe the bug**
When using User Fileds and RH code, the First Radion field created always shows a "required" asterisk. The second Radio Button will not show an asterisk when required.  However, they do work as they should. 

**To Reproduce**

1. Set up 2 User Fields Radio Buttons.
2. Do not require 1st (no)
3. Require 2nd (yes)
4. See how it displays

**Screenshots**
<img width="942" alt="Screenshot 2022-12-01 at 12 53 31 PM" src="https://user-images.githubusercontent.com/8083630/205125433-e9d5ed4b-977f-4e82-ad0b-2b5a9ad17848.png">

<img width="548" alt="Screenshot 2022-12-01 at 12 49 21 PM" src="https://user-images.githubusercontent.com/8083630/205124561-98141df0-14f1-4bd1-bdaa-81876527761f.png">

**Expected behavior**
The asterisk should show only when the field is required.

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only Paid Memberships Pro plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Memberlite](https://wordpress.org/themes/memberlite/).
- [x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
<details>
```
WordPress 6.1.1
PMPro 2.9.7
```
</details>

PR #3795, opened 2026-09-10
Show required asterisk once for radio user fields

### LINKED ISSUE #3794: Required radio user fields show asterisk on every option
**Describe the bug**
When a User Field of type Radio is required, the required asterisk appears on every option label (Yes, No, etc.) instead of only on the main field label.

**To Reproduce**
1. Create a Radio user field
2. Mark it "required"
3. View on checkout page
4. Each radio option shows its own asterisk.

**Screenshots**

Image

**Expected behavior**
Only the field label should have an asterick

- [ x] I have deactivated other plugins and confirmed this bug occurs when only Paid Memberships Pro plugin is active.
- [ x] This bug happens with a default WordPress theme active, or [Memberlite](https://memberlitetheme.com).
- [x ] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
WP 7.1
PMPro 3.8.6
</details>

faisalahammad:
### All Submissions:

`js/pmpro-checkout.js` adds the required asterisk to labels from two blocks. The second block walks up from each required input with `closest('.pmpro_form_field')` to find the label to mark. Radio options are each wrapped in their own `<div class="pmpro_form_field pmpro_form_field-radio-item">` (`class-pmpro-field.php`, radio branch), so the lookup stopped at the option wrapper and every option label (Yes, No, etc.) got its own asterisk instead of only the field label.

The fix resolves the wrapper through the radio group container first, with a fallback to the old lookup so every other field type is untouched. The first block already adds one asterisk to the field label, so its existing check prevents a duplicate and the result is a single asterisk on the label only.

`aria-required` is still set on all radio inputs, so screen reader behavior is unchanged.

Resolves #3794.

1. Go to Memberships > Settings > User Fields and create a Radio field with a few options (e.g. Red, Blue, Green) and set Required at Checkout to Yes.
2. View the checkout page.
3. The field label shows one asterisk and the individual radio options show none.
4. Repeat with a required Text field and a required Checkbox Grouped field to confirm they still show one asterisk on the label only.
5. Submit checkout with the radio field empty to confirm required validation still works.

### Other information:

Checked with a headless DOM test against the markup the field classes generate: before the change a required radio field rendered 4 asterisks (label plus all 3 options), after the change exactly 1 on the field label. Text, checkbox_grouped and the Terms of Service checkbox rendered identically before and after. Note for context: the per-option asterisk behavior was introduced with #2499 (for #2293), which was about showing the asterisk only when required. This keeps that part and fixes the placement.

### Changelog entry

> Bug fix: Required radio user fields now show the required asterisk on the field label only, instead of on every option label.

### DIFF
--- a/js/pmpro-checkout.js
+++ b/js/pmpro-checkout.js
@@ -156,8 +156,14 @@ jQuery(document).ready(function(){
 	});
 
 	jQuery('.pmpro_form_input-required').each(function() {
+		// Radio options are each wrapped in their own .pmpro_form_field div, so look up the
+		// label from the radio group container to avoid marking every option as required.
+		var $fieldDiv = jQuery(this).closest('.pmpro_form_field-radio-items').parent();
+		if ( ! $fieldDiv.length ) {
+			$fieldDiv = jQuery(this).closest('.pmpro_form_field');
+		}
+
 		// Check if there's an asterisk already
-		var $fieldDiv = jQuery(this).closest('.pmpro_form_field');
 		var $firstLabel = $fieldDiv.find('.pmpro_form_label').first();
 		var $hasAsterisk = $firstLabel.find('.pmpro_asterisk').length > 0;
 


dparker1005 (approved):
Reproduced the issue and verified the fix on a local site (4 asterisks before, 1 on the group label after; aria-required unchanged). Approving.

PR #3798, opened 2026-09-14
Extend the getfile extension blocklist with additional PHP file extensions

dparker1005:
## Summary

Adds `php7`, `php8`, `php9`, `phtm`, `pht`, and `phar` to the default extension blocklist in `services/getfile.php`. Files with these extensions are now redirected back to the web server (with the existing `noloop` guard) instead of being read and streamed by the script, matching how `php`, `phtml`, etc. are already handled.

This is a minor hardening change to the legacy, opt-in getfile script (`PMPRO_GETFILE_ENABLED`). It is not a security release. The default filterable list simply did not include several PHP-family extensions.

Sites that need a different list can continue to use the `pmpro_getfile_extension_blocklist` filter.

## Credit

The missing extensions were pointed out by security researcher **kta1kri**, who asked to be credited by that handle.

## Testing

- With `PMPRO_GETFILE_ENABLED` and the getfile rewrite active, request a `.php7`/`.phar` file under uploads through getfile. Expected: redirect with `noloop=1` rather than the file contents being streamed.
- Request a normal protected/unprotected file. Expected: unchanged behavior.

## Proposed changelog entries

```
* ENHANCEMENT: Added additional PHP file extensions to the default extension blocklist for the getfile script. #3798 (@dparker1005, @kta1kri)
```


flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3798 — Extend the getfile extension blocklist with additional PHP file extensions**
*dparker1005 → dev | 1 file, +1 -1*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3798>

---

**Summary**
Correct, minimal, one-line hardening of the opt-in getfile script. Ready to merge as-is.

---

**Looks Good**
- `services/getfile.php:116` — adds `php7`, `php8`, `phtm`, `pht`, `phar` to the single enforcement point. `$ext` is lowercased at line 113 before the `in_array()` check at line 120, so the new lowercase entries match case variants (`.PHT`, `.Phar`) just like the existing ones.
- The redirect path (lines 120-131: `noloop` param, `wp_safe_redirect()`) is untouched, so blocked extensions fall back to the web server's own handling rather than opening any new code path.
- `pmpro_getfile_extension_blocklist` (line 117) still lets hosts override the default, so no back-compat risk for sites that already customized the list.
- Confirmed via grep this is the only hardcoded copy of the blocklist in the repo. Nothing else needs the same edit.
- The short list ships today: `git grep` on `v2.5.9.1` shows the old array at `services/getfile.php:108`, so existing `PMPRO_GETFILE_ENABLED` sites pick up the hardening on upgrade.

---

**Questions**
1. Pre-existing, not introduced here: extension extraction (`explode('.')` + last segment, line 113) only inspects the final suffix. A file like `shell.pht.jpg` resolves to `jpg` and streams via `readfile()`. On a normally configured server that's fine (the server won't execute it either), but on Apache `AddHandler`-style setups multi-extension parsing can execute it. Worth confirming with kta1kri whether double extensions were in scope or a separate follow-up.
2. Also pre-existing: the deny-list is the only disclosure control for non-attachment files, so `.htaccess` and `.user.ini` under uploads remain streamable. Same one-line fix pattern if you want to fold `htaccess`/`ini` into this PR; otherwise fine as a separate change.

andrewlimaza (approved):
This looks good to me.

PR #3799, opened 2026-09-14
Add a dedicated URL / Embed user field type

### LINKED ISSUE #3797: Add a dedicated URL/Embed User Field type
User Field values are sanitized on save, which strips embed codes and some URL-bearing markup. Site owners who want members to share a video, a social post, or other embeddable content currently have no supported way to do it. The current workaround is using a custom snippet that disables sanitization for a specific field.

A dedicated "URL / Embed" User Field type:
- The field accepts a URL rather than a full embed code.
- On display, if the URL matches a supported oEmbed provider, convert it to an embed automatically.
- If it isn't embeddable, fall back to displaying it as a plain link.
- Keep it easy to configure either way, so a site can choose to show the raw URL instead of an embed if that's what they want.

**Describe alternatives you've considered**
A `wp_kses`-based allowed-HTML sanitization function instead of the current sanitization, permitting a restricted set of tags. 

**Additional context**
Mods: 720632

faisalahammad:
### All Submissions:

Resolves #3797.

There is no user field type built for links today. The closest option is a text field, but the value is run through `sanitize_text_field()` on save, which strips embed codes and can mangle query strings. Members end up pasting a full `<iframe>` into a text field and it never renders.

This adds a `url` field type, shown in the admin as "URL / Embed". It accepts a plain URL and takes care of the display:

- Saves the value through `esc_url_raw()` so the query string and other link characters survive.
- On display, converts the URL to an oEmbed when the provider supports it and falls back to a plain link when it does not. This reuses the URL handling that text and textarea fields already had, so there is no new embed logic.
- Adds a per-field "Display As" setting under the field type, with options for Embed (if supported), Link (show the URL), Link (show the field label), and Plain text. The value feeds into the existing `pmpro_field_value_link_display_type` filter as the default, so any site already using that filter keeps overriding it.
- Invalid submissions are blocked with an error instead of being saved. Only http and https URLs are accepted, so `javascript:` and other schemes are rejected. Checked at checkout, on the frontend profile, and when an admin edits a member. The previously stored value is left alone when a bad value comes in.
- Emails always show a plain link, since mail clients cannot render an iframe.
- Text and textarea fields keep their current behavior. They can now also use the Display As setting, but the default is unchanged.

1. Go to Memberships > User Fields, add a new field, set the type to "URL / Embed". A "Display As" dropdown appears below the type. Save and reload the page to confirm the type and dropdown both persist.
2. Place a test checkout, enter a valid link such as a YouTube URL in the field, and complete checkout. On the confirmation page and in the confirmation email the value should appear as a link, not an iframe (email never embeds).
3. As a member, open the profile page and view the field. A YouTube or Vimeo link should render as an embed. A link to a page that does not support oEmbed should fall back to a clickable link.
4. Submit an invalid value (for example `not a url`, or `javascript:alert(1)`) at checkout and on the profile page. Each should stop with "The <label> field must be a valid URL." and the old value should stay in place.
5. Change the field's Display As setting to each option and confirm the frontend display changes accordingly.
6. Check that a text and a textarea field still render and save the way they did before this change.

Expected result: the URL field saves valid links, blocks invalid ones with a clear message, embeds where the provider allows it, falls back to a link otherwise, and never puts an iframe in an email.

### Other information:

### Changelog entry

> FEATURE: Added a new URL / Embed user field type that saves a link, embeds it when the provider supports oEmbed, and falls back to a plain link otherwise. Each field can choose how the value is displayed, and invalid submissions are rejected instead of saved.


dparker1005:
Hi @faisalahammad, thank you for the work on #3797. The implementation is careful and the description is thorough.

We're going to close this one, though. Digging into the original report, the behavior it asks for already exists: a URL to a supported oEmbed provider in a text or textarea user field is converted to an embed on display in the member profile shortcode and the Member Directory, with a link fallback, and `pmpro_field_value_link_display_type` switches that per site. The report itself traces to 3.8.2, where we restored user field sanitization that had been inactive since 3.4. Sites that had been pasting full `<iframe>` embed codes into text fields were relying on that gap, and the fix is to enter the URL instead. That's a documentation and support answer rather than a new field type. Details on #3797.

Thanks again for the effort here.

PR #3800, opened 2026-09-15
feat(tos): require TOS re-acceptance after page update

### LINKED ISSUE #3633: Feature Request: Force members to re-accept TOS after Terms of Service page update
## Feature Request

When a site admin updates their Terms of Service page, there is currently no built-in way to require existing members to re-accept the updated TOS without going through full checkout again. This is a problem for sites with paid members — forcing a re-checkout is not an acceptable UX.

## Current behavior

PMPro stores consent log entries in user meta (via `pmpro_consent_log`) that include the `post_modified` timestamp of the TOS page at time of consent. `pmpro_is_consent_current()` already detects when a TOS page has been updated. However, there is no mechanism to surface this to members or prompt them to re-accept.

## Proposed behavior

1. **Admin trigger** — When editing the TOS page, admins should see a checkbox: "Require all members to re-accept this Terms of Service." Saving with this checked invalidates all prior consent entries.
2. **Member prompt** — Members with stale or missing consent should see a modal or interstitial (not a checkout redirect) prompting them to re-accept the TOS. This should work without requiring a new payment transaction.
3. **Consent logging** — The acceptance date/timestamp should be stored, consistent with how PMPro already logs TOS consent at checkout.

## Notes

- WordPress core added a privacy policy update flow but never shipped the "force re-check" piece. PMPro should roll its own.
- The Force Profile Fields add-on was considered as a workaround but uses a different mechanism and doesn't fit cleanly.
- For **free levels**, a workaround exists: redirect active members with stale consent to the checkout page. Snippet available.
- For **paid levels**, the redirect-to-checkout workaround is not acceptable.
- It would be valuable to get input from affected customers on their ideal UX walkthrough before building.

## Workaround (free levels only)

Redirect active members with stale/missing TOS consent to the checkout page with their current level pre-filled. Could be available as a code snippet.

faisalahammad:
### All Submissions:

Resolves #3633.

The consent API already stores the page modified time with each consent (`pmpro_save_consent()`), and `pmpro_is_consent_current()` can already tell whether a stored consent is out of date. What was missing was a way to surface stale consent to the member. This adds an opt-in prompt for that.

The feature is off by default. A new meta box on the edit screen of the page set as the Terms of Service page has a checkbox to require re-acceptance. When checked, logged in members whose consent does not match the current version of the page see a notice with an agree button. Accepting logs a new consent entry through the existing API, so it appears in the admin order view and member edit screen like any other TOS consent.

The setting stores the page ID it applies to, so it self-invalidates if the admin changes the Terms of Service page. Nothing changes on upgrade until an admin opts in.

Files changed:

- `includes/terms-of-service.php`: meta box and save handler, `pmpro_is_tos_reaccept_required()`, `pmpro_user_needs_tos_reaccept()`, the notice, and the accept handler. All in the existing TOS file, which already holds the consent API.
- `css/frontend/base.css` and `css/frontend/base-rtl.css`: notice styles, reusing the `alert2` tokens.

No JavaScript, so no build step and no new assets to enqueue (`base.css` already loads on every front end page). The notice renders on `wp_body_open` with a `wp_footer` fallback for themes that do not call `wp_body_open()`.

A double submit is handled: the accept handler checks `pmpro_user_needs_tos_reaccept()` before saving, so a double click or back then resubmit does not add duplicate consent entries, and it redirects after saving (POST/redirect/GET).

1. Set a Terms of Service page under Memberships > Settings > Advanced, and have a test member with consent on file.
2. Edit that page. A "Terms of Service" meta box appears in the sidebar. Check "Require all members to re-accept this Terms of Service", make a small content edit so the modified time changes, and update the page.
3. Log in as the test member in a private window. The notice appears at the top of front end pages with a link to the page and an "I agree" button.
4. Click "I agree". The page reloads, the notice is gone, and a new consent entry for the current version is recorded on the member edit screen and the order screen if the member has an order. Expected result: the notice stays gone after a refresh and only one new consent entry exists.
5. Uncheck the box on the TOS page and update. The notice stops showing, and existing consent entries are kept.

Edge cases worth checking: change the Terms of Service page under Advanced Settings and confirm the prompt stops (the setting only applies to the page it was set on); Quick Edit or bulk edit on the TOS page does not change the setting; the button works with JavaScript disabled; a theme without `wp_body_open()` still shows the notice; the meta box saves correctly in the block editor.

### Other information:

### Changelog entry

> Added an opt-in setting on the Terms of Service page edit screen to require members to re-accept the current version of the page. Members with out of date consent see a notice with an agree button, and accepting is logged in the existing consent history.


dparker1005:
Hi @faisalahammad, thank you for taking on #3633. The code here is solid: reusing the existing consent API, storing the page ID so switching TOS pages turns it off, the sentinel field so Quick Edit and REST saves don't silently uncheck the box, and the double-submit guard on the accept handler are all the right instincts.

I'm going to close this one, though, because reviewing it surfaced design questions we haven't answered yet, and it wouldn't be fair to ask you to keep iterating while we work those out. The big ones: whether a passive banner is the right mechanism at all versus a login interstitial or a content gate, since those produce very different guarantees about whether members actually re-accepted; who counts as a "member" for this purpose; and whether the setting belongs on the TOS page edit screen or with the other TOS settings under Advanced Settings. Those need to be settled on our side before an implementation can be reviewed against them.

I've written up the open questions on #3633. Thanks again for the work here.

Issue #1037, opened 2019-08-04
Missing translations


AiratHalitov:
Missing translations on order detail page. 
`test.site/wp-admin/admin.php?page=pmpro-orders&order=1`
Lines:

- Status
- Payment Transaction ID
- Subscription Transaction ID

**Screenshot**
![pmpro-tr](https://user-images.githubusercontent.com/4050715/62426928-ba02ab00-b704-11e9-92a7-678348ad166b.png)

Paid Memberships Pro  2.0.7

andrewlimaza:
Thank you for creating this issue.

The order status will require a bit of work to get this translation ready as it's the same as what's stored inside the database and this will cause issues if you translate it.

The Payment Transaction ID and Subscription Transaction ID are translatable, please see the code reference: https://github.com/strangerstudios/paid-memberships-pro/blob/dev/adminpages/orders.php#L793 and https://github.com/strangerstudios/paid-memberships-pro/blob/dev/adminpages/orders.php#L810

AiratHalitov:
Ok, thanks!

PR #3801, opened 2026-09-15
Translatable order/membership/subscription status labels

### LINKED ISSUE #3550: Order status labels can't translate
**Describe the bug**
Membership and order status labels (e.g. pending, success) are stored as raw values in the database and output directly in user-facing messages. Because these values are not passed through translation functions or filters, they cannot be translated.

This affects multilingual sites and any site that needs localized user-facing copy.

- Status values such as pending are stored in the database as internal identifiers.
- These values are output directly in user-facing strings, for example on the confirmation page:
- Your membership status is: <b>pending</b>.

Since the displayed value comes directly from the database, it bypasses localization and cannot be translated via .po/.mo files or translation plugins.

**Screenshots**
Confirmation Page when using Pay-by-Check Add On
Image

**Expected behavior**
The ability to translate status with a PO/MO file or Loco Translate, etc.

faisalahammad:
### All Submissions:

Resolves #3550. Also relates to #1037.

Order, membership, and subscription status values (e.g. `pending`, `success`) are stored as raw identifiers and were echoed straight into user-facing and admin output, so PO/MO files and tools like Loco Translate could not translate them.

This PR adds three label helper functions to `includes/functions.php` and routes every core status output site through them:

- `pmpro_get_order_status_label( $status )` — maps `''`/`success`/`cancelled` to "Paid", plus refunded, pending, token, review, error. Filterable via `pmpro_order_status_labels`. Unknown statuses (e.g. from Add Ons) fall back to a humanized slug.
- `pmpro_get_membership_status_label( $status )` — active, inactive, expired, cancelled, admin_cancelled ("Cancelled by Admin"), changed, admin_changed ("Changed by Admin"). Filterable via `pmpro_membership_status_labels`.
- `pmpro_get_subscription_status_label( $status )` — active, cancelled, sync_error. Filterable via `pmpro_subscription_status_labels`.

Same pattern as the existing `pmpro_get_site_types()` map, so the labels now land in the .pot file.

Call sites converted (label text only, `value` attributes and CSS classes keep the raw slug so filtering and styling still work): invoice page, account shortcode, profile (order history, subscriptions, level history), orders list table (column + filter dropdown + SQL sort CASE), subscriptions list table, view/edit order admin pages, orders print template, recent orders metabox, member edit panels (orders + memberships), view subscription admin page.

Two fixes that come out of this pass:

- The account shortcode left `$display_status` and `$tag_style` undefined for the `token`, `review`, and `error` statuses.
- The subscriptions list table labelled anything not `active` as "Cancelled", so a `sync_error` subscription displayed as Cancelled.

Filter dropdown `value=""` attributes are unchanged, so admin order filtering behaves exactly as before.

1. Switch the site language to a non-English one (or edit the .po file with Loco Translate) and translate the strings "Paid", "Pending", "Refunded", "Token", "Review", "Error".
2. Visit the frontend pages: Membership Invoice, Membership Account (order history section), and the WP profile order history.
3. Check the admin: Orders list table (status column + filter dropdown), a single order (view + edit), Orders print view, the member edit Orders and Memberships panels, and the Subscriptions list.
4. Statuses should now display translated. With no translation loaded, output is identical to before for all standard statuses.

### Other information:

No existing helper covered this. Each renderer had its own partial if/elseif chain covering only 2-3 statuses, with a `ucwords()` fallback. Standard status output is byte-identical to before. The new functions and filters are additive, no BC break.

Note: the exact string from the issue screenshot ("Your membership status is: pending") comes from the Pay by Check Add On, which interpolates the status into an otherwise translatable sentence. Core cannot fix an Add On directly, but with this helper shipped the Add Ons can pass their status through it. I can open a follow-up PR on the Add Ons side.

### Changelog entry

> Add translatable label functions for order, membership, and subscription statuses and use them for all status output.


dparker1005:
Hi @faisalahammad, thank you for taking on #3550. The call-site sweep is thorough and centralizing labels the way `pmpro_get_site_types()` does is a reasonable shape.

We're going to close this one, though. Looking at the issue again, the string in Kim's report comes from the Pay by Check Add On's confirmation message, which the description acknowledges core can't fix, and the member-facing surfaces in core, the account and invoice pages, already translate their status labels. What's left is admin-only output, and for that the trade isn't right: three new public functions and three filters that Add Ons will start depending on, plus two issues in the branch as it stands. The edit order status select and the orders list filter now show duplicate "Paid" options because `''`, `success`, and `cancelled` all map to the same label, and `sync_error` is a subscription meta key rather than a status, so the "fix" for it and the docblock entry describe something that doesn't exist.

The right fix for #3550 is in the Pay by Check Add On, where the status is interpolated into the sentence, and we'll handle it there. Thanks again for the work here.

Issue #3626, opened 2026-03-27
Email Settings page overhaul — hosting awareness, SendWP cleanup, deliverability UX


flintfromthebasement:
## Overview

The PMPro Email Settings page needs an overhaul to reflect PMPro Hosting as the primary transactional email solution, reduce SendWP prominence, and surface deliverability issues more proactively.

---

## 1. Hide SendWP section for PMPro Hosting customers

Hosting customers already have transactional email via `pmpromailer.com`. Hide the SendWP block for them and replace it with a status note confirming email sending is active. For non-hosting customers experiencing deliverability issues, pitch PMPro Hosting (Max plan includes email sending) instead of SendWP.

## 2. Remove/de-emphasize SendWP integration

Low adoption. The real estate is better used to highlight PMPro Hosting. Longer term, consider dropping the integration entirely.

## 3. Remove "Default WP notification email" setting

Low value, rarely understood by customers. Should default to unchecked. Consider removing entirely — a gist can cover the edge case for anyone who needs it.

## 4. Detect email failures → surface prominent alert

If the email log shows a high failure rate, trigger a visible CTA on the settings/log page: "Your emails are failing — here's how to fix it" → link to troubleshooting guide or PMPro Hosting upsell.

## 5. Expose configurable sender prefix for pmpromailer.com

Customers who are sensitive about the visible from address (`[sitename]@pmpromailer.com`) should have a setting to customize the prefix. **Constraint:** full domain override (e.g. `@underdogs.win`) is not possible on the SES side — only the prefix is configurable.

## 6. Docs/support note: reply-to is preserved

When PMPro Hosting transactional email is active, the from address changes to `[sitename]@pmpromailer.com`, but the reply-to stays as configured in WP/PMPro settings. This should be documented clearly and included in support notes for Donna and the team.

---

## Support flow (for reference)

1. Deactivate mailer plugins (Offload SES, WP Mail SMTP, etc.)
2. PMPro Hosting transactional email picks up automatically — no DNS changes needed
3. Heads up to customer: from address will change; reply-to stays as configured

PR #3804, opened 2026-09-15
fix(admin): remove default WP notification email setting

faisalahammad:
### All Submissions:

Part of #3626 (point 3: remove the "Default WP notification email" setting).

This removes the `email_member_notification` setting from Memberships > Email Settings. The checkbox was the only field in the "Other Email Settings" section, so that section is gone too, and the option is no longer saved, read, or sent to Wisdom.

The setting was confusing. Members already get a confirmation email from PMPro after checkout, so checking the box just sent a second, stock WordPress new user email. The recommended state was always unchecked, and the default is unchecked, so almost every site already had it off. To keep behavior identical for those sites, WordPress's new user notification stays suppressed by default. Only a site that deliberately turned the box on will see a change, and the `pmpro_wp_new_user_notification` filter restores it:

```php
add_filter( 'pmpro_wp_new_user_notification', '__return_true', 20 );
```

Changes:

- `adminpages/emailsettings.php` - removed the save call, the `get_option` read, and the whole `#other-email-settings` section.
- `includes/email.php` - the suppression filter is now unconditional instead of gated on the option. Still priority 0, so a site filter at a later priority can return `true` and win.
- `classes/class-pmpro-wisdom-integration.php` - dropped the option from the telemetry payload.
- `includes/updates/upgrade_3_8_5.php` - new routine that calls `delete_option( 'pmpro_email_member_notification' )`.
- `includes/upgradecheck.php` - registers the routine behind a `pmpro_db_version < 3.85` gate, following the existing pattern.
- `includes/functions.php` - comment only. `pmpro_maybe_send_wp_new_user_notification()` and the filter signature are unchanged, so this stays backward compatible.

The db version slot is `3.85`, the next increment past the current tail (`3.84`). Happy to rename it at release if you track these differently.

1. Go to Memberships > Email Settings. The "Other Email Settings" section and the "Default WP notification email" checkbox are gone, and the page still saves normally.
2. Run a test checkout as a new member. Only the PMPro confirmation email arrives, no stock WordPress new user email.
3. Add `add_filter( 'pmpro_wp_new_user_notification', '__return_true', 20 );` to the theme's functions.php and run another checkout. The WordPress new user email now arrives as well.
4. For existing sites, run the upgrade from 3.84 to 3.85. `wp option get pmpro_email_member_notification` returns nothing and `pmpro_db_version` is `3.85`.

### Other information:

Tested on a local WP install with PMPro active: settings page loads and saves, checkout sends only the PMPro confirmation email, and the filter above brings the WordPress email back. `php -l` passes on every changed file. There is no test suite in the repo for this area, so verification was manual plus a phpcs base-vs-new comparison per file (no new violations).

### Changelog entry

> Removed the "Default WP notification email" setting from Email Settings. The WordPress new user notification stays off by default, since PMPro already sends its own confirmation email after checkout. Use the `pmpro_wp_new_user_notification` filter to send it.


dparker1005:
Hi @faisalahammad, thank you for picking up point 3 of #3626. The description of why the setting confuses people is fair.

We're going to close this one rather than merge it. As written it's a breaking change: the upgrade routine deletes the option and `includes/email.php` suppresses the notification unconditionally, so any site that had the box checked stops receiving both the admin "New User Registration" notice and the member's set-password email after upgrading, with no indication why. Trimming it to a UI-only removal would avoid that, but it would leave a hidden option that can't be set through the UI, which isn't much better.

The setting is off by default and labelled as recommended off, so it isn't causing harm sitting there. #3626 is a discussion issue about the Email Settings page as a whole, and if we remove this we'd rather do it as part of that redesign, where the compatibility question can be handled properly. Thanks again for the work here.

PR #3805, opened 2026-09-15
feat(memberslist): filterable sortable columns

### LINKED ISSUE #3528: Enhancement: Make list of sortable Members List columns filterable
There isn't any way to modify the list of sortable Members List columns. 

A filter that allows one to modify the list of sortable columns for the Members List.

**Additional context**
Such a filter would make sorting the Members List by columns such as First Name and Last Name a possibility.

faisalahammad:
### All Submissions:

Resolves #3528.

The Members List table had no way to change which columns are sortable. This PR makes the sortable columns list filterable and also makes the two most requested columns, First Name and Last Name, sortable out of the box.

* New filter `pmpro_memberslist_sortable_columns` in `get_sortable_columns()`. A column added here shows the sort link in the header. Note that the column key also needs an entry in the allowed orderbys array, which is already filterable through `pmpro_memberslist_allowed_orderbys`. The filter docblock explains this.
* First Name and Last Name are now sortable by default. Their values come from user meta, so `sql_table_data()` joins the matching `usermeta` row only when one of them is the sort column. The normal query and the count query are unchanged.
* The name entries in `sanitize_orderby()` use `MIN(um_first.meta_value)` because the query groups by `u.ID, mu.membership_id` and hosts running `ONLY_FULL_GROUP_BY` would otherwise reject the ORDER BY.
* Fixed along the way: if a requested `orderby` was not in the allowed list, the query was built as `ORDER BY  DESC`, which is invalid SQL. The code now falls back to the default order instead. This also protects developers who add a sortable header without an allow list entry.

1. Go to Memberships > Members List. Open Screen Options and make sure First Name and Last Name are shown.
2. Click the First Name header. The list sorts by first name, click again to reverse. Same for Last Name.
3. While sorted, run a search. The results stay sorted by the chosen column.
4. Check the cancelled, expired, and old members views still load and sort.
5. Sort by each of the previously sortable columns (username, email, level, status, joined, start date, end date, ID) and confirm they still work.
6. In a plugin or code snippet, run `add_filter( 'pmpro_memberslist_sortable_columns', function( $columns ) { unset( $columns['user_email'] ); return $columns; } );`. Reload the Members List. The email header no longer shows a sort link.
7. Visit the list with `&orderby=notacolumn` in the URL. The list still loads using the default order instead of a SQL error.

### Other information:

### Changelog entry

* ENHANCEMENT: Added the `pmpro_memberslist_sortable_columns` filter and made the First Name and Last Name columns sortable in the Members List.


dparker1005:
Hi @faisalahammad, thank you for this one. Merged into v3.9 with a small follow-up commit (c3e7501c2).

Since the point of #3528 was letting custom columns be sortable, the missing piece after your `pmpro_memberslist_sortable_columns` filter was a way to add the JOIN a custom ORDER BY needs without string-editing the finished query through `pmpro_members_list_sql`. So I wrapped your two name joins in a `! $count` block that builds a `$join_sql` string and passes it through a new `pmpro_memberslist_sql_join` filter along with the sanitized `$orderby`. Core's name sorting now goes through the same path a developer would use, and the docblock covers the `MIN()` requirement from the GROUP BY. No behavior change to your sorting or the invalid-orderby fallback.

PR #3806, opened 2026-09-15
fix: store member edit panel membership dates as UTC

### LINKED ISSUE #3422: PMPro_Member_Edit_Panel_Memberships.save incorrectly uses local timezone when saving membership start date
**Describe the bug**
When an admin edits a user membership, PMPro saves the membership start date to the database using the server's local timezone. WordPress's `current_time` function returns the local time unless the `$gmt` argument is specified.
`'startdate'       => current_time( 'mysql' )`

https://github.com/strangerstudios/paid-memberships-pro/blob/60736254ac30e799f84129e371eda4bb51e95e2b/adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php#L757-L771

However, when retrieving the start date function `pmpro_getMemberStartdate` assumes that the startdate in the DB is UTC (+00:00), and applies a TZ conversion to it. This causes an inaccurate startdate to be returned.
https://github.com/strangerstudios/paid-memberships-pro/blob/60736254ac30e799f84129e371eda4bb51e95e2b/includes/functions.php#L2865-L2869

**To Reproduce**
1. Edit a user's PMPro membership level through the Wordpress dashboard
2. While logged in as the user, execute a snippet such as
```php
	global $current_user;
	$startdate = pmpro_getMemberStartdate($current_user->ID);
	echo 'Membership start date: ' . date("Y-m-d H:i:s", $startdate) . '<br/>';
``` 
3. Incorrect start date will be returned

**Expected behavior**
Dates should be saved in the database as UTC, not local timezone

- [x] This bug happens with a default WordPress theme active, or [Memberlite](https://memberlitetheme.com).

**WordPress Environment**
<details>
Wordpress 6.8.1
PHP 8.2.28
PMPro 3.4.6
</details>

faisalahammad:
### All Submissions:

Resolves #3422.

The member edit panel saved `startdate` and `enddate` in site local time, but the plugin reads those columns as UTC (`CONVERT_TZ(startdate, '+00:00', @@global.time_zone)` in `pmpro_getMemberStartdate()` and other queries). On a non-UTC site the saved date came back shifted by the site offset.

The fix is in `adminpages/member-edit/pmpro-class-member-edit-panel-memberships.php` only:

* The add level path now stores `startdate` as UTC (`current_time( 'mysql', true )`) and converts the expiration field from local time to UTC with `get_gmt_from_date()` before inserting.
* The edit level path converts the submitted expiration field to UTC the same way.
* The expiration input is a `datetime-local` field, so its value has to be local time. The prefill now uses `wp_date()` for the existing end date, so loading a level and saving it again no longer drifts.

Before, on a site at UTC-4:

```php
'startdate' => current_time( 'mysql' ), // 12:00 local stored as if it were 12:00 UTC
```

After:

```php
'startdate' => current_time( 'mysql', true ), // stored as 16:00 UTC
```

Scope is limited to this panel. Other writers (checkout, IPN) and existing rows are untouched, so no data migration is needed. This is not a breaking change; no public functions, hooks, or filters changed.

1. Set the WordPress timezone to a non-UTC offset, for example UTC-4.
2. Go to Members, edit a member, and add a level from the Memberships panel.
3. Run `pmpro_getMemberStartdate( $user_id, $level_id )` for that member.
4. Expected result: the timestamp matches the time the level was added, and is not shifted by the site offset.
5. Edit the expiration date on an existing level, save, reload the panel, and save again without changes. Expected result: the date stays the same, no drift.

### Other information:

### Changelog entry

Fixed the member edit panel saving membership start and expiration dates in local time while the plugin reads them as UTC, which shifted the dates by the site timezone offset.


dparker1005:
Hi @faisalahammad, thank you for digging into #3422. The reasoning in the description is careful and the diff is minimal, which I appreciate.

I'm going to close this one because it would introduce the inconsistency it sets out to fix. PMPro stores `startdate` and `enddate` in site-local time everywhere: checkout, `pmpro_changeMembershipLevel()`, the member edit panel, and several Add Ons all write `current_time( 'mysql' )` or equivalent. The `CONVERT_TZ` reads then produce a WordPress-style local timestamp, which is compared against `current_time( 'timestamp' )` in the expiration cron and displayed with `date()`. So the current system is local end to end. Converting only the panel to UTC means a level added there stores a time four hours off from one added at checkout on a UTC-4 site, its expiration fires late relative to checkout-created levels, and the `wp_date()` prefill shows existing rows four hours early.

The reporter's observation is real but points at documentation: `pmpro_getMemberStartdate()` returns a local timestamp in the same sense as `current_time( 'timestamp' )`, so it should be compared against that rather than `time()`. Moving the columns to true UTC would be the right long-term fix, but that is a migration touching every writer and reader plus the Add Ons, not a change we can land one writer at a time.

Thanks again for the work here.

PR #3808, opened 2026-09-15
fix(refunds): block refunds past the gateway refund window

### LINKED ISSUE #3148: Remove "refund" button based on order date and gateway
PayPal only allows you to refund a payment within 180 days. We should consider to remove our refund link for orders outside this 180 day window when the gateway is PayPal. [See their documentation on how to issue a refund to confirm this](https://www.paypal.com/us/cshelp/article/how-do-i-issue-a-refund-help101).

Stripe doesn't appear to have any specific "terms" for when a refund is possible. We may want to build this into the gateway classes either way and leave Stripe / some as "unknown" or "indefinite".

faisalahammad:
### All Submissions:

Resolves #3148.

Right now the refund action is always attempted. For PayPal Express, PayPal rejects a refund older than 180 days and the admin just gets the generic "check the order notes for more information" message, with the real reason buried in the API log.

This adds a gateway refund window and checks it before the refund is attempted, so the admin gets a specific reason instead.

- `PMProGateway::get_refund_window_days()` on the base class returns null (no known limit). PayPal Express overrides it to 180. Stripe overrides it to null, since Stripe does not publish a refund time limit.
- `pmpro_get_refund_window_days( $order )` reads the window from the order's gateway class and passes it through a new `pmpro_refund_window_days` filter. Returning null from the filter allows refunds at any time.
- `pmpro_can_refund_order( $order )` runs the existing `pmpro_allowed_refunds()` checks first, then compares the order timestamp against the window. It returns true or a WP_Error with a specific message.
- The check runs in the orders admin refund case, both refund blocks in the member edit panel, and the profile save loop. Messages show through the existing message rendering, so no new output paths.

The refund button is left in place. It is hidden in some views and the maintainer asked for a clear reason instead of a missing button. The window is filterable for sites that want the old attempt first behavior:

```php
add_filter( 'pmpro_refund_window_days', '__return_null' );
```

`pmpro_allowed_refunds()` and `pmpro_refund_order()` keep their current signatures and return values. `pmpro_refund_order()` still returns bool on purpose, since a WP_Error return would be truthy and read as success at the existing call sites.

1. Start with a site on a gateway that has no window, like Stripe or Test Gateway. Refund an order of any age from Memberships > Orders. Confirm it refunds as before.
2. Switch the site to PayPal Express (sandbox) and pay for a level. Immediately refund that order from the order edit screen. Confirm it refunds and there is no window message.
3. Point an existing PayPal Express order's timestamp to more than 180 days ago, then try to refund it. Confirm the refund is blocked and the message names the 180 day limit instead of the generic order notes text.
4. Repeat step 3 from the member edit screen and from a profile save that removes the level with refund checked. Confirm the same specific message shows and the level changes still apply.
5. Add `add_filter( 'pmpro_refund_window_days', '__return_null' );`. Retry step 3. Confirm the refund is attempted again, so sites can opt out of the window.

### Other information:

Notes:

- I kept this out of the orders list table. That view loads orders with `nogateway = true`, so the gateway object is not available there and the window cannot be read.
- Free orders have no payment date, so they skip the window and behave as before.
- No `CHANGELOG.txt` or `readme.txt` edits, since those are handled at release time.

### Changelog entry

Added a refund window to the gateways. Refunds on orders that are past the gateway limit, such as 180 days for PayPal Express, are now blocked with a message explaining why, instead of a failed API call.


dparker1005:
Hi @faisalahammad, thank you for picking up #3148. The implementation is clean and the messages are clear.

We're going to close this one for two reasons. First, direction: on the issue Jason leaned toward attempting the refund and surfacing the gateway's own error rather than checking the date ahead of time, since a fixed window can block refunds that would have gone through and goes stale when the gateway changes its policy. Second, PayPal Express is now a deprecated gateway in PMPro, so we don't want to add new gateway-level features that only PayPal Express uses.

The underlying idea, giving admins a specific reason when a refund fails instead of the generic "check the order notes" message, is still valid and I've noted that on #3148. Thanks again for the work here.

PR #3809, opened 2026-09-16
Fix Stripe refunds for charges already refunded at Stripe

### LINKED ISSUE #2057: If a Stripe order was already refunded, let's still refund the order in PMPro instead of bailing.
From a discussion with @mircobabini:

> Also, for Stripe, it would be cool to catch the
> \Stripe\Exception\ApiErrorException
> and check if 'charge_already_refunded' === $e->getStripeCode()
> In that case, it succeeded. The refund status was just not synced with the gateway. Then you can create a function (post_process_refund) which:
> set the order status to refunded
> add the success order note
> send the refunded emails
> So you can use that function even on that exception case.

Currently we just throw an error. https://github.com/strangerstudios/paid-memberships-pro/blob/v2.8/classes/gateways/class.pmprogateway_stripe.php#L4803

faisalahammad:
### All Submissions:

Resolves #2057.

If a charge is refunded in the Stripe dashboard, PMPro still shows the order as paid or active. When an admin then refunds the order in PMPro, the Stripe API returns `charge_already_refunded` and the gateway adds a generic problem note, so the PMPro order never catches up and refund emails never go out.

This catches the error code and treats it as a successful sync:

- `PMProGateway_stripe::process_refund()` now catches `\Stripe\Exception\ApiErrorException` first. When `getStripeCode()` returns `charge_already_refunded`, the refund is treated as already done and the normal post-refund handling runs. Other API errors keep the existing error note path.
- The status save, success note, and refund emails are extracted into a new `PMProGateway_stripe::post_process_refund()` helper. The normal successful refund response and the already-refunded branch both use it, so both paths stay identical.
- The immediate success path now requires the Stripe refund status to be `succeeded` before marking the order refunded. Pending refunds complete later through the `charge.refunded` webhook, which already skips orders that are refunded.

The old `Stripe\Error\Base` catch blocks stay as they are. They are dead code since the bundled SDK moved to namespaced exceptions, and updating every catch is outside this issue's scope.

1. Pay for a level with Stripe in test mode. In the Stripe dashboard, refund that charge directly.
2. In PMPro, go to Memberships > Orders, open that order, and click Refund.
3. Confirm the refund reports success, the order status changes to Refunded, the success order note names the transaction ID, and the member and admin refund emails are sent.
4. Trigger the `charge.refunded` webhook for the same charge and confirm no duplicate status update or duplicate emails, since the order was saved as refunded first.
5. Refund an order that is not refunded at Stripe and confirm the normal path still works. Then force an unrelated API error and confirm the problem note still shows the message.

### Other information:

Tested on WordPress 6.8 with PHP 8.3 and the bundled Stripe SDK v18.0.0.

No `CHANGELOG.txt` or `readme.txt` edits, since those are handled at release time.

### Changelog entry

Updated the Stripe gateway to treat a charge that is already refunded at Stripe as a successful refund in PMPro. The order is marked refunded and the refund emails are sent instead of showing a failed refund.


dparker1005:
Hi @faisalahammad, thank you for picking up #2057 and for following the approach Mirco and Jason sketched out in the issue. Catching `charge_already_refunded` is the right call.

After reviewing, I opened #3811 with a trimmed version of the same fix and I'm going to close this PR in favor of it. The differences are small but deliberate:

1. **Pending refunds keep the current behavior.** Requiring `'succeeded'` would treat a `pending` refund as an error and show the admin a failure message, even though the refund is fine and the `charge.refunded` webhook finishes it. That was a behavior change outside the scope of the issue.
2. **No new public method.** `post_process_refund()` would become public API that Add Ons could start calling. #3811 wraps only the `refunds->create()` call and reuses the existing success block via a local flag, so the two paths stay identical without an extraction.
3. **Separate order note for the sync case.** Reusing the "successfully refunded by {admin}" note isn't accurate when the refund actually happened earlier at Stripe, so the already-refunded path records that the status was synced instead.

Thanks again for the work on this one, and for the steady stream of fixes lately. Closing in favor of #3811.

PR #3810, opened 2026-09-16
Add pmpro_is_page() wrapper function

### LINKED ISSUE #1958: We need a pmpro_is_page() wrapper.
There are numerous places in add ons and custom code where we write something like this:
`
global $pmpro_pages;
if ( is_page( $pmpro_pages['confirmation'] ) ) {
    // do something on the confirmation page
}
`

This looks good and works most of the time. The problem is that sometimes the confirmation page may not be set, and so our code there becomes effectively a check for `is_page(null)` which returns true on ANY page, not just the confirmation page.

A function that returned false if the request page tag/id was empty and otherwise returned true if we were on that specific page would be useful.

We should probably just accept the PMPro page string (confirmation, checkout, billing, etc) and look up the post id from the $pmpro_pages global in the new function.

In the meantime, until we have something like this available, you can write something like this:
`
global $pmpro_pages;
if ( ! empty( $pmpro_pages['confirmation'] && is_page( $pmpro_pages['confirmation'] ) ) {
    // do something on the confirmation page
}
`

faisalahammad:
### All Submissions:

Resolves #1958.

`is_page()` returns true when passed an empty value, so a direct call like `is_page( $pmpro_pages['confirmation'] )` matches every page when that page setting is unset.

This adds a `pmpro_is_page( $page, $shortcode = false )` helper in `includes/functions.php`:

* Returns false when the page key is empty or not set in `$pmpro_pages`, so an unset page can no longer match every page.
* Accepts add-on page keys registered through the `pmpro_extra_page_settings` filter.
* The optional second parameter checks the queried page content for the matching PMPro shortcode or block, so custom pages built with `[pmpro_checkout]` or a PMPro block can be detected too. Block names are mapped explicitly since they are not uniform (`pmpro/login-form`, `pmpro/member-profile-edit`).
* `pmpro_is_checkout()` now calls the helper with content detection on and keeps its existing `pmpro_is_checkout` filter and return behavior.
* The admin membership access check in `pmpro_getMembershipLevelsForUser()` now uses the strict form instead of repeating the same empty check for seven pages.

Other existing `is_page( $pmpro_pages[...] )` call sites are already empty-guarded, so they are left unchanged.

1. Unset one of the PMPro page settings, e.g. delete the Confirmation page under Memberships > Settings > Pages. Visit any front end page. With a snippet like `var_dump( pmpro_is_page( 'confirmation' ) );` in the footer, it returns `bool(false)` everywhere instead of true on every page.
2. Create a regular page containing the `[pmpro_confirmation]` shortcode. `pmpro_is_page( 'confirmation' )` returns false there, `pmpro_is_page( 'confirmation', true )` returns true. Same with the Confirmation Page block instead of the shortcode.
3. Create a custom checkout page with the `[pmpro_checkout]` shortcode and confirm `pmpro_is_checkout()` returns true on it, same as before.
4. Enable the "Admins only" membership access restriction and confirm a level still grants admin access on the Levels, Account, Billing, Checkout, Confirmation and Invoice pages as before.

### Other information:

### Changelog entry

Add a `pmpro_is_page()` wrapper function to safely check PMPro pages, with optional shortcode/block detection for custom pages.


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3810 — Add pmpro_is_page() wrapper function**
*faisalahammad → dev | 1 file, +64 -32*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3810>

---

**Summary**

Ready to merge after one docblock tweak. The bug is real: WP core's `WP_Query::is_page()` (`class-wp-query.php:4583`) returns `true` for any page when `$page` is empty, so the old unguarded `is_page( $pmpro_pages[...] )` calls matched every page whenever a page setting was unset. The new helper kills that at the API boundary, the admin-access refactor in `pmpro_getMembershipLevelsForUser()` is a 1:1 truth-table match for the seven old guards, and `pmpro_is_checkout()` keeps its exact cascade (`is_page` → shortcode → block → filter). All nine block names match the `name` fields in `blocks/build/*/block.json`; the shortcode names match the static registrations in `shortcodes/*.php` and the dynamic `pmpro_<page>` convention in `includes/init.php:88`. No prior `pmpro_is_page` declaration exists on `upstream/dev`.

---

**Issues**

- **Minor** `includes/functions.php:4149` — `@since 3.9` is a guessed version. Latest tag on `upstream/dev` is `3.8.6` and the plugin header on `dev` reads `Version: 3.8.6`, so the next release may not be 3.9. The repo convention is `@since TBD`, swapped en masse at release (see prior dev commits "Fixing named version release numbers to TBD" / "TBDs"). Change to `@since TBD`.

---

**Looks Good**

- The `$shortcode` content-detection path is short-circuited behind `! $is_page && $shortcode && ! empty( $wp_query )`, so the seven admin-access calls never build the lookup arrays or call `get_queried_object()`. No added per-request cost on the hot path.
- The explicit block map is more correct than the pattern in `pmpro_wp()` (`includes/init.php:65`), which builds `'pmpro/' . $name . '-page'` and would produce wrong names for `login` and `member_profile_edit`.
- `isset()` checks on the maps mean unknown or add-on page keys degrade to `false` rather than erroring.
- Removing `global $pmpro_pages` from `pmpro_getMembershipLevelsForUser()` is safe. Nothing else in that function references it.

---

**Questions**

1. Add-on page keys registered via `pmpro_extra_page_settings` work on the `is_page()` path, but `$shortcode = true` silently returns `false` for any key not in the hardcoded maps. The description says the helper "accepts add-on page keys," which slightly over-promises for content detection. Intentional, or worth a filter (e.g. `pmpro_is_page_shortcodes` / `pmpro_is_page_blocks`) so add-ons can register their own mappings?
2. The seven admin-access calls use `$shortcode = false`, matching old behavior, so an admin on a custom `[pmpro_account]` page still sees admin-override levels there. Fine to keep as-is for a pure bugfix, but is enabling detection there a follow-up you want tracked?

PR #3811, opened 2026-09-16
Treat Stripe charges already refunded at the gateway as a successful refund

### LINKED ISSUE #2057: If a Stripe order was already refunded, let's still refund the order in PMPro instead of bailing.

dparker1005:

Closes #2057.

If a charge was refunded directly in the Stripe dashboard, refunding the order in PMPro fails with `charge_already_refunded` and the order is left showing as paid. This catches that specific Stripe error code in `PMProGateway_stripe::process_refund()` and treats it as a successful refund: the order is marked refunded, an order note records that the charge was already refunded at Stripe and who synced it, and the member and admin refund emails are sent.

Only the `refunds->create()` call is wrapped. Any other Stripe API error is rethrown to the existing catch blocks, so error handling is otherwise unchanged. The existing success block is reused via a local flag, so the normal refund path and the already-refunded path stay identical. Pending refunds keep their current behavior.

Supersedes #3809 (same fix, smaller diff, no new public method, and no change to how pending refunds are handled).

1. Pay for a level with Stripe in test mode. In the Stripe dashboard, refund that charge directly.
2. In PMPro, open the order under Memberships > Orders and click Refund.
3. Expected: the refund reports success, the order status is Refunded, the order note says the transaction was already refunded at Stripe and names the admin who synced it, and the member and admin refund emails are sent.
4. Refund an order that has not been refunded at Stripe. Expected: normal path, unchanged note and emails.
5. Force an unrelated Stripe API error (for example a bad charge ID). Expected: the existing "There was a problem processing the refund" note.

### Other information:

Verified by reading and `php -l` only; not yet exercised against Stripe test mode.

### Changelog entry

> BUG FIX: Refunding an order in PMPro whose charge was already refunded at Stripe now marks the order as refunded and sends the refund emails instead of failing. #2057


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3811 — Treat Stripe charges already refunded at the gateway as a successful refund**
*dparker1005 → dev | 1 file, +22 -8*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3811>

---

**Summary**
Ready to merge. Narrow, correct fix: wraps only `refunds->create()`, catches the single `charge_already_refunded` code, rethrows everything else to the unchanged outer catch blocks, and reuses the existing success path via a local flag. The one thing I'd tighten is a cheap defensive tweak, not a blocker.

---

**Issues**
- **[Minor]** `classes/gateways/class.pmprogateway_stripe.php:4668-4682` — `$refund` is unset on the already-refunded path, and only PHP's short-circuit `||` in `$already_refunded || $refund->status != 'failed'` keeps it from being read. Correct today (`$refund` isn't referenced anywhere else in the function after that line), but it's fragile: swapping the operand order or adding an earlier `$refund` check later would throw an undefined-variable warning. Initialize `$refund = null;` alongside `$already_refunded = false;` before the `try` so the intent is explicit and static analysis stays quiet.
- **[Minor]** `classes/gateways/class.pmprogateway_stripe.php:4675` — the magic string `'charge_already_refunded'` has a matching constant in the bundled SDK, `\Stripe\ErrorObject::CODE_CHARGE_ALREADY_REFUNDED` (`includes/lib/Stripe/lib/ErrorObject.php:76`). Using it protects against drift across SDK bumps. Cosmetic.

---

**Looks Good**
- The exception type is right. `\Stripe\Exception\ApiErrorException` (`includes/lib/Stripe/lib/Exception/ApiErrorException.php`) is the base for `InvalidRequestException`, which is what Stripe throws for this code, and `getStripeCode()` is defined on it. `ApiConnectionException` also extends it but carries no Stripe code, so `getStripeCode()` returns null and the strict `!==` rethrows. Network failures cannot accidentally mark an order refunded.
- No double-fire risk. `pmpro_refund_order()` (`includes/functions.php:4966`) returns early when `$order->status == 'refunded'` without calling the gateway, so the sync path can't run twice and re-send refund emails.
- Reusing the existing success block rather than duplicating save/note/email logic keeps the two paths behaviorally identical, and the early `saveOrder()` preserves the existing webhook race mitigation.
- The distinct order note ("already refunded at Stripe... synced by") is the right call. It gives support an accurate audit trail instead of a misleading "successfully refunded" entry.
- Authorization and input surface are unchanged. The flag derives solely from Stripe's authenticated API response, not from anything user-controllable, and all callers remain the existing capability-gated admin paths.

---

**Questions**
1. Stripe only returns `charge_already_refunded` for a fully refunded charge. A partially refunded charge will refund the remainder and go down the normal path, which marks the PMPro order fully `refunded`. That matches pre-PR behavior, so no change requested, just confirming it's intentional.
2. The description says this hasn't been exercised against Stripe test mode yet. Worth one live run before merge to confirm the code string matches exactly, since the whole fix hinges on that comparison.

dparker1005:
Hi @flintfromthebasement, thank you for the review.

Both suggestions are in fed831db0: `$refund` is initialized to null before the inner try, and the code comparison now uses `\Stripe\ErrorObject::CODE_CHARGE_ALREADY_REFUNDED`. On your questions: partial refunds intentionally keep the existing behavior, and I'll do a test-mode run before merging.

PR #3812, opened 2026-09-16
fix(addons): block network activation

### LINKED ISSUE #1796: Filter plugin install screen to hide network activate when applicable.
Almost none of our plugins should be network activated. However, when a user installs a plugin from the link on our add ons list, it returns to this page in WP where the next proposed action is to network activate.

network_activate

We should either filter this to offer something like "return to previous blog to activate" or catch when a plugin gets networked activated and throw an error.

For the latter, we do this for the PMPro plugin itself. We maybe can do it for when add ons are activated too. It's probably okay to centralize this in the PMPro code and do it only if PMPro is also already active.

faisalahammad:
### All Submissions:

Resolves #1796.

PMPro add-ons should not be network activated on multisite, but the install screen still showed a Network Activate link for them. This PR removes that link for add-ons and adds a guard that network-deactivates an add-on if it gets activated network-wide anyway (single click, bulk activate, or activate after install).

Three parts, all in `classes/class-deny-network-activation.php`:

1. The `install_plugin_complete_actions` filter now unsets the `network_activate` action when the just-installed plugin is a PMPro add-on, so the link does not appear on the install screen.
2. An `activated_plugin` hook catches network-wide activations of add-ons, silently deactivates them, and records them in a short-lived site transient so bulk activations are not interrupted mid-batch.
3. A network admin notice on the plugins screen lists every denied plugin using the plugin's display name, pulled from `get_plugin_data()` so the message is readable even when the folder name does not match the add-on slug.

Add-on detection uses the cached `pmpro_addons` option (matching the `Slug` field against the plugin's folder name) with a `pmpro-` folder prefix fallback for add-ons not yet in the cache.

On a multisite install:

1. Delete a PMPro add-on (e.g. PMPro Mailchimp) if installed, then reinstall it from the Add Plugins screen. The success screen shows no Network Activate link. Reinstall a non-PMPro plugin and confirm its Network Activate link still appears.
2. Network activate the add-on from Network Admin > Plugins (single click or checkbox + Apply). The add-on ends up deactivated and a notice appears saying it should not be network activated, naming the plugin correctly.
3. Select two add-ons and network activate both in one bulk action. Both end up deactivated and two notices appear, one per plugin.
4. Activate the add-on from a single site's Plugins screen. It activates normally with no notice.

### Other information:

### Changelog entry

Hide and refuse network activation of PMPro add-ons on multisite installs.


dparker1005:
Hi @faisalahammad, thank you for picking up #1796. The implementation is careful, and the path sanitizer and bulk-activation transient are both well thought out.

Looking at this again, the issue predates the Add Ons page rebuild. Since PMPro 2.9 the Memberships > Add Ons screen installs and activates in place via AJAX, always single-site, so admins no longer land on the WordPress install-complete screen with the Network Activate link that prompted the issue. What remains is a super admin deliberately network-activating an add-on from Network Admin > Plugins, which is rare and self-correcting. Adding detection for that also carries some risk, since the `pmpro_addons` list includes third-party plugins like LifterLMS and WP Fusion that networks do legitimately network-activate.

So I'm going to close this and #1796 rather than ask for a rework. Thanks again for the work here.

Issue #918, opened 2019-01-29
BUG: Lower discount code field does not show on PayPal Express and Pay by Check gateways


greathmaster:

greathmaster:
https://github.com/strangerstudios/paid-memberships-pro/pull/966

ideadude:
The idea was to show the discount code field near the credit card fields where you are paying. If you don't have billing fields, it's not as necessary to show the discount code twice. Is this a problem for some?

I worry about the repercussions of having the billing fields shown for PayPal and offsite gateways. Talk me through this next time we chat.

dparker1005:
Thanks for opening this. Closing this because the lower discount code field is tied to the payment information section, and PayPal Express / Pay by Check do not show that section during checkout.

Discount codes can still be entered in the pricing area on the checkout page, so we do not plan to add a duplicate lower discount code field for gateways that skip the payment fields.

PR #3813, opened 2026-09-17
fix(checkout): open discount code field by default

### LINKED ISSUE #1489: Options to have discount code box "open" by default on checkout page.
Sites wanting to make it more obvious how to add a Discount Code when checking out.

Option to have the discount code field "open" without "click to add" message.  Also, add a box for PayPal Express users in the footer (similar to stripe)

faisalahammad:
### All Submissions:

Resolves #1489.

The discount code field on checkout always starts collapsed behind a "Do you have a discount code?" link. Sites that expect members to arrive with a code have to tell them to click that link first. This adds a setting to render the field open on first load.

Two files change:

- `adminpages/advancedsettings.php` reads and saves a new "Discount Code Field" checkbox under Advanced Settings > Other Settings, using the same `pmpro_setOption()` and `get_option()` pattern as the other rows on that page.
- `pages/checkout.php` reads the option and uses it for the initial `style` on the `#other_discount_code_p` prompt and the `#other_discount_code_fields` container. When the setting is on and no code is applied, the field is open and the prompt is hidden.

Nothing changes when the setting is off. An already applied code still collapses the field and shows the change link, review mode still hides the editable field, and the AJAX apply flow in `services/applydiscountcode.php` and the checkout script are untouched.

The other half of the original issue, a second discount code field for PayPal Express below the payment fields, is not included. That was decided against in #918 since the pricing-area field already works for those gateways.

1. Go to Memberships > Advanced Settings > Other Settings and check the new "Show the discount code field open by default on the checkout page" box, then save.
2. Visit a checkout page as a logged out user. The discount code input is visible right away and the "Do you have a discount code?" link is gone.
3. Enter a valid code and click Apply. The field collapses and the "Click here to change your discount code" link appears, same as before.
4. Uncheck the setting and reload checkout. The field is collapsed with the prompt link, the previous default behavior.
5. Reload checkout with `?discount_code=CODE` in the URL. The field stays collapsed and shows the change link rather than reopening.
6. Run a PayPal Express checkout through to the review step. The review summary is shown and the editable field stays hidden.

### Other information:

### Changelog entry

> Added an Advanced Settings option to show the discount code field open by default on the checkout page.


dparker1005:
Hi @faisalahammad, thank you for picking up #1489. The checkout logic here is correct and the diff is clean.

That said, we don't think this needs a core setting. Opening the discount code field by default is already covered by a published recipe using two lines of CSS: https://www.paidmembershipspro.com/always-show-discount-code-field/. Since that works against the current markup, we'd rather keep the Advanced Settings page focused and leave this as a site-level customization.

I'm going to close this one and #1489, but thanks again for the work here.

PR #3814, opened 2026-09-17
feat(discountcodes): add description field for discount codes

### LINKED ISSUE #1557: Feature request: Include a "Description" field for Discount Codes
Many codes can be created, and users can forget what they are for.

Add a field to the Discount Codes admin area for a "description" of the code.

faisalahammad:
### All Submissions:

Adds a Description field to discount codes, so admins can note what each code is for. Right now the code list has no way to record that, and people either put it in the code itself or keep it outside the site.

Changes:

- `includes/upgradecheck.php`: adds a `description` TEXT NOT NULL column to `pmpro_discount_codes` and a db version 3.9 gate that runs `pmpro_db_delta()` so dbDelta adds the column on upgrade.
- `adminpages/discountcodes.php`: reads `$_POST['description']` with `sanitize_textarea_field()`, includes it in the `pmpro_insert_or_replace()` data array, and renders a textarea in General Settings. The value carries over when a code is copied. Note the insert array is needed here: that helper uses `$wpdb->replace()` on edit, so a column left out would be reset on every save.
- `classes/class-pmpro-discount-code-list-table.php`: registers a Description column and renders the trimmed text with the full value in a title attribute.
- `classes/class-pmpro-discount-codes.php`: the legacy class reads and writes the field too. Without this, its raw INSERT would omit a NOT NULL column and fail on MySQL strict mode for REST creates and other callers of the class.

The description is plain text and admin only. It is not rendered on the front end and it does not show to members.

Resolves #1557.

1. Go to Memberships > Discount Codes and add a new code. Fill in the Code and Description fields, then Save.
2. Check the Discount Codes list. The new code shows the description you entered in the Description column, and hovering it shows the full text.
3. Edit that code and change the description. Save and confirm the list shows the new text.
4. Clear the description completely and Save. Confirm it is cleared and stays cleared after a reload.
5. Enter a multi line description. Save, then edit again and confirm the line breaks are kept.
6. Copy a discount code. Confirm the copied code keeps the description.
7. Load an install upgrading from an older db version and confirm the Description column exists and editing codes still works.

### Other information:

### Changelog entry

Added a Description field to discount codes, with the value shown in the Discount Codes list table. This is an internal admin note and is not shown to members.


dparker1005:
Hi @faisalahammad, thank you for picking up #1557. The implementation is careful, especially catching that the legacy discount code class needs the column in its INSERT for strict-mode hosts.

We're going to pass on this one, though, for a direction reason rather than a code one. We've been discussing an overhaul of notes across PMPro, covering member notes, order notes, and likely discount codes too, as a single system rather than a field per object. Adding a dedicated description column now would mean migrating it out later, so we'd rather hold the discount code piece for that work. There is also a practical conflict with #3727, which reworks the discount code screens and schema for 3.9.

I'm going to close this in favor of that longer-term plan, but thanks again for the effort here.

PR #3818, opened 2026-09-17
Fix double-escaped quotes in email template variable autocomplete descriptions

dparker1005:
## Summary
In the email template editor, the Liquid autocomplete showed variable descriptions with literal HTML entities, e.g. `The &quot;From Email&quot; set in Memberships > Settings > Email Settings.`

This PR changes every `esc_html__()` inside `get_base_email_template_variables_with_description()` and the 25 `get_email_template_variables_with_description()` implementations to `__()`. 26 files, 402 lines, mechanical, no other changes.

## Root cause
The description strings are returned as data from getters and consumed in two places:

1. The variable reference table in `adminpages/emailtemplates-edit.php`, which echoes each description through `esc_html()`.
2. The Liquid autocomplete, which passes them through `wp_json_encode()` into TinyMCE and renders them in JS with `textContent`.

Because the getters wrapped each string in `esc_html__()`, the data was already HTML-encoded before it reached either consumer. The reference table hid this because `esc_html()` tolerates existing entities. The autocomplete exposed it because `textContent` renders characters literally, so `&quot;` shows up on screen. Any translation containing an apostrophe would show `&#039;` in the same way, so this is not limited to the two English strings with quotes.

## Why fix the source instead of decoding in the autocomplete
Decoding in the autocomplete (either PHP-side before JSON encoding or JS-side before render) works and is safe, but it papers over the actual bug: data that is HTML-escaped at storage has to be decoded by every non-HTML consumer that comes later (JSON, plain-text email, REST, CLI). Escaping once at output is the WordPress rule for exactly this reason. From the developer handbook (Security > Escaping): escape as late as possible, ideally at output, so the state of the data is always known.

Core follows this itself. Post type labels, settings field titles, and the strings core hands to TinyMCE and other scripts are all plain `__()` and are escaped by whichever function renders them. `esc_html__()` is a convenience for a translated string that is echoed directly into HTML on the same line, not a stronger form of translation. The WordPress Coding Standards sniffs enforce escaping at output statements only and never require it on values stored in arrays.

## History
The abstract class originally used `__()` for these descriptions (c895c4538, Oct 2024) with `esc_html()` at output. The switch to `esc_html__()` came from a review request in #3234 to escape every `__()` in that PR. That was defensive and never a security issue, since extra escaping cannot cause XSS, but it did make the arrays mean two different things depending on the consumer. The autocomplete (3.8) is the first consumer that could not tolerate that.

## Safety
No output becomes unescaped. Both consumers already escape or render as text, verified by reading both code paths. The strings are authored in PMPro's PHP source, translation files, and add-on classes, so the trust boundary is unchanged.

## Notes
- Other `esc_html__()` calls in these files (email bodies, template descriptions, and help text) are unchanged. Only the variable description arrays were touched.
- Add-ons that supply their own description arrays with `esc_html__()` will still show entities if a description contains quotes. They should follow the same pattern.
- @ideadude, flagging since this reverses the escaping request from #3234 for these specific arrays.

## Proposed changelog entry
* BUG FIX: Fixed an issue where quotes in email template variable descriptions were shown as HTML entities in the editor autocomplete.


kimcoleman (approved):

flintfromthebasement (approved):
**PR: strangerstudios/paid-memberships-pro#3818 — Fix double-escaped quotes in email template variable autocomplete descriptions**
*dparker1005 → dev | 26 files, +402 -402 lines*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3818>

---

**Summary**
Ready to merge. This is a mechanical `esc_html__()` → `__()` swap across `get_base_email_template_variables_with_description()` and the 25 concrete `get_email_template_variables_with_description()` implementations, and the diff contains nothing else. Both consumers of these arrays escape at output, so no string becomes unescaped, and the autocomplete stops rendering `&quot;` literally.

---

**Looks Good**
- **Consumers verified at head `16b4abed6c`.** The only callers of the changed getters are in `adminpages/emailtemplates-edit.php`. The reference table echoes each description through `esc_html()` at `adminpages/emailtemplates-edit.php:257`. The autocomplete path goes through `pmpro_get_liquid_autocomplete_settings()` (`includes/functions.php:4520`, passes `$description` straight through at `includes/functions.php:4578`) into `wp_json_encode()` at `adminpages/emailtemplates-edit.php:86`, and the JS renders it with `textContent` at `js/pmpro-liquid-autocomplete.js:153`. The only `innerHTML` uses in that file (lines 95 and 114) clear the container. That matches the PR's stated root cause exactly.
- **Fixing the source instead of decoding in the consumer** is the right call. Escaping at storage forces every later non-HTML consumer (JSON, plain text, REST) to decode, and the autocomplete was the first one that couldn't tolerate it. Escape-late is the WordPress rule and this restores it.
- **Scope discipline.** Every touched function body is fully converted with no leftover `esc_html__()`, and the other `esc_html__()` calls in these files (email bodies, help text, template descriptions) are untouched. Diff stat matches the description.
- **Trust boundary unchanged.** Strings originate in PMPro PHP source and `.mo` files. Anyone who can swap translation files already reaches thousands of plain `__()` output sites, so this adds no new privileged path.
- The author already flagged the residual: add-ons overriding `get_email_template_variables_with_description()` with `esc_html__()` will still show entities. That's cosmetic (the sinks escape regardless) and correctly left out of this PR.

---

**Questions**
1. The legacy `!!variable!!` description arrays at `adminpages/emailtemplates-edit.php:24-53` still use `esc_html__()`. They're only consumed by the reference table (escaped at line 257) and are replaced when an email template class exists (line 67), so the double-escape is harmless today. Leaving them alone for now is fine, just confirming it's intentional rather than a follow-up you meant to include.

PR #3819, opened 2026-09-17
A11y: Announce the Liquid autocomplete to screen readers

kimcoleman:
### All Submissions:

* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/strangerstudios/paid-membership-pro/pulls) for the same update/change?

Supersedes #3702. The approach, the announcement strings and the diagnosis are @RachelRVasquez's, with follow-up work by @dparker1005. This PR keeps that behavior and trims the mechanism after testing with NVDA.

The Liquid autocomplete in the email template editor was silent for screen reader users. The listbox lives in the admin document while focus stays inside the TinyMCE iframe, so `aria-activedescendant` never engages. This announces the list through live regions instead.

- Uses core's `wp.a11y.speak()` rather than custom live regions. `common.js` depends on `wp-a11y`, so it is present on every wp-admin screen.
- Announces "Autocomplete list opened" plus the first option when the list appears, and each option with its position ("3 of 24") as you arrow through.
- Opening and filtering are debounced by 150ms so a keystroke burst is described once, when you pause. Closing and arrow navigation speak immediately, which removes the queued-close edge cases from #3702.
- Announces "Autocomplete list closed" on Escape, click-away, and filtering to no matches. Blur stays silent.
- Choosing an option now announces "Inserted display_name". NVDA does not read content inserted at the caret, so the previous silent-on-select left the user with no confirmation and a half-finished option announcement. The assertive message cuts that off and confirms the insertion.
- Spoken text uses the bare token name (`display_name`, `if`, `upcase`) instead of the braced label, so tags are no longer read as "left brace percent if dot dot dot".
- Live regions are emptied on close so the last option cannot be re-read later.
- Keeps the `shouldIgnoreKeyup()` modifier-key additions from #3702.

Tested with NVDA. VoiceOver + Safari still worth a pass.

1. Go to Memberships > Settings > Email Templates and open a Liquid template such as **Checkout - Free**.
2. Click into the Body field and type `{{`. Expect "Autocomplete list opened, display_name, The display name of the user, 1 of N".
3. Arrow up and down. Each option is read with its position.
4. Type to filter. The list is re-described once you pause.
5. Press Escape. Expect "Autocomplete list closed" right away. Reopen and click elsewhere in the body; same announcement.
6. Reopen and type until nothing matches. Expect "closed" once.
7. Reopen, arrow to an item, and press Enter or Tab mid-announcement. Expect the option to cut off and "Inserted <name>". Note that Tab chooses the option; it does not tab out of the editor.
8. Reopen and click outside the editor entirely. The list closes with no announcement.

### Other information:

```
* ENHANCEMENT: Added screen reader announcements to the Liquid autocomplete in the email template editor.
```


flintfromthebasement (commented):
**PR: strangerstudios/paid-memberships-pro#3819 — A11y: Announce the Liquid autocomplete to screen readers**
*kimcoleman → v3.9 | 2 files, +161 -15*
<https://github.com/strangerstudios/paid-memberships-pro/pull/3819>

---

**Summary**

Ready to merge. The open/filter/navigate/close announcement state machine in `js/pmpro-liquid-autocomplete.js` holds up under tracing: the debounce and dedupe in `announceActive()` interact correctly (an early return on identical filter text leaves the earlier timer to fire once, which is the intended behavior), `cancelAnnouncement()` mirrors what `@wordpress/a11y`'s own `clear()` does before each message, and the `wp-a11y` availability claim checks out (`wp-admin/admin.php` enqueues `common`, which depends on `wp-a11y` in `wp-includes/script-loader.php`). Two minor items, neither blocking.

---

**Issues**

- **Minor** `js/pmpro-liquid-autocomplete.js:35` — `handleResize()` still calls `closeMenu()` without the `silent` flag, so a window resize while the list is open speaks "Autocomplete list closed." Every other non-deliberate close in this PR (blur, hide, remove) was switched to `closeMenu( true )`; resize is the one that was missed. Zoom changes and virtual-keyboard shows also fire `resize`, so this will surface for real users. Change to `closeMenu( true )`.

- **Minor** `js/pmpro-liquid-autocomplete.js:616-621` — `( strings.autocompleteInserted || 'Inserted %s' ).replace( '%s', item.name )` passes `item.name` as a string replacement, where `$&`, `$$`, `` $` ``, `$'` are special. Variable names flow through filterable suggestion builders (`pmpro_get_liquid_autocomplete_variable_suggestions()` and friends in `includes/functions.php`), so add-ons control some of them. Spoken-text only, no security impact, but cheap to harden: `.replace( '%s', function () { return item.name; } )`. The `%1$d`/`%2$d` replacements in `describeActive()` are fine since the values are numbers.

---

**Looks Good**

- Wrapping the `click` and `blur hide` handlers (`:708-713`) instead of passing `closeMenu` straight to `editor.on` is the right catch. Passing it directly would have coerced TinyMCE's event object into `silent` and suppressed every "closed" announcement.
- `filterItems()` (`:577-582`) is the single construction point for all three suggestion lists, so `item.name` is guaranteed present everywhere `describeActive()` and `chooseActive()` read it.
- The `announceOpened` flag surviving until something is actually spoken handles the `{{` double-render cleanly, and navigation interrupting a pending open timer still carries the prefix rather than dropping or doubling it.
- The mousedown path (`:76-79`) calls `setActive()` with no reason, which schedules a polite timer, but `chooseActive()` immediately runs `closeMenu( true )` and `cancelAnnouncement()` clears it before it fires. Correct as written; an explicit reason would just make the intent self-documenting.
- New strings in `includes/functions.php:4566-4573` all have translator comments and ride the existing `wp_json_encode()` transport.

---

**Questions**

1. `updateMenu()` (`:592`) also closes non-silently when the caret leaves the `{{` context, for example after typing a space or deleting the braces. That reads as intentional and matches the "no matches" case, but confirm that was tested with NVDA rather than only Escape and click-away.
2. Is the VoiceOver + Safari pass a merge blocker or a follow-up?

kimcoleman:
Thanks for the review. Both minor notes addressed in 978775222:

- **Resize close is now silent.** `handleResize()` calls `closeMenu( true )`, matching blur/hide/remove. Window resize fires on zoom and virtual keyboard changes, so those closes no longer announce "Autocomplete list closed."
- **Inserted announcement hardened.** The `%s` replacement in `chooseActive()` uses a function replacer so variable names containing `$&`, `$$`, etc. are spoken verbatim. Left the numeric `%1$d`/`%2$d` replacements in `describeActive()` as-is.

dparker1005 (approved):

